PR #2, opened 2020-02-19
Chet 92 - AWS region rest endpoint

bpar476:
Introduces small endpoint to query all AWS regions for populating drop-downs in frontend.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -117,7 +117,6 @@
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>cloudformation</artifactId>
-            <scope>compile</scope>
             <exclusions>
                 <exclusion>
                     <groupId>software.amazon.awssdk</groupId>
--- a/pom.xml
+++ b/pom.xml
@@ -117,7 +117,6 @@
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
             <artifactId>cloudformation</artifactId>
-            <version>${aws.sdk.version}</version>
             <scope>compile</scope>
             <exclusions>
                 <exclusion>
@@ -125,6 +124,11 @@
                     <artifactId>apache-client</artifactId>
                 </exclusion>
             </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>regions</artifactId>
+            <scope>compile</scope>
         </dependency>
         <dependency>
             <groupId>com.amazonaws</groupId>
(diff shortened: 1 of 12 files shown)


wrm.d.ts
```diff
@@ -0,0 +1,3 @@
+declare module 'wrm/context-path' {
+    export default function contextPath(): string;
```
theghostwhoforks: will this return `https://<app-name>/jira` or  `https://<app-name>/`?

src/main/dc-migration-assistant-fe/components/aws/AwsRoutes.tsx
```diff
@@ -0,0 +1,32 @@
...
+        <Route path={quickstartPath}>
+            <QuickStartDeploy />
+        </Route>
+        <Route path={awsAuthPath}>
+            <AuthenticateAWS
+                getRegions={getRegions}
+                onSubmitCreds={(creds): Promise<string> => {
+                    // This should be replaced with API call that stores the credentials
+                    // eslint-disable-next-line no-console
+                    console.log(creds);
```
nanux: Should we log them here?

wrm.d.ts
bpar476: It will simply return the tomcat context path e.g. `/jira` or `/confluence`. It doesn't return the URL as it is a same origin request e.g. full path may be `/jira/rest/1.0/migration` or `/rest/1.0/migration`

src/main/dc-migration-assistant-fe/components/aws/AwsRoutes.tsx
bpar476: No we probably shouldn't log them in the production version. This is just there as a placeholder for where we can put the submit action in the future.

wrm.d.ts
bpar476: This file itself is just the module declaration that tells the typescript compiler that the `wrm/context-path` module is available at runtime even though it isn't present in the node modules. It's built into the bundle via some webpack magic.

pom.xml
```diff
@@ -126,6 +125,11 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>regions</artifactId>
+            <scope>compile</scope>
```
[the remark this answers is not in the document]
bpar476: Hmm, I was getting the osgi error when it wasn't specified as compile. Let me test further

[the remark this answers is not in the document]
nanux: Interesting - but the OSGi path of a warrior is winding... so it is entirely possible.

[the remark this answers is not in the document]
bpar476: Lol I just need to alias `mvn X` to `mvn clean X` I swear it solves all my problems

PR #4, opened 2020-02-20
CHET-28 Use DirectoryStream to iterate over directories instead of using the …

andrewpowell:
Addresses TransferManager issue of listing all files first in [CHET-28]

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -4,9 +4,9 @@
 
     <modelVersion>4.0.0</modelVersion>
     <parent>
-      <groupId>com.atlassian.pom</groupId>
-      <artifactId>public-pom</artifactId>
-      <version>6.0.2</version>
+        <groupId>com.atlassian.pom</groupId>
+        <artifactId>public-pom</artifactId>
+        <version>6.0.2</version>
     </parent>
 
     <repositories>
@@ -133,6 +133,12 @@
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk-s3</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.projectlombok</groupId>
+            <artifactId>lombok</artifactId>
+            <version>1.18.12</version>
+        </dependency>
+
 
         <!-- Test Dependencies follow -->
         <dependency>
--- a/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
@@ -15,7 +15,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
 
+import java.io.IOException;
+import java.nio.file.DirectoryStream;
+import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.concurrent.Executors;
 
 @Component
 public class S3FilesystemMigrationService implements FilesystemMigrationService {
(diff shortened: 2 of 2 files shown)


src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -27,21 +31,36 @@
...
-                    DEFAULT_INCLUDE_SUBDIRECTORIES);
-            progress.setStatus(FilesystemMigrationStatus.RUNNING);
-            S3UploadListener listener = new S3UploadListener(progress);
-            upload.addProgressListener(listener);
-            FilesystemMigrationStatus status = waitToComplete(upload);
-            progress.setStatus(status);
+
+            int cores = Runtime.getRuntime().availableProcessors();
+
+            try (DirectoryStream<Path> directoryStream = Files.newDirectoryStream(config.getDirectoryToMigrate().toFile().toPath())) {
```
theghostwhoforks: @nanux will chime in here with more details - when we use directory stream, we get a handle to the first level of the directories. For example, on a confluence/jira instance, we get these directories -

```
[confluence@conf-dc shared-home]# tree -L 1
.
├── analytics-logs
├── attachments
├── backups
├── config
├── confluence.cfg.xml
├── confluence.version
├── imgEffects
├── index-snapshots
├── plugins
├── synchrony-args.properties
└── viewfile
```

Specifically, the folder `attachments` is heavily nested and has lots and lots of small files.

Now, the issue we noticed with using transfer manager is that the code walks over the entire folder structure. (See - https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-s3/src/main/java/com/amazonaws/services/s3/transfer/TransferManager.java#L1926).

My suspicion is that even if we were to give the transfer manager a handle to the `attachments` directory, on a reasonably sized instance, this approach will still be slow. 

Thoughts @nanux? Would taking this approach a few steps further and giving the transfer manager a handle to a nested directly 2 or 3 levels deep make a positive impact on performance?

pom.xml
```diff
@@ -133,6 +133,12 @@
             <groupId>com.amazonaws</groupId>
             <artifactId>aws-java-sdk-s3</artifactId>
         </dependency>
+        <dependency>
```
theghostwhoforks: Perhaps, we don't need lombok as yet?

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
bpar476: I agree that the attachments really is the biggest culprit when it comes to the file tree depth. I think there is definitely some optimisation we can get from parallelism. I think the approach I would take if we were to implement this ourselves is a producer/consumer model where you have a number of threads crawling the file system and adding files to some conccurent queue and a number of threads polling the queue for files to upload to S3.

That said, I think most approaches are going to need to be verified with testing on an actual Jira instance.

nanux: As the directory structure is not evenly distributed, we are not going to have a lot of improvement with this approach. Eventually, all directory levels will be loaded into a LinkedList so we are saving only first layer. Most of the depth is inside the `attachments` folder.
@andrewpowell what are the implications of switching to SDKv2 and don't use `TransferManager`? We would need to reimplement some parts of it (e.g. uploadDirectory and also the multipart upload but we would have more control over the execution. If we would combine it with DirectoryStream it might be just sufficient.

nanux:
Hi @andrewpowell, thank you for the recommendation and PoC. We are now working with @bpar476 on implementation that uses `DirectoryStream` and `PutObject` with publisher and subscriber pattern. This will allow us to have more control over the upload and it also uses just AWS SDKv2 API with Async client. I will close this PR as we've effectivelly deprecated it and open a new one with our changes.

PR #7, opened 2020-02-21
MOT-85: Add REST endoint and change S3Upload to ScheduledJob

nanux:


src/main/java/com/atlassian/migration/datacenter/core/fs/S3UploadMonitorRunner.java
```diff
@@ -0,0 +1,35 @@
+package com.atlassian.migration.datacenter.core.fs;
```
bpar476: Do we need to create a follow-up to schedule and unschedule this job for the duration of the upload?

pom.xml
```diff
@@ -114,6 +114,29 @@
...
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.woodstox</groupId>
+            <artifactId>woodstox-core</artifactId>
+            <version>6.0.3</version>
+        </dependency>
+        <dependency>
+            <groupId>jta</groupId>
+            <artifactId>jta</artifactId>
+            <version>1.0.1b</version>
```
nanux: Required as `1.0.1` required by `jira-api` doesn't exist in the maven repos enymore.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3UploadMonitorRunner.java
nanux: Yes, you are correct. I've created https://aws-partner.atlassian.net/browse/CHET-108

PR #8, opened 2020-02-21
Feature/CHET-80 api for checking provisioning status

andrewpowell:
Adding API for retrieving CFN stack status


bpar476:
There seems to be a bit of duplication between this PR and #9 so it may be best to wait until the common stuff has merged before reviewing this properly.

bpar476:
This isn't required anymore as @theghostwhoforks has implemented this functionality in #10

PR #11, opened 2020-02-25
CHET-79: POST quickstart form data to API

theghostwhoforks:
This PR acts as a POC to show how we can poll the CloudFormation API periodically and stop polling once the status has either reached created or failed.

The component can be accessed by hitting `plugins/servlet/dc-migration-assistant/aws/<stackName>/status`

This includes -

1. Changes to routing to accept a path parameter with stack Name
2. A new status component that uses a flag to symbolize if a stack has been created successfully or not
3. Introduces a Setinterval timer to repeatedly call the rest API. The timer is cleared if a success/error response has been reached


src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -1,23 +1,24 @@
-import React, { FunctionComponent, ReactElement, useState, useEffect } from 'react';
+import React, {FunctionComponent, ReactElement, useEffect, useState} from 'react';
```
bpar476: It looks like your editor style isn't set correctly. Are you using vs code? You should be using the `prettier` formatter.

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -63,8 +80,14 @@ const QuickstartForm = ({
                 />
 
                 {quickstartParamGroups.map(group => {
+                    let stackNameField : React.ReactElement = createQuickstartFormField({
```
bpar476: You should probably use internationalisation for the label and description

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -35,25 +36,35 @@ const QuickstartSubmitButton = styled(Button)`
...
-            // TODO: Send this to the API
-            // eslint-disable-next-line no-console
-            console.log(transformedCfnParams);
+            callAppRest('POST', 'aws/cloudformation/create', {
+                templateUrl: 'https://dcd-slinghost-templates.s3-ap-southeast-2.amazonaws.com/mothra/test-create-s3-bucket.yaml',
+                stackName: stackName,
+                params: transformedCfnParams,
+            }).then(x => x.text())
+                // eslint-disable-next-line no-console
+                .then(console.log);
```
bpar476: We probably don't need to log anything here anymore. However, it may be good to show a popup if the POST fails.

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -79,13 +90,29 @@ const QuickstartForm = ({
     </Form>
 );
 
+const buildAdditionalParameters = () : Array<QuickstartParameterGroup> => {
```
bpar476: Instead of going through the process of generating the form parameter, I would just add the stack title directly to the form i.e.

```
<QuickstartFormContainer {...formProps}>
    <FormHeader
        title={I18n.getText('atlassian.migration.datacenter.provision.aws.form.title')}
    />
    <!-- This is where the stack name field should go instead of using the atlaskit form element generator -->
    <Field label={i18n.getText('theLabel')} name="theLabel">
            {({ fieldProps }: any): ReactElement => (
                <TextField width="mediuM" {...fieldProps}/>
            )
    {quickstartParamGroups.map(group => {
        return (
            <FormSection key={group.groupLabel} title={group.groupLabel}>
                {group.parameters.map(parameter => {
                    return createQuickstartFormField(parameter);
                 })}
            </FormSection>
        );
    })} 
   <QuickstartSubmitButton type="submit" appearance="primary">
        {I18n.getText('atlassian.migration.datacenter.generic.submit')}
   </QuickstartSubmitButton>
</QuickstartFormContainer>
```

bpar476: The param generator is for dynamically generating the form content based on the quick start parameters. We know we will always need a stack name so we don't really need to generate it from the `QuickStartParameter` construct

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
bpar476: This can be done with 
```
callAppRest(...).catch(err => {
   // Render an atlaskit flag or something
});
```
See `example.tsx` in this codesandbox for details on how to do error flag with Atlaskit if you like https://codesandbox.io/s/atlaskitflag-01-flag-without-flaggroup-1nyzl?fontsize=14&hidenavigation=1&theme=dark

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
theghostwhoforks: Yup - Intellij is mucking around with this 🤷‍♂

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
theghostwhoforks: Yup - I'm logging to ensure we get the CFN identifer out and see how we can poll it until complete. Will wire the promise onError callback as part of this story

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
theghostwhoforks: Haha. I think I had this in before injecting this field into the params array.

I'm not opinionated on this one way or another. Adding it to the params array makes it part of the react state, so I'm not sure if that's a desired side-effect or not.

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
bpar476: Try with this plugin: https://www.jetbrains.com/help/idea/prettier.html

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
bpar476: It probably shouldn't be in the state as it's not really stateful. I reckon it should just be its own form, it's not a part of the quick start template so it doesn't need to go through quickstart to atlaskit hydrator

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -13,11 +14,17 @@ import {
...
+import { callAppRest } from '../../../utils/api';
+import { awsStackCreateRestPath } from '../../../utils/RoutePaths';
+
 const QUICKSTART_PARAMS_URL =
-    'https://dcd-slinghost-templates.s3.amazonaws.com/mothra/quickstart-jira-dc-with-vpc.template.parameters.yaml';
+    'https://dcd-slinghost-templates.s3.amazonaws.com/mothra/test-create-s3-bucket.parameters.yaml';
+
+const STACK_NAME_FIELD_NAME = 'QSStackName';
+const stackProvisioningTemplateUrl =
+    'https://dcd-slinghost-templates.s3-ap-southeast-2.amazonaws.com/mothra/test-create-s3-bucket.yaml';
```
nanux: Were the QS S3 urls just for testing purposes?

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -35,33 +42,78 @@ const QuickstartSubmitButton = styled(Button)`
...
 }: Record<string, Array<QuickstartParameterGroup>>): ReactElement => (
     <Form
         onSubmit={(data: Record<string, any>): void => {
             const transformedCfnParams = data;
+            let stackNameValue = '';
             Object.entries(data).forEach(entry => {
                 // Hoist value from Select/Multiselect inputs to root of form value
                 const [key, value] = entry;
-                if (isOptionType(value)) {
+                if (key === STACK_NAME_FIELD_NAME) {
```
nanux: `// stackName is a special case as it is sent as a separate parameter to AWS API`

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -35,33 +42,78 @@ const QuickstartSubmitButton = styled(Button)`
...
+
 const QuickstartForm = ({
     quickstartParamGroups,
 }: Record<string, Array<QuickstartParameterGroup>>): ReactElement => (
     <Form
         onSubmit={(data: Record<string, any>): void => {
             const transformedCfnParams = data;
+            let stackNameValue = '';
             Object.entries(data).forEach(entry => {
                 // Hoist value from Select/Multiselect inputs to root of form value
```
nanux: /nit This comment could be for the `else if` branches

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -35,33 +42,78 @@ const QuickstartSubmitButton = styled(Button)`
...
+                stackName: stackNameValue,
+                params: transformedCfnParams,
+            })
+                .then(response => {
+                    if (response.status !== 202) {
+                        throw Error('Stack provisioning failed');
+                    }
+                    return response;
+                })
+                .then(x => x.text())
```
nanux: `r` => `r.text()`?

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -35,33 +42,78 @@ const QuickstartSubmitButton = styled(Button)`
...
+                .catch(err => {
+                    console.log(err);
```
nanux: `console.error(err)`?

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
```diff
@@ -0,0 +1,62 @@
...
+    return (
+        <Button appearance={isStackProvisioned ? 'primary' : 'warning'}>
+            Proceed to Filesystem migration
+        </Button>
+    );
+};
+
+export const QuickStartStatus: FunctionComponent = (props: Record<any, any>): ReactElement => {
+    const [inProgress, setInProgress]: [boolean, Function] = useState(true);
+    const [provisioningFailed, setProvisioningFailed]: [boolean, Function] = useState(false);
```
nanux: ok, I can see what you meant when you mentioned `React` landscape is still rapidly changing 🦥

src/test/resources/cfn/test-create-s3-bucket.parameters.yaml
```diff
@@ -0,0 +1,16 @@
+ParameterGroups:
```
nanux: Are we going to have integration tests for the frontend as well?

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
theghostwhoforks: Yeah - will be fixed now

package.json
```diff
@@ -5,7 +5,10 @@
     "author": "Atlassian Pty Ltd",
     "license": "Apache 2",
     "dependencies": {
+        "@atlaskit/icon": "^20.0.1",
+        "@atlaskit/css-reset": "latest",
```
bpar476: Nice 😎

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -13,11 +14,18 @@ import {
     QuickStartParameterYamlNode,
     QuickstartParamGroupYamlNode,
     QuickstartParamLabelYamlNode,
-    // eslint-disable-next-line import/extensions
 } from './QuickStartTypes';
 
+import { callAppRest } from '../../../utils/api';
+import { awsStackCreateRestPath } from '../../../utils/RoutePaths';
```
bpar476: I think the REST API paths should be in a different file or encapsulated in the API utility as `RoutePaths` is meant to be for client-side routing.

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -13,11 +14,18 @@ import {
...
+
 const QUICKSTART_PARAMS_URL =
-    'https://dcd-slinghost-templates.s3.amazonaws.com/mothra/quickstart-jira-dc-with-vpc.template.parameters.yaml';
+    'https://dcd-slinghost-templates.s3.amazonaws.com/mothra/test-create-s3-bucket.parameters.yaml';
```
bpar476: We should change this back to the original template now, or make it configurable based on the deployment environment. E.g. read from environment when in dev mode and use hard coded value for production

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -35,33 +43,79 @@ const QuickstartSubmitButton = styled(Button)`
...
             const transformedCfnParams = data;
+            let stackNameValue = '';
             Object.entries(data).forEach(entry => {
                 // Hoist value from Select/Multiselect inputs to root of form value
                 const [key, value] = entry;
-                if (isOptionType(value)) {
+                if (key === STACK_NAME_FIELD_NAME) {
+                    // stackName is handled separately as it is sent as a separate parameter to AWS API
+                    stackNameValue = value;
+                    delete transformedCfnParams[key];
```
bpar476: If you apply the above suggestion we can remove this extra edge-case to handle
```suggestion
```

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -35,33 +43,79 @@ const QuickstartSubmitButton = styled(Button)`
...
+    );
+};
+
 const QuickstartForm = ({
     quickstartParamGroups,
 }: Record<string, Array<QuickstartParameterGroup>>): ReactElement => (
     <Form
         onSubmit={(data: Record<string, any>): void => {
             const transformedCfnParams = data;
+            let stackNameValue = '';
```
bpar476: Because you've directly created the `TextField` now this is much simpler 😄
```suggestion
            const stackNameValue = transformedCfnParams['stackName']
            delete transformedCfnParams['stackName']
```

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { awsStackStatusRestPath } from '../../../utils/RoutePaths';
+
+const QuickStartStatusContainer = styled.div`
+    display: flex;
+`;
+
+const stageStatusFlag = (currentProvisioningStatus: string): ReactElement => {
+    const renderProvisioningStatus = (status: string) => {
+        if (status === 'CREATE_COMPLETE') {
+            return <SuccessIcon primaryColor="#36B37E" label="Success" />;
```
bpar476: We should probably be using a theme to manage these colours so we are consistent in our appearance. We can defer this, but we should create a ticket to track it.

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
```diff
@@ -0,0 +1,94 @@
...
+
+    return (
+        <Flag
+            actions={[
+                {
+                    content: 'CloudFormation Console',
+                },
+            ]}
+            icon={renderProvisioningStatus(currentProvisioningStatus)}
+            description="All good things take time. Like your next Uber Eats delivery!"
```
bpar476: 😂 Truer words have never been spoken

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
```diff
@@ -0,0 +1,94 @@
...
+            id="1"
+            key="1"
+            title="Provisioning Status"
+        />
+    );
+};
+
+export const QuickStartStatus: FunctionComponent = (props: Record<any, any>): ReactElement => {
```
bpar476: We should probably use a custom type for the props instead of `Record<any, any>`

```
type QuickstartStatusProps = {
  match: <whatever type match is>
}
```

src/main/dc-migration-assistant-fe/components/aws/AwsRoutes.tsx
```diff
@@ -17,6 +18,7 @@ export const AWSRoutes: FunctionComponent = () => (
         <Route exact path={quickstartPath}>
             <QuickStartDeploy />
         </Route>
+        <Route exact path={quickstartStatusPath} component={QuickStartStatus} />
```
bpar476: Is there a reason you used the `component={...}` syntax instead of following the children syntax?

Also, we aren't providing the props to the `QuickStartStatus` component, how does it know which stack ID to query?

As I think about this, I feel like it should be the responsibility of the backend to know which stack ID to query the stack status of. It reduces the number of network calls we need to make from the UI and keeps the state of the migration in the backend.

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
```diff
@@ -0,0 +1,94 @@
...
+    const [inProgress, setInProgress]: [boolean, Function] = useState(true);
+    const [provisioningStatus, setProvisioningStatus]: [string, Function] = useState('UNKNOWN');
+    let intervalId: any = 0;
```
bpar476: Why is this an `any` and not a `number`?


Also, the interval should probably be a part of the state if the value changes 🤔

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
```diff
@@ -0,0 +1,94 @@
...
+    };
+
+    useEffect(() => {
+        const { match } = props;
+        const stackIdentifier = match.params.stackId;
+
+        intervalId = setInterval(() => {
+            getStackStatusFromApi(stackIdentifier).then(() => {
+                console.log(
+                    'Retrieved status from API once. TODO: periodically poll this API endpoint'
```
bpar476: This is already periodically polling isn't it?

theghostwhoforks: oh yeah. The message will go

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
theghostwhoforks: moved this to `Api.ts` as an enum for now

theghostwhoforks: will use dotenv and change this to behave in an env appropriate manner

bpar476: Will dotenv work in the browser? 🤔

theghostwhoforks: https://medium.com/@trekinbami/using-environment-variables-in-react-6b0a99d83cf5

I'm trying this out

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
theghostwhoforks: yeah. Will do when the FE stories come in. I used `@atlaskit/theme` but had an issue getting that to work and didn't spend time digging in

src/main/dc-migration-assistant-fe/components/aws/AwsRoutes.tsx
theghostwhoforks: No particular reason. I saw doco on how to pass path params over and this was what was being used. The stack ID is made available through the `props.match.params` object that the status component calls.

And yeah - the call should change to `/migration/<id>/stack/status` and the migration with the identifier should be used to track the status. That way, we don't have to surface the ARN all the way up

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
theghostwhoforks: so, setInterval returns a value of type called `NodeJs.Timeout`. Once the interval is set, it does not change. I've changed the type to reflect instead of `any`

package.json
```diff
@@ -31,6 +31,7 @@
         "atlassian-webresource-webpack-plugin": "^4.6.0",
         "awesome-typescript-loader": "^5.2.1",
         "classnames": "^2.2.6",
+        "dotenv": "^8.2.0",
```
bpar476: This can be a dev dependency I believe.

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -78,22 +76,20 @@ const QuickstartForm = ({
     <Form
         onSubmit={(data: Record<string, any>): void => {
             const transformedCfnParams = data;
-            let stackNameValue = '';
+            const stackNameValue = transformedCfnParams[STACK_NAME_FIELD_NAME];
```
bpar476: Does this work? I'm pretty sure the stack name field will be in the `data` object as `stackName` I thought it used the `name` of the field to determine the key 🤔

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
bpar476: The change from unset to set should be managed by the component state

src/main/dc-migration-assistant-fe/components/aws/AwsRoutes.tsx
bpar476: We should pick one style and stick with it I reckon.

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartStatus.tsx
theghostwhoforks: Remove the props all together to use `useParams` hook instead.

https://reacttraining.com/blog/react-router-v5-1/

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
theghostwhoforks: I've hardcoded this for now to the production templates and parameters. The webpack plugin is supposed to replace the variable references in the js/ts(?) files at build time, but I wasn't able to verify this.

I'm leaving the dotenv as is (after converting to a dev dependency) and moving onwards

https://www.robertcooper.me/front-end-javascript-environment-variables

src/main/dc-migration-assistant-fe/components/aws/AwsRoutes.tsx
theghostwhoforks: Converted to use a react hook and child components 👍

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
bpar476: Could you put a comment in the webpack config around the block that isn't used? Webpack config is so hard to read already 😓

package.json
theghostwhoforks: yup. Done.

src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
bpar476: Just tested locally:
image

bpar476: See the bottom-most field

theghostwhoforks: Ah yeah - Fixed this and pushed

theghostwhoforks: Screenshot 2020-02-28 at 15 06 19

theghostwhoforks: az is not stringified as well

PR #13, opened 2020-02-25
CHET-93: Reimplement S3 Uploader Service

nanux:
* Introduce Queue for asynchronious S3 upload
* Crawler for filesystem traversal using DirectoryStream
* Use `localstack` for integration testing
* Fix the `integration` profile to run integration tests only
* Use jira-api 8.5.3 to get Jira shared home location


src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
```diff
@@ -0,0 +1,96 @@
...
+import java.util.Map;
+import java.util.Queue;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public class S3Uploader implements Uploader {
+    private static final Logger logger = LoggerFactory.getLogger(S3Uploader.class);
+    private static final int SLEEP_TO_WAIT_FOR_CRAWLER = 500; // ms
```
bpar476: /nit this could probably be MS_TO_WAIT_FOR_CRAWLER and we can dodge the comment

src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
```diff
@@ -0,0 +1,96 @@
...
+        while (!((path = queue.poll()) == null) || !isCrawlDone.get()) {
+            if (path == null) {
+                try {
+                    Thread.sleep(SLEEP_TO_WAIT_FOR_CRAWLER);
+                } catch (InterruptedException e) {
+                    logger.error("Interrupted S3 upload, adding all remaining files to failed collection");
+                    queue.forEach(p -> addFailedFile(p, e));
+                }
+            } else {
+                if (path == null) {
```
bpar476: Looks like this code was left behind from iteration

src/main/java/com/atlassian/migration/datacenter/spi/fs/FilesystemMigrationProgress.java
```diff
@@ -1,15 +1,21 @@
...
     private long bytesTransferred = 0L;
+    private Set<Path> failedFiles;
 
     public FilesystemMigrationProgress() {
         this(FilesystemMigrationStatus.NOT_STARTED);
     }
 
     public FilesystemMigrationProgress(FilesystemMigrationStatus status) {
         this.status = status;
+        this.failedFiles = new HashSet<>();
```
bpar476: This is dead code that I added when I started working on surfacing the failed files to the end user. It can probably be dropped for now

.github/workflows/mvn_test.yml
```diff
@@ -31,5 +31,4 @@ jobs:
         with:
           java-version: ${{ matrix.java }}
       - name: Test with maven
-        run: mvn -B verify
-
+        run: mvn -B clean verify
```
theghostwhoforks: It's interesting why you needed to do this? With each container being fresh, this would have been unnecessary? 

It's good to clean, but I just wanted to check in on the reasoning :)

pom.xml
```diff
@@ -123,7 +123,7 @@
         <dependency>
             <groupId>com.atlassian.jira</groupId>
             <artifactId>jira-api</artifactId>
-            <version>8.6.1</version>
+            <version>8.5.3</version>
```
theghostwhoforks: I wonder if it's a prudent thing to pin the version to the previous ER - just to ensure migrations can work with older products?

src/main/java/com/atlassian/migration/datacenter/core/fs/DirectoryStreamCrawler.java
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    private void listDirectories(ConcurrentLinkedQueue<Path> queue, DirectoryStream<Path> paths) {
+        paths.forEach(p -> {
+            if (Files.isDirectory(p)) {
+                try (final DirectoryStream<Path> newPaths = Files.newDirectoryStream(p.toAbsolutePath())) {
+                    listDirectories(queue, newPaths);
+                } catch (IOException e) {
+                    logger.error("Error when traversing directory {}, with exception {}", p, e);
+                    failedPaths.put(p.toString(), e);
```
theghostwhoforks: Is it better to hold just the message instead of the full exception?

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -1,82 +1,114 @@
...
+                .region(Region.of(regionService.getRegion()))
+                .build();
+        S3UploadConfig config = new S3UploadConfig(getS3Bucket(), s3AsyncClient, getSharedHomeDir());
+
+
+        AtomicBoolean isDoneCrawling = new AtomicBoolean(false);
+        ConcurrentLinkedQueue<Path> uploadQueue = new ConcurrentLinkedQueue<>();
+
+        Queue<Future<?>> uploadResults = new LinkedList<>();
+        ExecutorService uploadService = Executors.newFixedThreadPool(NUM_UPLOAD_THREADS);
```
theghostwhoforks: I'm thinking how much nicer this would have looked with go channels 😀

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -1,82 +1,114 @@
...
+import java.util.stream.IntStream;
+
+import static com.atlassian.migration.datacenter.spi.fs.FilesystemMigrationStatus.FAILED;
 
 @Component
 public class S3FilesystemMigrationService implements FilesystemMigrationService {
-    private static final Logger log = LoggerFactory.getLogger(S3FilesystemMigrationService.class);
+    private static final Logger logger = LoggerFactory.getLogger(S3FilesystemMigrationService.class);
+
+    public static final int NUM_UPLOAD_THREADS = 2;
```
theghostwhoforks: /nit could be an externally configurable value

`Long.getLong('....', 2)`

src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
```diff
@@ -0,0 +1,96 @@
...
+                }
+
+                if (Files.exists(path)) {
+                    String key = config.getSharedHome().relativize(path).toString();
+                    final PutObjectRequest putRequest = PutObjectRequest.builder()
+                            .bucket(config.getBucketName())
+                            .key(key)
+                            .build();
+                    final CompletableFuture<PutObjectResponse> response = config.getS3AsyncClient().putObject(putRequest, path);
+                    final S3UploadOperation uploadOperation = new S3UploadOperation(path, response);
```
theghostwhoforks: Is there a reason we need this instead of `ComposableFuture#thenapply` or `ComposableFuture#thenCompose` which returns a composable chain that we add the response failure check onto?

nanux:
@bpar476 reflected PR comments. For now, I am not adding integration tests into github actions, let's tackle this as part of another task.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -1,82 +1,114 @@
...
+
+        AtomicBoolean isDoneCrawling = new AtomicBoolean(false);
+        ConcurrentLinkedQueue<Path> uploadQueue = new ConcurrentLinkedQueue<>();
+
+        Queue<Future<?>> uploadResults = new LinkedList<>();
+        ExecutorService uploadService = Executors.newFixedThreadPool(NUM_UPLOAD_THREADS);
+        IntStream.range(0, NUM_UPLOAD_THREADS).forEach(i -> {
+            Uploader uploader = new S3Uploader(config);
+            Future<?> uploadResult = uploadService.submit(() -> {
+                uploader.upload(uploadQueue, isDoneCrawling);
```
theghostwhoforks: Similar to the comment on the interface, if the upload returns a result, we could perform some action on the result. i.e. - dump file paths in the db, retry upload etc..

src/main/java/com/atlassian/migration/datacenter/core/fs/Uploader.java
```diff
@@ -0,0 +1,11 @@
+package com.atlassian.migration.datacenter.core.fs;
+
+import com.atlassian.migration.datacenter.spi.FailableOperation;
+
+import java.nio.file.Path;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+
+public interface Uploader extends FailableOperation {
+    void upload(ConcurrentLinkedQueue<Path> queue, AtomicBoolean isCrawlDone);
```
theghostwhoforks: Should this return a `UploadResult` type that wraps over the status and list of failed files?

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
```diff
@@ -0,0 +1,61 @@
...
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.time.Instant;
+import java.util.Collections;
+import java.util.UUID;
+
+import static org.mockito.Mockito.when;
+
+@Tag("integration")
```
theghostwhoforks: Like

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
```diff
@@ -0,0 +1,61 @@
...
+@ExtendWith({LocalstackDockerExtension.class, MockitoExtension.class})
+@LocalstackDockerProperties(services = {"s3"})
+class S3FilesystemMigrationServiceIT {
+    @TempDir
+    Path dir;
+    @Mock
+    RegionService regionService;
+
+    @Mock
```
theghostwhoforks: Does this need to be mocked with localstack?

src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploadTest.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.atomic.AtomicBoolean;
+
+@Disabled("Exploratory test")
+public class S3UploadTest {
+
+    public static final String FOLDER = "/Users/abrokes/code/taskcat";
+    final Queue<CompletableFuture<PutObjectResponse>> responseQueue = new LinkedList<>();
+
+    @Test
+    void listAllFilesInDirectory() throws IOException, ExecutionException, InterruptedException {
```
theghostwhoforks: is this missing an assert?

Edit :- ignore as test is disabled

.github/workflows/mvn_test.yml
nanux: It failed for me when I've added the integration step but now, when we have only one step, I don't think we need it.

pom.xml
nanux: I think, it is a good idea, let me test it with it.

src/main/java/com/atlassian/migration/datacenter/core/fs/DirectoryStreamCrawler.java
nanux: Yes, I had also concern whether this structure might not get too large. It is a valid point but I can imagine in some cases having an exception might come handy but if there is a lot of failures than this can get too large. Let's tackle this down the line when we have some results.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
nanux: Are we switching Slingshot v3? :trollface:

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
nanux: Yup, we wanted to make it configurable but as we don't have config storage, we've skipped. As this might be handy for testing, I am going to add it with a string property for now.

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
nanux: Do you prefer this comparing to `*IT` pattern? I had the tags in the beginning, but eventually switched to the pattern.

src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploadTest.java
nanux: The whole test shouldn't be probably there, it was just exploratory test.

src/main/java/com/atlassian/migration/datacenter/core/fs/Uploader.java
nanux: Correct, we just didn't need it for now, so we've exposed less. I think we will revisit this when we will consume the results.

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
nanux: Yes, the `AtlassianPluginAWSCredentialsProvider` is our interface combining SDK v1 and v2. We can simplify this when we remove SDK v1.

PR #14, opened 2020-02-25
CHET-113: Make creation of CfnApi lazy to avoid requiring keys/region before they are available.

tarka:


tarka:
> There are two failing tests (Checks tab), what looks like we are missing credential provider as we are lazily loading the cfn API.

Yep, I'm just working out the best way to stub this.

src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
```diff
@@ -36,6 +26,7 @@ public CfnApi(CloudFormationAsyncClient cloudFormationClient) {
     }
 
     private static CloudFormationAsyncClient buildCloudFormationClient() {
+        // FIXME: This should be able to use credentials/region/etc. that are input in the UI.
```
bpar476: This can be done by injecting the `AtlassianPluginCredentialsManager` and `RegionService` into the constructor and creating the client with that credentials provider and using `regionService.getRegion()` to set the region on the client.

src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApiFactory.java
```diff
@@ -0,0 +1,31 @@
...
+     * Fetch a CfnApi implementation.
+     *
+     * FIXME: We should implement a teardown of the object once finished?
+     */
+    public CfnApi getCfnApi() {
+        if (cfnApi.isPresent()) {
+            return cfnApi.get();
+        }
+
+        // FIXME: This could potentially be a BeanFactory lookup, but we currently only have a single impl.
```
bpar476: I'm personally not a fan of `FixMe` and `TODO:` comments unless they're pending a blocker being resolved or some immediate-ish decision. Are we ever going to have more than one implementation of the CfnApi? In my opinion, we won't in which case this FIXME is perhaps not necessary.

That's just like, my opinion man though.

src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApiFactory.java
```diff
@@ -0,0 +1,31 @@
...
+ */
+@Component
+public class CfnApiFactory
+{
+    private Optional<CfnApi> cfnApi;
+
+    /**
+     * Fetch a CfnApi implementation.
+     *
+     * FIXME: We should implement a teardown of the object once finished?
```
bpar476: Does "finished" mean the migration is finished? After a quick google, it looks like the clients implement `close()` [creating java clients](https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/creating-clients.html) [SdkAutoCloseable interface](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/utils/SdkAutoCloseable.html)

To be strictly correct, we should close the client at least during plugin shut down and we could also close it when not in a migration to save resources (if that is an optimisation worth making, probably is)

src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
tarka: AFAICS, at this point we don't have all the information required to build a credential manager. The credentials and region don't exist at this point, and won't until the user provides them in the UI. At that point we need to build a Supplier with the metadata required. So this is pretty-much a place-holder for the time-being until we have functionality fleshed out.

src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApiFactory.java
tarka: Finished can mean a few things here (e.g. end of migration vs plugin/jira is restart); we really need to work out what they are, but that's not critical for the MVP.

src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApiFactory.java
tarka: If it's done through a beanfactory lookup at that point any dependent beans will be injected into the generated bean, which isn't done if we do manual instantiation.

re: FIXME/TODO, I agree, but this is very much a placeholder ATM.

tarka:
> There are two failing tests (Checks tab), what looks like we are missing credential provider as we are lazily loading the cfn API.

Fixed.

src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
bpar476: We have these classes [AtlassianPluginAWSCredentialsProvider.java](https://github.com/atlassian-labs/dc-migration-assistant/blob/master/src/main/java/com/atlassian/migration/datacenter/core/aws/auth/AtlassianPluginAWSCredentialsProvider.java) [PluginSettingsRegionManager.java](https://github.com/atlassian-labs/dc-migration-assistant/blob/master/src/main/java/com/atlassian/migration/datacenter/core/aws/region/PluginSettingsRegionManager.java)
which manage the user's chosen credentials and region. They should be set through the REST API before the user gets to the CloudFormation deployment step. If somehow they're not authenticated then the requests will fail.

tarka: Ah, I see what you mean. I'll take a look.

PR #15, opened 2020-02-26
CHET-86: Create new type to manage file migration error

bpar476:
I've refactored the file migration error handling to only store a string, not the whole exception. I also created some neater types to manage errors. This is a precursor to presenting the errors to the user by incorporating them into the FileSystemMigrationStatus.

I tried to incorporate @theghostwhoforks 's feedback about using CompletableFuture chaining into the S3Uploader but it proved too difficult to do this correctly, elegantly and while preserving all information required to construct the FileOperationError.


nanux:
How this relates to #17? Seems this PR is going in a different direction.

bpar476:
> How this relates to #17? Seems this PR is going in a different direction.

It was meant to be a precursor and a quick fix to the OOM when errors occur but you're right they've diverged now.

PR #17, opened 2020-02-26
CHET-86: handle s3 errors in file migration service

bpar476:
# Show upload progress

Reports errors, status and files uploaded in real time

## Error Reporting
* File path and error message stored

## Progress Reporting
* Number of files uploading
* couldn't figure out how to get number of bytes, looks like there is no listener facility in sdk v2

## Status
* Reports internal state of file migration

Also some light optimisations around concurrency in S3FileSystemMigrationService.

Sorry for the chongos PR.


src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -1,37 +1,35 @@
...
 import com.atlassian.migration.datacenter.core.aws.region.RegionService;
+import com.atlassian.migration.datacenter.spi.fs.FailedFileMigrationReport;
 import com.atlassian.migration.datacenter.spi.fs.FilesystemMigrationProgress;
 import com.atlassian.migration.datacenter.spi.fs.FilesystemMigrationService;
 import com.atlassian.migration.datacenter.spi.fs.FilesystemMigrationStatus;
 import com.atlassian.plugin.spring.scanner.annotation.imports.ComponentImport;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.stereotype.Component;
+import org.tuckey.web.filters.urlrewrite.Run;
```
nanux: I don't think this is needed.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -45,6 +43,10 @@
 
     private MultipleFileUpload upload;
     private FilesystemMigrationProgress progress = new FilesystemMigrationProgress(FilesystemMigrationStatus.NOT_STARTED);
+    private FailedFileMigrationReport errorReport;
```
nanux: I know it is not directly the purpose of this change, but I think down the path we could have one class encapsulating the whole FS migration operation, something like:

```java
class FileMigrationReport {
    FailedFileMigrationReport
    FileMigrationProgress -> which can encapsulate our progress metrics
}
```

src/main/java/com/atlassian/migration/datacenter/core/aws/auth/AtlassianPluginAWSCredentialsProvider.java
```diff
@@ -21,6 +21,7 @@ public AtlassianPluginAWSCredentialsProvider(ReadCredentialsService readCredenti
 
     /**
      * AWS SDK V1 credentials API
+     *
```
nanux: This will clash with the SDKv1 removal

src/main/java/com/atlassian/migration/datacenter/core/fs/DirectoryStreamCrawler.java
```diff
@@ -32,16 +34,11 @@ private void listDirectories(ConcurrentLinkedQueue<Path> queue, DirectoryStream<
                     listDirectories(queue, newPaths);
                 } catch (IOException e) {
                     logger.error("Error when traversing directory {}, with exception {}", p, e);
-                    failedPaths.put(p.toString(), e);
+                    report.reportFileNotMigrated(new FailedFileMigration(p, e.getMessage()));
```
nanux: Should we pass full exception and then on lower level decide how we want to process it (convert it to `getMessage`)?

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -54,33 +55,47 @@ public S3FilesystemMigrationService(RegionService regionService,
...
      * Start filesystem migration to S3 bucket. This is a blocking operation and should be started from ExecutorService
      * or preferably from ScheduledJob
      */
     public void startMigration() {
-        S3AsyncClient s3AsyncClient = S3AsyncClient.builder()
-                .credentialsProvider(credentialsProvider)
-                .region(Region.of(regionService.getRegion()))
-                .build();
-        S3UploadConfig config = new S3UploadConfig(getS3Bucket(), s3AsyncClient, getSharedHomeDir());
+        progress.setStatus(RUNNING);
```
nanux: good catch, I had to do the same modification on my branch :)

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
bpar476: I might tackle that once I've figured out how to get progress.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
```diff
@@ -56,18 +52,14 @@ public void upload(ConcurrentLinkedQueue<Path> queue, AtomicBoolean isCrawlDone)
 
                     responsesQueue.add(uploadOperation);
                 } else {
+
```
nanux: /nit just to align our styles, do you prefer empty lines after control statements?

src/main/java/com/atlassian/migration/datacenter/spi/fs/FailedFileMigrationReport.java
```diff
@@ -0,0 +1,52 @@
...
+import java.util.concurrent.ConcurrentLinkedQueue;
+
+/**
+ * Manages files which have had an error throughout the file migration
+ * This class is threadsafe and is intended to be added to by any agents which
+ * are a part of the file system migration
+ */
+public class FailedFileMigrationReport {
+
+    private final ConcurrentLinkedQueue<FailedFileMigration> failedMigrations;
```
nanux: Is the queue used to simplify the multi-threading lock? Is the other option to `synchronize` on `reportFileNotMigrated` ?

src/main/java/com/atlassian/migration/datacenter/spi/fs/FailedFileMigrationReport.java
```diff
@@ -0,0 +1,52 @@
...
+        this.failedMigrations = new ConcurrentLinkedQueue<>();
+    }
+
+    public void reportFileNotMigrated(FailedFileMigration failedFileMigration) {
+        failedMigrations.add(failedFileMigration);
+    }
+
+    /**
+     * @return an immutable copy of the FailedFileMigrations in this report. Note the returned value
+     * is not backed by the underlying collection so will not be updated as other producers add to it.
```
nanux: `not backed by the underlying collection` -> `just a copy of original collection`

src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploaderIT.java
```diff
@@ -23,30 +30,50 @@
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.mockito.Mockito.when;
 
 @Tag("integration")
-@ExtendWith(LocalstackDockerExtension.class)
-@LocalstackDockerProperties(services = {"s3"})
+@ExtendWith({LocalstackDockerExtension.class, MockitoExtension.class})
```
nanux: I was thinking about introducing new-meta annotation so we have the version and overall configuraiton in one place but I don't think we can pass the list of `services` from the annotation to the higher one.

```java
@Tag("integration")
@ExtendWith({LocalstackDockerExtension.class, MockitoExtension.class})
@LocalstackDockerProperties(services = {"s3"}, imageTag = "0.10.6") // I think services here is not possible so we would need to define the largest set of our tested services, not ideal
public @interface AWSIntegrationTest {
       String[] services();
}
```

src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploaderIT.java
```diff
@@ -63,6 +90,16 @@ void uploadShouldUploadPathsFromQueueToS3() throws IOException {
...
 
+        assertEquals(
+                errorReport.getFailedFiles().size(),
+                0,
+                String.format(
+                        "expected no upload errors but found %s",
+                        errorReport.getFailedFiles()
+                                .stream()
+                                .reduce("",
+                                        (acc, failedMigration) -> String.format("%s%s: %s\n", acc, failedMigration.getFilePath().toString(),
```
nanux: having 🎉 with reducers :)

README.md
```diff
@@ -16,6 +16,7 @@ Build the plugin via `mvn package` then install the plugin into your Jira instan
 * Node/NPM ([NVM](https://github.com/nvm-sh/nvm) is recommended to manage
   versions)
 * Yarn (can be installed from package or with `npm install -g yarn`)
+* [Atlassian SDK])(https://developer.atlassian.com/server/framework/atlassian-sdk/) (optional)
```
nanux: It is actually hard to recommend when we know it doesn't work with the maven version enforcer rule out of the box.

nanux:
Can you please resync with `master`? Seems there are changes which are already on master.

src/main/java/com/atlassian/migration/datacenter/core/fs/DirectoryStreamCrawler.java
bpar476: Having the full exception is definitely better for debugging purposes as you can see the stack trace but I can't really think of how to do this without running into the memory troubles again.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
bpar476: I do not, I'm not sure where this line came from 👀

src/main/java/com/atlassian/migration/datacenter/spi/fs/FailedFileMigrationReport.java
bpar476: That's correct 👍

src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploaderIT.java
bpar476: #FeelingFunctional

README.md
bpar476: Should be resolved after syncing with master

src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploaderIT.java
bpar476: So like one super annotation for integration tests which tags it, uses all extensions and localstack properties?

src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
```diff
@@ -0,0 +1,53 @@
...
+
+import java.nio.file.Path;
+import java.util.List;
+
+public class DefaultFileSystemMigrationReport implements FileSystemMigrationReport {
+
+    private final FileSystemMigrationErrorReport errorReport;
+    private final FileSystemMigrationProgress progress;
+
+    private FilesystemMigrationStatus status;
```
nanux: This will lead to NullPointerException in `S3UploadJobRunner` where we are calling:

`if (s3Service.isRunning()) { }` which calls:

```java
    public boolean isRunning() {
        return report.getStatus().equals(RUNNING);
    }
```

I think we can instantiate it with:
```
    public DefaultFileSystemMigrationReport(FileSystemMigrationErrorReport errorReport, FileSystemMigrationProgress progress) {
        this.errorReport = errorReport;
        this.progress = progress;
        this.status = FilesystemMigrationStatus.NOT_STARTED;
    }
```

src/main/java/com/atlassian/migration/datacenter/core/fs/DirectoryStreamCrawler.java
nanux: We don't need to store the full exception, just pass it to `reportFileNotMigrated` and then pick the message there. Maybe in future that would allow us to do some sort of post-processing - let's say report it to CloudWatch without changing the signature.
It was just an idea, no need to change it if you don't want to.

src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploaderIT.java
nanux: I wanted to pin the version for localstack in one place but the need to define different services for each test won't make the superannotation possible.

src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
bpar476: Yeah I noticed that happened in the tests. I've done that now and added a test to make sure everything in the reporting package is initialised properly.

PR #18, opened 2020-02-27
CHET-119: Time FS migration

bpar476:
# Summary
* Use `java.time.Instant` to capture elapsed time of FS migration
* Stop "timing" the migration when it reaches terminal state (`DONE` or `FAILED`)
* Time migration based on when migration enters `RUNNING` state from non-`RUNNING` state
* Test with mocked clock

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
@@ -125,6 +125,7 @@ public class S3FilesystemMigrationService implements FilesystemMigrationService 
             logger.error("Failed to traverse home directory for S3 transfer", e);
             report.setStatus(FAILED);
         } finally {
+            // FIXME: the uploader will continue uploading until the queue is empty even though we probably need to abort in this scenario as it's indeterminate whether all files have been uploaded or not (should we try fix this now or create a bug and follow up?)
             isDoneCrawling.set(true);
         }
     }
--- a/src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
@@ -25,12 +25,12 @@ public class DefaultFileSystemMigrationReport implements FileSystemMigrationRepo
 
     private Instant startTime;
     private Instant completeTime;
-    private FilesystemMigrationStatus status;
+    private FilesystemMigrationStatus currentStatus;
 
     public DefaultFileSystemMigrationReport(FileSystemMigrationErrorReport errorReport, FileSystemMigrationProgress progress) {
         this.errorReport = errorReport;
         this.progress = progress;
-        this.status = NOT_STARTED;
+        this.currentStatus = NOT_STARTED;
         this.clock = Clock.systemUTC();
     }
 
@@ -42,12 +42,12 @@ public class DefaultFileSystemMigrationReport implements FileSystemMigrationRepo
             completeTime = Instant.now(clock);
         }
 
-        this.status = status;
+        this.currentStatus = status;
     }
 
     @Override
     public FilesystemMigrationStatus getStatus() {
-        return status;
+        return currentStatus;
     }
 
     @Override
@@ -60,7 +60,7 @@ public class DefaultFileSystemMigrationReport implements FileSystemMigrationRepo
     }
 
     private boolean isRunning() {
-        return status == RUNNING;
+        return currentStatus == RUNNING;
     }
 
     @Override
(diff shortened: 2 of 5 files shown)


src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
```diff
@@ -6,25 +6,44 @@
...
 
+    private Instant startTime;
+    private Instant completeTime;
     private FilesystemMigrationStatus status;
 
     public DefaultFileSystemMigrationReport(FileSystemMigrationErrorReport errorReport, FileSystemMigrationProgress progress) {
         this.errorReport = errorReport;
         this.progress = progress;
         this.status = NOT_STARTED;
+        this.clock = Clock.systemUTC();
```
bpar476: This is the `Clock` used by `Instant.now()` which we will default to if not testing.

src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
```diff
@@ -52,5 +84,21 @@ public void reportFileNotMigrated(FailedFileMigration failedFileMigration) {
...
+    private boolean isStartingMigration(FilesystemMigrationStatus status) {
+        return this.status != RUNNING && status == RUNNING;
+    }
+
+    private boolean isendingMigration(FilesystemMigrationStatus status) {
+        return this.status == RUNNING && isTerminalState(status);
+    }
+
+    private boolean isTerminalState(FilesystemMigrationStatus status) {
+        return status == DONE || status == FAILED;
```
bpar476: I'm not completely sure if `FAILED` is actually a terminal state. Intuitively it should be however currently the migration continues partially with a `FAILED` state under certain circumstances (see comment in `S3FileSystemMigrationService.java` made in this PR

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -125,6 +125,7 @@ private void populateUploadQueue() {
             logger.error("Failed to traverse home directory for S3 transfer", e);
             report.setStatus(FAILED);
         } finally {
+            // FIXME: the uploader will continue uploading until the queue is empty even though we probably need to abort in this scenario as it's indeterminate whether all files have been uploaded or not (should we try fix this now or create a bug and follow up?)
```
nanux: We can empty the `uploadQueue` in the `catch` branch, is that what you have in mind?

src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
```diff
@@ -52,5 +82,21 @@ public void reportFileNotMigrated(FailedFileMigration failedFileMigration) {
     public void reportFileMigrated(Path path) {
         progress.reportFileMigrated(path);
     }
+
+    public void setClock(Clock clock) {
```
nanux: It probably matters of preference but I rather use Powermock rather than injecting own `Clock`. The problem is that Powermock seems unsupported in JUnit 5 (few open issues in their repository), so let's move forward with the setter.

src/test/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReportTest.java
```diff
@@ -73,4 +76,34 @@ void shouldDelegateToWrappedProgress() {
...
+        assertEquals(Duration.ZERO, sut.getElapsedTime());
+
+        sut.setClock(Clock.offset(testClock, Duration.ofDays(1).plusSeconds(5)));
+
+        assertEquals(1L, sut.getElapsedTime().toDays());
+        assertEquals(5L, sut.getElapsedTime().minusDays(1).getSeconds());
+    }
+
+    @Test
+    void shouldNotIncrementElapsedTimeAfterMigrationEnds() {
```
nanux: Props for the clock tests. When we have the setup, I think it would be useful to add test for `FAILED` state and also setting the `RUNNING` twice and check that we don't restart the elapsed time.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
bpar476: I hadn't thought of how to do it. That is an option :)

bpar476: I'm wondering if we should add the rest of the queue to the failed files but it's not really meaningful as there may be uncrawled files which will have "failed" but there is no way to track them so maybe just dumping the queue is the way to go

src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
bpar476: Is this so that you don't need to have the setter?

src/test/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReportTest.java
bpar476: 👍

PR #21, opened 2020-02-28
CHET-120: FS migration progress rest endpoint

bpar476:
# Summary

* Able to get full status of FS migration
    * status
    * completed files
    * failed files
* Fixes bug where migration is not set to running after starting

## Note
If there is a large number of errors then the response can get very large. I had a migration fail completely (local jira with amps:run) and the API response was 1.2MB. We may want to have a separate endpoint for the errors and implement pagination. Thoughts?

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
+++ b/src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
@@ -45,27 +45,13 @@ public class FileSystemMigrationProgressEndpoint {
 
     static class FSMigrationProgressWebObject {
         public String status;
-        public List<FailedFileMigrationWebObject> failedFileMigrations;
+        public List<FailedFileMigration> failedFileMigrations;
         public List<String> migratedFiles;
 
         FSMigrationProgressWebObject(String status, List<FailedFileMigration> failedFileMigrations, List<String> migratedFiles) {
             this.status = status;
+            this.failedFileMigrations = failedFileMigrations;
             this.migratedFiles = migratedFiles;
-
-            this.failedFileMigrations = failedFileMigrations
-                    .stream()
-                    .map(failedFileMigration -> new FailedFileMigrationWebObject(failedFileMigration.getFilePath().toString(), failedFileMigration.getReason()))
-                    .collect(Collectors.toList());
-        }
-    }
-
-    static class FailedFileMigrationWebObject {
-        public String path;
-        public String reason;
-
-        public FailedFileMigrationWebObject(String path, String reason) {
-            this.path = path;
-            this.reason = reason;
         }
     }
 }
--- a/src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
+++ b/src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
@@ -3,6 +3,11 @@ package com.atlassian.migration.datacenter.api.fs;
 import com.atlassian.migration.datacenter.spi.fs.FilesystemMigrationService;
 import com.atlassian.migration.datacenter.spi.fs.reporting.FailedFileMigration;
 import com.atlassian.migration.datacenter.spi.fs.reporting.FileSystemMigrationReport;
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonAutoDetect.Visibility;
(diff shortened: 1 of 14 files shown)


src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
```diff
@@ -0,0 +1,73 @@
...
+                report.getFailedFiles(),
+                report.getMigratedFiles().stream().map(java.nio.file.Path::toString).collect(Collectors.toList())
+        );
+
+        return Response
+                .ok(progress)
+                .build();
+    }
+
+    static class FSMigrationProgressWebObject {
```
bpar476: I tried to implement the serialization in `FileSystemMigrationReport` however Jackson was having trouble serializing the `FailedFileMigration` nested object :(

Actually, now that I think about it, it's probably because it's an inner class of an Interface. I might have another crack at putting the serialization there instead of duplicating the data structure.

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
```diff
@@ -0,0 +1,57 @@
...
+        if (report == null) {
+            return Response
+                    .status(Response.Status.BAD_REQUEST)
+                    .entity("no file system migration exists")
+                    .build();
+        }
+        final FSMigrationProgressWebObject progress = new FSMigrationProgressWebObject(
+                report.getStatus().name(),
+                report.getFailedFiles(),
+                report.getMigratedFiles().stream().map(java.nio.file.Path::toString).collect(Collectors.toList())
```
bpar476: Only return the first X

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
bpar476: Fixed! Now all the interfaces and the FailedFileMigration type are directly serializable

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
bpar476: I've implemented code to drop attempted reports if the list has 100 errors already. It's likely these migrations will be failure scenarios anyway and that lost errors have the same cause.

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
```diff
@@ -0,0 +1,50 @@
...
+    public Response getFilesystemMigrationStatus() {
+        FileSystemMigrationReport report = migrationService.getReport();
+        if (report == null) {
+            return Response
+                    .status(Response.Status.BAD_REQUEST)
+                    .entity("no file system migration exists")
+                    .build();
+        }
+        ObjectMapper mapper = new ObjectMapper();
+        mapper.setVisibility(PropertyAccessor.ALL, Visibility.ANY);
```
theghostwhoforks: It may be useful to build the object mapper as a spring component and inject it in - we will end up reusing this going forward, I think?

Also, I'm still peeved that we have to do this and not get a properly configured object mapper instance injected in - checking how similar plugins are built should address this?

src/main/java/com/atlassian/migration/datacenter/spi/fs/reporting/FileSystemMigrationProgress.java
```diff
@@ -1,11 +1,14 @@
...
 
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
+
 import java.nio.file.Path;
 import java.util.List;
 
 /**
  * Tracks the progress of the file system migration
  */
+@JsonSerialize(as = FileSystemMigrationProgress.class)
```
theghostwhoforks: Is JSON serialize needed here?

bpar476: Yes this is needed so that it is serialized as the interface and not the implementing class.

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpoint.java
bpar476: I feel like there must be some way to configure it somewhere 🤔 Let's have this as a follow-up if it comes up again.

PR #24, opened 2020-02-28
CHET-112: Additional logging

nanux:


pom.xml
```diff
@@ -229,7 +229,12 @@
             <version>1.11.728</version>
             <scope>test</scope>
         </dependency>
-
+        <dependency>
```
nanux: I find `hamcrest` generally useful, but I am happy to hear other opinions.

src/test/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReportTest.java
```diff
@@ -126,4 +130,24 @@ void shouldNotIncrementElapsedTimeAfterMigrationEnds(FilesystemMigrationStatus s
         sut.setClock(Clock.offset(testClock, Duration.ofSeconds(20)));
         assertEquals(10L, sut.getElapsedTime().getSeconds());
     }
+
+    @Test
+    void testToString() {
+        final List migratedList = mock(List.class);
+        final int successfullyMigrated = 888;
+        final int failedFiles = 666;
+        when(migratedList.size()).thenReturn(successfullyMigrated);
```
nanux: This might be unnecessary `mockery`...

bpar476:
Would it be worthwhile to list the directories found by the crawler? It could potentially be massive 🤔 but could help debug if certain paths aren't uploaded

nanux:
> Would it be worthwhile to list the directories found by the crawler? It could potentially be massive 🤔 but could help debug if certain paths aren't uploaded

For now, I've decided to not go that path as we were able to identify the missing files.

PR #27, opened 2020-03-03
No issue general improvements

theghostwhoforks:
1. Adds checked exceptions and sends appropriate response codes to client
2. Pulls object mapper instantiation to constructor

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
@@ -95,7 +95,7 @@ public class AWSMigrationService implements MigrationService {
      * {@inheritDoc}
      */
     @Override
-    public String provisionInfrastructure(ProvisioningConfig config) {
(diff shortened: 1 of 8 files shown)


theghostwhoforks:
The tests are failing because the origin branch name was incorrect. They can be ignored

src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
```diff
@@ -95,7 +95,7 @@ private Migration getMigration() {
      * {@inheritDoc}
      */
     @Override
-    public String provisionInfrastructure(ProvisioningConfig config) {
+    public String provisionInfrastructure(ProvisioningConfig config) throws InvalidMigrationStageError, InfrastructureProvisioningError {
```
tarka: Have we decided that we're doing checked exceptions?

theghostwhoforks: The options arr either using an optional with a None Type implementation that holds the error (seems familiar, yes) or using checked exceptions.

Let's get this in and change the service layer later if the optional approach gives us more flexibility?

PR #28, opened 2020-03-03
CHET-128: Cloudformation State Candidate Implementation

bpar476:
# Summary

This PR has a candidate implementation for the State pattern. It shows how we could implement the following state sequence

`STACK_FORM -> STACK_DEPLOY -> MIGRATION_STACK_DEPLOY`
and
`STACK_FORM -> STACK_DEPLOY -> ERROR`

The assumed flow with this implementation is:

1. User has filled out the quick start form
2. User clicks "Deploy"
3. UI calls out to `MigrationService` and requests to move to transition to the next state. It includes the stack parameters in the request
4. The stack parameters are added to a `MigrationContext`
5. The `STACK_FORM` state checks whether it is able to transition to the next stage
6. The `STACK_FORM` state approves the transition and sets the state on the migration service
to `STACK_DEPLOY`
7. The `MigrationService` calls `run()` on the new `State` (asynchronously)
8. The request from (3) returns to the UI with the new state as the response body
9. The UI renders the appropriate screen for `STACK_DEPLOY`
10. -problem- how do we get feedback on the stack deployment here without leaking abstraction into the state machine

This is more of a request for feedback than a true contribution. After wrapping my head around it for the better part of a day, I'm not sold that the state pattern is actually best for us. Here are my key concerns

* It is difficult to get user-input into the state machine and maintain type-safety
* The actions of each state are quite different, this makes it difficult to have one State interface
    * Some do something right away
    * Some are polling states
    * Some do nothing and wait for a user to submit something
* It is difficult to get progress when the only thing the user interacts with is the state machine (you don't want the UI to interact directly with the currently acting service as this goes around the state management)
* It would be near-impossible to re-use the state machine for any purpose other than migrating to AWS.

# Alternative
I think perhaps it would be best to leave the state pattern behind and try to avoid such a generic approach. Instead of the entire process being managed by a StateMachineService, each component of the migration is managed by its own service (i.e. our current implementation). This would be backed up by a very lightweight state management system which exists purely to ensure all actions are taken in the correct order.

Consider the previous example with this different architecture.

1. The user has filled out the quick start form
2. User clicks "Deploy"
3. UI calls out to `QuickStartService` and requests to provision a new stack
4. The `QuickStartService` checks the `MigrationState` is equal to `STACK_FORM` (the only state where a quick start can be deployed)
5. The `QuickStartService` initiates deployment.
6. The `QuickStartService` sets the `MigrationState` to `STACK_DEPLOY`
7. The request from (3) returns
8. The UI goes to the next page: waiting for stack to finish deploying
9. The UI polls the stack status from the backend until the migration is complete
10. Once the migration is completed, the backend sets its state to `MIGRATION_STACK_DEPLOY`
...

The downside of this approach is that the state machine is less strictly enforced. Services have to check with the state machine that they can run rather than the state machine calling the services when it is their turn.
The benefit is each service is completely independent of the others. It doesn't matter what the implementation of each stage of the migration is, it only matters that there is an implementation for that state. There is no enforcing on when a state transition can occur (either automatically or by user action).

I'm personally more fond of the alternative approach as it has a greatly reduced complexity at the cost of reduced type-safety and a bit of boiler-plate (which can probably be eliminated with inheritance).

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
@@ -55,6 +55,11 @@ public class CfnApi {
         return client;
     }
 
+    /**
+     * Gets the status of the supplied stack. If there is no stack with this name a runtime exception will be
+     * thrown.
+     * @param stackName the name of the stack the get the status of
+     */
     public StackStatus getStatus(String stackName) {
         Optional<Stack> stack = getStack(stackName);
         if (!stack.isPresent()) {
--- a/src/main/java/com/atlassian/migration/datacenter/core/statemachine/AWSMigrationContext.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/statemachine/AWSMigrationContext.java
@@ -1,0 +1,47 @@
+package com.atlassian.migration.datacenter.core.statemachine;
+
+import net.java.ao.Entity;
+
+import java.util.Map;
+
+public interface AWSMigrationContext extends Entity {
+
+    /**
+     * AppStackName is the property used to determine the name of the AWS Cloudformation Stack
+     * used to run the target application
+     * @return the AppStackName
+     */
+    String getAppStackName();
+
+    /**
+     * @see AWSMigrationContext#getAppStackName()
+     */
+    void setAppStackName(String appStackName);
+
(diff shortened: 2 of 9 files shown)


bpar476:
Closing so can be opened as draft

PR #30, opened 2020-03-03
CHET-128: Implement alternative state management strategy

bpar476:
This is the alternative strategy outlined in: #29 

> I think perhaps it would be best to leave the state pattern behind and try to avoid such a generic approach. Instead of the entire process being managed by a StateMachineService, each component of the migration is managed by its own service (i.e. our current implementation). This would be backed up by a very lightweight state management system which exists purely to ensure all actions are taken in the correct order.
Consider the previous example with this different architecture.

    The user has filled out the quick start form
    User clicks "Deploy"
    UI calls out to QuickStartService and requests to provision a new stack
    The QuickStartService checks the MigrationState is equal to STACK_FORM (the only state where a quick start can be deployed)
    The QuickStartService initiates deployment.
    The QuickStartService sets the MigrationState to STACK_DEPLOY
    The request from (3) returns
    The UI goes to the next page: waiting for the stack to finish deploying
    The UI polls the stack status from the backend until the migration is complete
    Once the migration is completed, the backend sets its state to MIGRATION_STACK_DEPLOY

> The downside of this approach is that the state machine is less strictly enforced. Services have to check with the state machine that they can run rather than the state machine calling the services when it is their turn.
The benefit is each service is completely independent of the others. It doesn't matter what the implementation of each stage of the migration is, it only matters that there is an implementation for that state. There is no enforcing on when a state transition can occur (either automatically or by user action).

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/api/ProvisionAppEndpoint.java
+++ b/src/main/java/com/atlassian/migration/datacenter/api/ProvisionAppEndpoint.java
@@ -1,0 +1,31 @@
+package com.atlassian.migration.datacenter.api;
+
+import com.atlassian.migration.datacenter.core.exceptions.InvalidMigrationStageError;
+import com.atlassian.migration.datacenter.spi.infrastructure.AppProvisioningService;
+import com.atlassian.migration.datacenter.spi.infrastructure.ProvisioningConfig;
+
+import javax.ws.rs.Consumes;
+import javax.ws.rs.POST;
+import javax.ws.rs.Path;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+
+@Path("/migration/provision/app")
+public class ProvisionAppEndpoint {
+
+    AppProvisioningService appProvisioningService;
+
+    @POST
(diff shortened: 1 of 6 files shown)


bpar476:
This is not intended to be merged, it has no tests and hasn't even been executed. It serves as a demonstration of a state implementation.

tarka:
This seem like a better approach; I'd prefer the simplest possible thing that works. At this point we should assume that anything we do will be heavily changed over the course of development, so we shouldn't get too bogged down in abstract structures and patterns.

nanux:
As commented in the channel, I like flexibility and lower overhead of this approach. Each of the stage is responsible to surface the restrictions under which it can run. I just wonder if we should provide a "dry run" validation method. This method will be used in the actual run of the stage but it also allows consumers to validate whether the stage can be transition into without actually running it (might be useful for frontend). But maybe it is also in the bucket with other YAGNI.

```
public StageReport runStage() {
     if (!isValid()) {
        throws new InvalidTransition()
     }
     runstage...
}

public boolean isValid() {
      validatePrerequisites()
}
```

bpar476:
Closing because this was a demonstration PR

PR #32, opened 2020-03-03
CHET-71 migrating in EncryptedCredentialsStorage to main line

andrewpowell:
Changed encryption to AES (Spring Crypto std) and stored encryption key as read-only file on file system that is readable only by the process running Jira.


src/test/java/com/atlassian/migration/datacenter/core/aws/auth/EncryptedCredentialsStorageTest.java
```diff
@@ -0,0 +1,53 @@
...
+import org.junit.jupiter.api.Test;
+
+import java.lang.reflect.InvocationTargetException;
+import java.lang.reflect.Method;
+import java.util.Random;
+
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+
+public class EncryptedCredentialsStorageTest {
```
bpar476: Love the test! Are we able to test that the key is actually persisted to the file system and that calling `getSecretAccessKey()` after `setSecretAccessKey(string)` returns the correct string? (and the same for AccessKeyId methods)

pom.xml
```diff
@@ -137,6 +137,17 @@
             <version>1.0.1b</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.security</groupId>
```
nanux: We have `bouncycastle` library provided by the system OSGi bundle and I believe it was the recommendation from our security team to use it for encryption. I don't think it should block the PR as we want to move forward but if this significantly increase the plugin footprint we might reconsider it.

pom.xml
```diff
@@ -281,6 +281,12 @@
                             com.atlassian.jira.*;resolution:="optional",
                             com.atlassian.analytics.api.annotations;resolution:="optional",
                             com.atlassian.confluence.*;resolution:="optional",
+                            com.sun.tools.javac.*;resolution:="optional",
```
bpar476: What do all these imports do and why can they have optional resolution?

andrewpowell: these are downstream compile dependencies

bpar476: If they're compile dependencies then how come they're optional resolution? Going to push this through anyway but keen to understand.

PR #35, opened 2020-03-04
CHET-136: Add support for finding and parsing dbconfig.xml.

tarka:


src/main/java/com/atlassian/migration/datacenter/core/application/JiraConfiguration.java
```diff
@@ -0,0 +1,90 @@
...
+
+            return val;
+
+        } catch (Exception e) {
+            throw new ConfigurationReadException("Failed to parse dbconfig.xml, XPath: "+path, e);
+        }
+    }
+
+    private DatabaseConfiguration parseDbConfig() throws ConfigurationReadException
+    {
```
bpar476: I believe the password can be Base64 "encrypted" in the DB config.xml. There is a property in the XML which is set if this is true. We probably need to handle that case.

On further inspection, it looks like there are multiple supported methods for [encrypting the database password](https://confluence.atlassian.com/adminjiraserver/encrypting-database-password-974378811.html). It may be too difficult to decrypt the database password so perhaps we should throw an exception if we see a cipher element in the `dbconfig.xml` and give them the chance to put the password in via UI?

tarka: Useful info, thank. For MVP I'll implement the basic ones and throw an exception for the rest, we can add something more advanced after that.

src/main/java/com/atlassian/migration/datacenter/core/application/ApplicationConfiguration.java
```diff
@@ -0,0 +1,27 @@
+package com.atlassian.migration.datacenter.core.application;
+
+/**
```
bpar476: Do we need copyrights in all our classes? Is there a policy somewhere?

tarka: Intellij adds it automatically, but it's common practice; even opensource projects do it.

PR #36, opened 2020-03-04
CHET-133: Add multipart upload

nanux:
S3 PUT uploads has a hard limit of 5GB, for anything bigger we need to run multipart upload. This was the reason why we were missing 150GBs of data - all the backups from Confluence weren't in the bucket.

https://docs.aws.amazon.com/AmazonS3/latest/dev/UploadingObjects.html

This PR:
* Add multipart upload for files larger than 5GBsmo
* Merge filesystem API under one roof (not literally)
* Create a limit on the size of the UploadOperation queue - unbounded I was hitting the limit of the open HTTP clients from netty async client from AWS SDK 
* Add more defensive code to prevent starting the filesystem migration when it is already running
* Add more logging and javadocs
* Have integration tests


src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.java
```diff
@@ -1,35 +1,59 @@
...
+        } else {
+            boolean started = migrationService.startFilesystemMigration();
+            response = Response.status(Response.Status.ACCEPTED).entity("S3 Migration started = " + started).build();
+        }
+        return response;
+    }
+
     @GET
     @Produces(APPLICATION_JSON)
+    @Path("/report")
```
nanux: Changes are reflected in shared Postman

src/main/java/com/atlassian/migration/datacenter/core/fs/S3MultiPartUploader.java
```diff
@@ -0,0 +1,108 @@
...
+    private List<CompletedPart> completedParts;
+
+    public S3MultiPartUploader(S3UploadConfig config) {
+        this.config = config;
+    }
+
+    /**
+     * Upload the file in multiple parts
+     * @param file file to upload
+     * @throws ExecutionException
```
bpar476: This is missing an `@param` for the key

bpar476: I'm not fussed about having javadoc or not as this method is quite self-documenting, however if we are going to have an @param we should probably have it for both

src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
```diff
@@ -20,6 +20,8 @@
 public class S3Uploader implements Uploader {
     private static final Logger logger = LoggerFactory.getLogger(S3Uploader.class);
     private static final int MS_TO_WAIT_FOR_CRAWLER = 500;
+    private static final int MAX_OPEN_CONNECTIONS = 50;
```
bpar476: Is this the limitation of the Netty async http client? Is it possible we're going to need this value in other places?

src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
```diff
@@ -43,9 +45,20 @@ public void upload(ConcurrentLinkedQueue<Path> queue, AtomicBoolean isCrawlDone)
                     logger.error("Interrupted S3 upload, adding all remaining files to failed collection");
                     queue.forEach(p -> addFailedFile(p, e.getMessage()));
                 }
+            } else if (responsesQueue.size() >= MAX_OPEN_CONNECTIONS) {
+                responsesQueue.forEach(this::handlePutObjectResponse);
             } else {
                 if (Files.exists(path)) {
                     String key = config.getSharedHome().relativize(path).toString();
+                    if (path.toFile().length() > MAXIMUM_FILE_SIZE_TO_UPLOAD) {
+                        logger.debug("File {} is larger than 5 GBs, running multipart upload", path);
```
bpar476: should the "5 GBs" string in the debug message be generated from the `MAXIUMUM_FILE_SIZE_TO_UPLOAD` constant?

src/main/java/com/atlassian/migration/datacenter/core/fs/S3MultiPartUploader.java
nanux: We should be consistent. I am not using the safe rename always and I've missed this.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3MultiPartUploader.java
```diff
@@ -0,0 +1,109 @@
...
+
+                partNo++;
+            }
+        } catch (IOException e) {
+            logger.error("Cannot open file for the multi-part upload", e);
+        }
+
+        completeUpload(key, uploadId).get();
+        logger.debug("Finished multipart upload for {} with {} parts", key, completedParts.size());
+    }
```
andrewpowell: It is recommend that you enable the AbortIncompleteMultipartUpload lifecycle rule on the S3 bucket so there's not abandoned parts hanging around on an aborted upload.

This rule directs Amazon S3 to abort multipart uploads that don’t complete within a specified number of days after being initiated. When the set time limit is exceeded, Amazon S3 aborts the upload and then deletes the incomplete upload data.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3MultiPartUploader.java
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Upload the file in multiple parts
+     * @param file file to upload
+     * @param key S3 key for the file
+     * @throws ExecutionException
+     * @throws InterruptedException
+     */
+    public void multiPartUpload(File file, String key) throws ExecutionException, InterruptedException {
```
andrewpowell: Configure this to take a bucket name as well so it can be re-used for the Liquibase file upload.

PR #39, opened 2020-03-05
CHET-124: implement state management service

bpar476:
Paired with @theghostwhoforks 

This is the new state management service based on the specification from #30 

There is just one distinction and that is that there is no ability to set the stage directly.
Services either tell the state machine to move to the next stage (once they are done). The next stage is determined by the current stage. Services can also tell the state machine to move to an Error state.

The only pitfall with this approach is recovering from the Error state as it can potentially have multiple next states that aren't known until runtime, depending on the recovery approach. I suspect we may need to revert to setting the stage directly.
Should we defer that decision to when we start handling the unhappy path?

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
@@ -177,7 +177,11 @@ public class AWSMigrationService implements MigrationService, MigrationServiceV2
     }
 
     @Override
-    public void setCurrentStage(MigrationStage stage) {
+    public void nextStage() {
+        setCurrentStage(getCurrentStage().getNext());
+    }
+
+    private void setCurrentStage(MigrationStage stage) {
         migration.setStage(stage);
         migration.save();
     }
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
@@ -42,7 +42,6 @@ public class AWSMigrationService implements MigrationService, MigrationServiceV2
     private final SchedulerService schedulerService;
     private final CfnApi cfnApi;
     private ActiveObjects ao;
-    private Migration migration;
 
     /**
      * Creates a new, unstarted AWS Migration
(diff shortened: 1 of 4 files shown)


src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
```diff
@@ -24,19 +25,23 @@
...
+import static com.atlassian.migration.datacenter.spi.MigrationStage.PROVISION_APPLICATION;
+import static com.atlassian.migration.datacenter.spi.MigrationStage.WAIT_PROVISION_APPLICATION;
 import static java.util.Objects.requireNonNull;
 
 /**
  * Manages a migration from on-premise to self-hosted AWS.
  */
 @Component
-public class AWSMigrationService implements MigrationService {
+public class AWSMigrationService implements MigrationService, MigrationServiceV2 {
```
bpar476: This is a transition period. As we bring the individual services into this new state model we can remove the old MigrationService and drop the `V2` suffic.

src/main/java/com/atlassian/migration/datacenter/spi/MigrationStage.java
```diff
@@ -4,26 +4,32 @@
...
-    AUTHENTICATION("authentication"),
-    PROVISION_APPLICATION("provision_app"), WAIT_PROVISION_APPLICATION("wait_provision_app"),
-    PROVISION_MIGRATION_STACK("provision_migration"), WAIT_PROVISION_MIGRATION_STACK("wait_provision_migration"),
-    READY_FS_MIGRATION("ready_fs_migration"), FS_MIGRATION_EXPORT("fs_migration_up"), FS_MIGRATION_IMPORT("fs_migration_down"),
-    OFFLINE_WARNING("cutover_warning"),
-    DB_MIGRATION_EXPORT("db_migration_up"), DB_MIGRATION_IMPORT("db_migration_down"),
-    VALIDATE("validate"),
-    CUTOVER("cutover"),
-    FINISHED("finished");
+    ERROR("error", null),
```
bpar476: These. need to be declared in reverse order otherwise, the next value will not be defined in time 🤦‍♂

PR #40, opened 2020-03-06
CHET-128: Bring quickstart deployment into state machine

bpar476:
This was paired with @theghostwhoforks 

This implements a service which is dedicated to deploying the application stack. It deploys the stack, watches its status and transitions the state machine when the deployment has completed

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
@@ -1,12 +1,15 @@
 package com.atlassian.migration.datacenter.core.aws.infrastructure;
 
+import com.atlassian.activeobjects.external.ActiveObjects;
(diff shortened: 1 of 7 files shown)


src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
```diff
@@ -0,0 +1,107 @@
...
+
+        switch (status) {
+            case CREATE_COMPLETE:
+                return ApplicationDeploymentStatus.CREATE_COMPLETE;
+            case CREATE_FAILED:
+                return ApplicationDeploymentStatus.CREATE_FAILED;
+            case CREATE_IN_PROGRESS:
+                return ApplicationDeploymentStatus.CREATE_IN_PROGRESS;
+            default:
+                throw new RuntimeException("Unexpected stack status");
```
bpar476: Should we do a stack cleanup if we hit this case?

bpar476:
Being re-worked

PR #44, opened 2020-03-06
CHET-131: Bring AWS Configuration into state machine

bpar476:
# Summary

* Implement service for configuring AWS connection
* Remove old API endpoint for authentication and region setting
* Implement transitions based on authentication
* Remove old plain text credentials manager


src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/AWSConfigurationService.java
```diff
@@ -0,0 +1,50 @@
...
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component
+public class AWSConfigurationService implements CloudConfigurationService {
+    private static final Logger logger = LoggerFactory.getLogger(AWSConfigurationService.class);
+
+    private final WriteCredentialsService writeCredentialsService;
+    private final RegionService regionService;
+    private final MigrationServiceV2 migrationService;
```
nanux: What is the plan with `MigrationService`, depracate and then rename `MigrationServiceV2` to `MigrationService`?

src/main/java/com/atlassian/migration/datacenter/spi/cloud/CloudConfigurationService.java
```diff
@@ -0,0 +1,15 @@
+package com.atlassian.migration.datacenter.spi.cloud;
+
+import com.atlassian.migration.datacenter.core.exceptions.InvalidMigrationStageError;
+
+public interface CloudConfigurationService {
+
+    /**
+     * Configures access to a cloud provider to facilitate the migration
+     * @param entity an identifier for the entity accessing the cloud provider e.g. IAM user
```
nanux: I think this is a bit misleading. The entity, in this case, is access key ID. I am not sure if it is even possible to have IAM user & secret.

src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/AWSConfigurationService.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public AWSConfigurationService(WriteCredentialsService writeCredentialsService, RegionService regionService, MigrationServiceV2 migrationService) {
+        this.writeCredentialsService = writeCredentialsService;
+        this.regionService = regionService;
+        this.migrationService = migrationService;
+    }
+
+    @Override
+    public void configureCloudProvider(String entity, String secret, String geography) throws InvalidMigrationStageError {
```
nanux: I think these warrants own Javadoc describing the `secret` and `entity` implementation.

src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/AWSConfigurationService.java
bpar476: Yes, do you see any issues with this approach?

src/main/java/com/atlassian/migration/datacenter/spi/cloud/CloudConfigurationService.java
bpar476: I've updated the example in the javadoc to access key ID.

src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/AWSConfigurationService.java
bpar476: I've added javadoc stating exactly what each parameter value should be.

src/main/java/com/atlassian/migration/datacenter/api/aws/TestAWSCredentialsEndpoint.java
```diff
@@ -15,31 +15,19 @@
 import static javax.ws.rs.core.MediaType.APPLICATION_JSON;
 
 @Path("aws/credentials")
-public class AWSCredentialsEndpoint {
+public class TestAWSCredentialsEndpoint {
```
theghostwhoforks: Is this class just temporary? Do we see us using this in the plugin flow?

If yes (to the latter), then then we should drop the test name (replace with Probe AwsCredentials), perhaps?

src/main/java/com/atlassian/migration/datacenter/spi/cloud/CloudConfigurationService.java
```diff
@@ -0,0 +1,15 @@
+package com.atlassian.migration.datacenter.spi.cloud;
+
+import com.atlassian.migration.datacenter.core.exceptions.InvalidMigrationStageError;
+
+public interface CloudConfigurationService {
```
theghostwhoforks: /s/CloudConfigurationService/CloudProviderConfigurationService

src/main/java/com/atlassian/migration/datacenter/api/aws/TestAWSCredentialsEndpoint.java
bpar476: This is mostly for developer convenience as an easy way to test if you're authenticated properly. I can drop it if we think it's not useful.

src/main/java/com/atlassian/migration/datacenter/spi/cloud/CloudConfigurationService.java
bpar476: Done :)

src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/AWSConfigurationService.java
nanux: No :)

PR #47, opened 2020-03-09
Chet 127 fs migration state machine

theghostwhoforks:
**This is in DRAFT state - I created the PR without selecting the draft option**

Comments are welcome


src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -49,15 +52,16 @@
...
+                                        @ComponentImport JiraHome jiraHome, MigrationServiceV2 migrationService) {
         this.regionService = regionService;
         this.credentialsProvider = credentialsProvider;
         this.jiraHome = jiraHome;
+        this.migrationService = migrationService;
     }
 
     @Override
     public boolean isRunning() {
-        return report != null && report.getStatus().equals(RUNNING);
```
theghostwhoforks: We don't need the report to tell us if the state machine is running. The state machine is should give us this information

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -87,11 +90,14 @@ public void startMigration() {
...
     }
 
-    private void initialiseMigration() {
+    private void initialiseMigration() throws InvalidMigrationStageError {
         report = new DefaultFileSystemMigrationReport(new DefaultFileSystemMigrationErrorReport(), new DefaultFilesystemMigrationProgress());
-        report.setStatus(RUNNING);
         isDoneCrawling = new AtomicBoolean(false);
         uploadQueue = new ConcurrentLinkedQueue<>();
+
+        report.setStatus(RUNNING);
```
theghostwhoforks: TODO: If we move start and complete time to migration context, we do not need the report to setStatus

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -104,16 +110,15 @@ private S3AsyncClient buildS3Client() {
...
+        CompletionService<Void> completionService = new ExecutorCompletionService<>(uploadExecutorService);
 
         Runnable uploaderFunction = () -> {
             Uploader uploader = new S3Uploader(config, report, report);
             uploader.upload(uploadQueue, isDoneCrawling);
         };
 
-        Collections.nCopies(NUM_UPLOAD_THREADS, uploaderFunction)
-                .forEach(runnable -> completionService.submit(runnable, null));
+        IntStream.range(0, NUM_UPLOAD_THREADS).forEach((_x) -> completionService.submit(uploaderFunction, null));
```
nanux: Why `_x` and parentheses?

theghostwhoforks: Question - Do we need `nCopies` call at all? 

From the docs -
> Returns an immutable list consisting of <tt>n</tt> copies of the
     specified object.  The newly allocated data object is tiny (it contains
     a single reference to the data object).  This method is useful in
     combination with the <tt>List.addAll</tt> method to grow lists...


In the case if using 

`List<Runnable> runnables = Collections.nCopies(NUM_UPLOAD_THREADS, uploaderFunction)`
`runnables` each will contain the same reference to `uploaderFunction`

In which case, using an IntStream is good enough for us as we construct the `S3Uploader` inside the runnable. 

What am I missing here?

src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
```diff
@@ -30,15 +30,15 @@
...
         this.config = config;
+        this.report = report;
         this.progress = progress;
     }
 
     @Override
     public void upload(ConcurrentLinkedQueue<Path> queue, AtomicBoolean isCrawlDone) {
         Path path;
-        while (!((path = queue.poll()) == null) || !isCrawlDone.get()) {
+        while ((path = queue.poll()) != null || !isCrawlDone.get()) {
```
theghostwhoforks: The negations tripped me a little so I simplified it a bit

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
```diff
@@ -37,6 +40,10 @@
     @Mock
     AtlassianPluginAWSCredentialsProvider credentialsProvider;
 
+    //TODO: Review why we use mocks in integration tests.
```
theghostwhoforks: TODO: Integration tests should be using fully formed objects instead of mocks. This needs to be fixed

src/main/java/com/atlassian/migration/datacenter/core/fs/S3UploadJobRunner.java
```diff
@@ -35,10 +36,15 @@ public JobRunnerResponse runJob(JobRunnerRequest jobRunnerRequest) {
         }
 
         log.info("Starting S3 migration job");
-        fsMigrationService.startMigration();
+        try {
+            fsMigrationService.startMigration();
+        } catch (InvalidMigrationStageError e) {
+            log.error("Invalid migration error - {}", e.getMessage());
```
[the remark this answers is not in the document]
theghostwhoforks: I'll create a static factory for this

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
bpar476: It was just a concise way to get access to the stream API. I thought it would be less confusing than using an IntStream where you're discarding the argument. It's probably down to personal preference.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3UploadJobRunner.java
```diff
@@ -35,10 +36,15 @@ public JobRunnerResponse runJob(JobRunnerRequest jobRunnerRequest) {
...
+            return JobRunnerResponse.failed("Encountered an invalid migration stage");
```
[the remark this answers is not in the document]
theghostwhoforks: `failed` takes in a throwable. I'll pass that along so that the logs have a detailed exception trace

src/main/java/com/atlassian/migration/datacenter/spi/MigrationStage.java
```diff
@@ -10,7 +10,7 @@
     VALIDATE("validate"),
     DB_MIGRATION_IMPORT("db_migration_down"), DB_MIGRATION_EXPORT("db_migration_up"),
     OFFLINE_WARNING("cutover_warning"),
-    FS_MIGRATION_IMPORT("fs_migration_down"), FS_MIGRATION_EXPORT("fs_migration_up"), READY_FS_MIGRATION("ready_fs_migration"),
+    WAIT_FS_MIGRATION_COPY("wait_fs_migration_copy"), FS_MIGRATION_COPY("fs_migration_copy"),
```
bpar476: The import state and the export state were created separately so that the upload to S3 and the download to EFS would be independent. Do we want to combine those states?

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceTest.java
```diff
@@ -3,9 +3,14 @@
 import com.atlassian.jira.config.util.JiraHome;
```
bpar476: Should there be a test case for when the current state is not FS_MIGRATION_COPY?

src/main/java/com/atlassian/migration/datacenter/spi/MigrationStage.java
theghostwhoforks: From our conversation last week, I remember us speaking about how the `up` and `down` stages will happen concurrently and we can't have the state machine be in 2 stages at the same time. As a result of this I created a `copy` stage instead of using explicit `import` and `export` stages.

Thoughts?

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceTest.java
theghostwhoforks: Yes. There should. Will create one and resolve

src/main/java/com/atlassian/migration/datacenter/spi/MigrationStage.java
bpar476: I think at this stage the up and down are going to happen sequentially. It's probably not a big deal at this point and will likely change again once we attempt to solve the S3 -> EFS story.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
theghostwhoforks: removed

theghostwhoforks:
I'll merge this in and implement a small refactor to clean up how we wire up the configuration objects and raise a new PR.

PR #48, opened 2020-03-09
CHET-142: implement different modes for state machine enforcement

bpar476:
This is a proof of concept for how we could have multiple modes for the migration assistant.

Default - enforce the current stage is as expected before running the operation
no-verify - always run the migration operation regardless of the current stage. This lets you test the database migration stage without having to wait to have moved through a file system migration.
passthrough - doesn't run the operation and sets the stage to the specified short circuit stage. This is good for if you just want to click through the UI.

I'm not completely happy with this approach as it warps how you write tests for migration operations in a couple of ways:

* Your tests need to have a ModalMigrationStageWorker
* When you create a new operation as a part of the migration you need to remember to call the worker so that it does the mode check.

I wanted to use a template method approach at first however there would need to be an abstract class doing this logic for each migration operation so there is still a bit of pain every time you implement a new migration operation.

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/ModalMigrationStageWorker.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/ModalMigrationStageWorker.java
@@ -1,0 +1,60 @@
+package com.atlassian.migration.datacenter.core;
+
+import com.atlassian.migration.datacenter.core.exceptions.InvalidMigrationStageError;
+import com.atlassian.migration.datacenter.spi.MigrationServiceV2;
+import com.atlassian.migration.datacenter.spi.MigrationStage;
+import com.atlassian.plugin.spring.scanner.annotation.imports.ComponentImport;
+import com.atlassian.sal.api.pluginsettings.PluginSettings;
+import com.atlassian.sal.api.pluginsettings.PluginSettingsFactory;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.stereotype.Component;
+
+@Component
+public class ModalMigrationStageWorker {
+
+    private static final Logger logger = LoggerFactory.getLogger(ModalMigrationStageWorker.class);
+    private static final String MIGRATION_MODE_PLUGIN_SETTINGS_KEY = "com.atlassian.migration.datacenter.core.mode";
+
+    private final MigrationServiceV2 migrationService;
+    private final PluginSettingsFactory pluginSettingsFactory;
+
+    public ModalMigrationStageWorker(@ComponentImport PluginSettingsFactory pluginSettingsFactory, MigrationServiceV2 migrationService) {
+        this.pluginSettingsFactory = pluginSettingsFactory;
+        this.migrationService = migrationService;
+    }
+
+    /**
+     * Determines whether a migration operation should be run given the current mode the app is in.
+     * If the app is in the default mode, it will verify the current stage the expected current stage and run the migration operation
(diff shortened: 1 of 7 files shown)


src/main/java/com/atlassian/migration/datacenter/core/ModalMigrationStageWorker.java
```diff
@@ -0,0 +1,97 @@
...
+
+        if (mode == null || mode.equals(DCMigrationAssistantMode.DEFAULT)) {
+            if (migrationService.getCurrentStage().equals(expectedCurrentStage)) {
+                operation.migrate();
+            }
+            return;
+        }
+
+        if (mode.equals(DCMigrationAssistantMode.PASSTHROUGH)) {
+            try {
```
bpar476: If an operation is meant to have side-effects we may need to change this from a simple transition to another operation which fills the migration with dummy data. We are already missing the application stack ID in the migration context after this.

bpar476:
This PR is a draft until I get feedback on the approach.

theghostwhoforks:
Now that I've seen a bit more of `AwsMigrationService`, I wonder if it makes sense to harvest the work done in this PR and inject the behaviour into the service based on a spring profile, say?

If we can apply an advice if a profile is set that will ignore the error that transition will raise, wouldn't that, in effect, give us the same effect as using the worker as described above?

Alternately, Getting the advice injected dynamically can be done by refreshing the bean based on a new (or edit of) plugin property

bpar476:
> Now that I've seen a bit more of `AwsMigrationService`, I wonder if it makes sense to harvest the work done in this PR and inject the behaviour into the service based on a spring profile, say?
> 
> If we can apply an advice if a profile is set that will ignore the error that transition will raise, wouldn't that, in effect, give us the same effect as using the worker as described above?
> 
> Alternately, Getting the advice injected dynamically can be done by refreshing the bean based on a new (or edit of) plugin property

I'm all for implementing an alternative solution. This one isn't very Java-y. However, being able to set the mode through the UI when in development seems important to me. I'm not sure how you do this with Spring as I'm not deeply familiar. Keen to see what you mean

src/main/java/com/atlassian/migration/datacenter/core/ModalMigrationStageWorker.java
theghostwhoforks: > > Now that I've seen a bit more of `AwsMigrationService`, I wonder if it makes sense to harvest the work done in this PR and inject the behaviour into the service based on a spring profile, say?
> > If we can apply an advice if a profile is set that will ignore the error that transition will raise, wouldn't that, in effect, give us the same effect as using the worker as described above?
> > Alternately, Getting the advice injected dynamically can be done by refreshing the bean based on a new (or edit of) plugin property
> 
> I'm all for implementing an alternative solution. This one isn't very Java-y. However, being able to set the mode through the UI when in development seems important to me. I'm not sure how you do this with Spring as I'm not deeply familiar. Keen to see what you mean

Do we need to expose this to customers via the UI or is this only for us (devs) to help with testing

nanux:
Following up our IRL conversation I think the solution is trying to cover two cases.
# Add ability to jump into a stage that we need to be to test certain stages
# Add NoOp transition state machine that can be useful to go through when working on frontend.

I think at the moment we need the 1. and we should consider doing 2. on the frontend as this change has a quite big impact on production code to provide utility testing functionality. I would suggest introducing `setStage()` endpoint that would set a stage and would be exposed in developer mode and consider adding the dummy transition machine when we start working on frontend and potentially stub it in Typescript instead of backend.

PR #49, opened 2020-03-09
CHET-144: pg_dump implementation and testing

tarka:


src/main/java/com/atlassian/migration/datacenter/core/aws/db/PostgresMigration.java
```diff
@@ -0,0 +1,104 @@
...
+        }
+        return Optional.empty();
+    }
+
+    /**
+     * Invoke `pg_dump` against the database details store in the supplied ApplicationConfiguration. Some important notes:
+     *
+     * <ul>
+     * <li>It is the responsibility of the caller to ensure that the filesystems the target resides on has sufficient space.</li>
+     * <li>stdout is redirected to the stdout of the calling process.</li>
```
bpar476: Is stdout of the process not redirected to the output file? 🤔

tarka: Oops, I meant stderr. I'll fix.

PR #51, opened 2020-03-10
CHET-143: Make sure we are uploading the full file with MultiPartUpload

nanux:


src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
```diff
@@ -22,7 +22,7 @@
     private static final Logger logger = LoggerFactory.getLogger(S3Uploader.class);
     private static final int MS_TO_WAIT_FOR_CRAWLER = 500;
     private static final int MAX_OPEN_CONNECTIONS = 50;
-    private static final long MAXIMUM_FILE_SIZE_TO_UPLOAD = 5 * 1024 * 1024 * 1024L; // 5GB  https://docs.aws.amazon.com/AmazonS3/latest/dev/UploadingObjects.html
+    private static final long MAXIMUM_FILE_SIZE_TO_UPLOAD = 50 * 1024 * 1024L; // 5GB  https://docs.aws.amazon.com/AmazonS3/latest/dev/UploadingObjects.html
```
tarka: Is this in bytes? I think you were right the first time.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3MultiPartUploader.java
```diff
@@ -119,6 +99,45 @@ public void setSizeToUpload(int sizeToUpload) {
...
+        UploadPartRequest uploadPartRequest = UploadPartRequest.builder()
+                .bucket(config.getBucketName())
+                .key(key)
+                .uploadId(uploadId)
+                .partNumber(uploadPartNumber)
+                .build();
+
+        AsyncRequestBody body;
+        if (readBytes < buffer.limit()) {
+            body = AsyncRequestBody.fromByteBuffer((ByteBuffer) buffer.limit(readBytes));
```
bpar476: What's the effect of not doing this?

src/main/java/com/atlassian/migration/datacenter/core/fs/S3Uploader.java
nanux: Good catch, this is leftover from local testing - waiting for 5GB uploads would give me more than just grey hair :)

src/main/java/com/atlassian/migration/datacenter/core/fs/S3MultiPartUploader.java
nanux: Funny you should have asked. This is the root cause for https://aws-partner.atlassian.net/browse/CHET-143

The way how the byte buffer works is that first, we allocate the size of the expected chunk. When we are reading the file, we read N bytes. If the amount of bytes read is smaller than the allocated size, the rest of the `ByteBuffer` is still allocated. When we are sending this as AsyncRequestBody, we always sent the full size of the `ByteBuffer` => resulting in rounding up the files always to the closest `sizeToUpload` (100MB in our case).

The reason why I didn't see it before was that always generated random file with a size N * 100MB :)

I have added `S3MultiPartUploaderTest.lastChunkShouldHaveExactRemainingSize()` to cover this scenario.

nanux: I've added a comment describing the scenario.

PR #52, opened 2020-03-10
CHET-154: Build thin wrapper over SSM API

bpar476:
# Summary

* Manages SSM client
* Exposes key SSM operations
* Reduced complexity of return types where possible
* Builds in sensible defaults.

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
@@ -4,7 +4,6 @@ import com.atlassian.migration.datacenter.core.aws.region.RegionService;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 import software.amazon.awssdk.regions.Region;
 import software.amazon.awssdk.services.ssm.SsmClient;
-import software.amazon.awssdk.services.ssm.model.Command;
 import software.amazon.awssdk.services.ssm.model.GetCommandInvocationRequest;
 import software.amazon.awssdk.services.ssm.model.GetCommandInvocationResponse;
 import software.amazon.awssdk.services.ssm.model.SendCommandRequest;
@@ -12,7 +11,6 @@ import software.amazon.awssdk.services.ssm.model.SendCommandResponse;
 
 import java.util.List;
 import java.util.Map;
-import java.util.function.Consumer;
 
 public class SSMApi {
 
@@ -45,8 +43,8 @@ public class SSMApi {
                 .timeoutSeconds(600)
                 .comment("command run by Jira DC Migration Assistant")
                 // FIXME: Pending migration stack
-                .outputS3BucketName("migration-bucket")
-                .outputS3KeyPrefix("fs-copy-down-log")
+                .outputS3BucketName(System.getProperty("ssmDocumentLoggingBucket", "migration-bucket"))
+                .outputS3KeyPrefix("trebuchet-ssm-document-logs")
                 // END FIXME
                 .build();
 
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
@@ -1,4 +1,32 @@
 package com.atlassian.migration.datacenter.core.aws;
 
+import com.atlassian.migration.datacenter.core.aws.region.RegionService;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.ssm.SsmClient;
+
 public class SSMApi {
+
+    private final AwsCredentialsProvider credentialsService;
+    private final RegionService regionService;
+
+    // This may be null, use getClient to safely get a non-null SSM client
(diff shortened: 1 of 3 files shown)


src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
```diff
@@ -0,0 +1,94 @@
...
+    public String runSSMDocument(String documentName, String targetEc2InstanceId, Map<String, List<String>> commandParameters) {
+        SendCommandRequest request = SendCommandRequest.builder()
+                .documentName(documentName)
+                .documentVersion("$LATEST")
+                .instanceIds(targetEc2InstanceId)
+                .parameters(commandParameters)
+                .timeoutSeconds(600)
+                .comment("command run by Jira DC Migration Assistant")
+                // FIXME: Pending migration stack
+                .outputS3BucketName("migration-bucket")
```
andrewpowell: outputS3BucketName should be unique

src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
```diff
@@ -0,0 +1,94 @@
...
+     *                          is a plain string then the list only have one element in it.
+     * @return the command ID of the invoked command.
+     */
+    public String runSSMDocument(String documentName, String targetEc2InstanceId, Map<String, List<String>> commandParameters) {
+        SendCommandRequest request = SendCommandRequest.builder()
+                .documentName(documentName)
+                .documentVersion("$LATEST")
+                .instanceIds(targetEc2InstanceId)
+                .parameters(commandParameters)
+                .timeoutSeconds(600)
```
nanux: I think we will need to have the timeout configurable (and higher probably)

src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
nanux: I believe the intention here is to replace it with a proper S3 bucket name when we have the migration stack ready. We can replace it with a variable comming from system property with default for now.

src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
bpar476: This is unfortunate naming on behalf of the AWS API. This timeout parameter is the timeout for the invocation to reach the machine, not for the actual document being invoked. i.e. if the EC2 instance doesn't receive the document within `timeoutSeconds` the command will be marked as failed. The actual runtime of the command depends on the document itself.

Is this still a problem you think?

src/test/java/com/atlassian/migration/datacenter/core/aws/SSMApiIT.java
```diff
@@ -0,0 +1,147 @@
...
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.List;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+
+@Tag("integration")
+//@ExtendWith({LocalstackDockerExtension.class})
+//@LocalstackDockerProperties(services = {"cloudformation", "s3", "ssm"}, imageTag = "0.10.8")
```
nanux: It means that we cannot test SSM with Localstack?

src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
```diff
@@ -0,0 +1,94 @@
...
+                .comment("command run by Jira DC Migration Assistant")
+                // FIXME: Pending migration stack
+                .outputS3BucketName("migration-bucket")
+                .outputS3KeyPrefix("fs-copy-down-log")
+                // END FIXME
+                .build();
+
+        SendCommandResponse response = getClient().sendCommand(request);
+
+        return response.command().commandId();
```
nanux: What will happen if there is an error executing the command?

src/test/java/com/atlassian/migration/datacenter/core/aws/SSMApiIT.java
bpar476: Unfortunately sendCommand appears to be unsupported by localstack. It gives 500 error responses.

GetCommandInvocation works however no commands exist so it's kinda useless.

src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
bpar476: The command ID will still exist. If an error has occurred, you won't find out until you check the command status with `getSSMCommand`

src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
nanux: Nope, it is not a problem. I should have considered this might be the case as SSM document has own timeout as well.

src/main/java/com/atlassian/migration/datacenter/core/aws/SSMApi.java
nanux: That might be worth metioning in the javadoc but not crucial.

PR #53, opened 2020-03-10
Chet 141 migration service cleanup

theghostwhoforks:
1. Changes configuration from spring scanner to pure java osgi config - https://developer.atlassian.com/server/framework/atlassian-sdk/spring-java-config/
2. Starts moving away from `MigationService` to `MigartionServiceV2`
3. An empty interface `MigrationService` is still retain to assist in reviews that the new service is being worked on
4. Fixes Cyclic reference between File system service and Migration service

I've verified the plugin start and accessed the Cfn and migration endpoints. Could you please help with testing the file system endpoint @nanux?


pom.xml
```diff
@@ -78,18 +78,29 @@
             <version>1.1.1</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
```
nanux: Interesting, @theghostwhoforks can you provide some context how we are going to use components going forward? Are we providing our bean configuration to `MigrationAssistantImportConfiguration`?

Is this the recommended approach from Andrew Swan?

pom.xml
```diff
@@ -78,18 +78,29 @@
...
-            <artifactId>atlassian-spring-scanner-annotation</artifactId>
-            <version>${atlassian.spring.scanner.version}</version>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context</artifactId>
+            <version>${spring.version}</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>com.atlassian.plugins</groupId>
+            <artifactId>atlassian-plugins-osgi-javaconfig</artifactId>
```
nanux: Is there any impact on the OSGi components and the amount we need to import/exclude?

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
```diff
@@ -0,0 +1,22 @@
+package com.atlassian.migration.datacenter.configuration;
+
+import com.atlassian.migration.datacenter.core.aws.region.RegionService;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.s3.S3AsyncClient;
+
+@Configuration
```
nanux: Plain 🌸 Spring 🌸, nice.

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantImportConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+import com.atlassian.sal.api.pluginsettings.PluginSettingsFactory;
+import com.atlassian.scheduler.SchedulerService;
+import com.atlassian.soy.renderer.SoyTemplateRenderer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import static com.atlassian.plugins.osgi.javaconfig.OsgiServices.importOsgiService;
+
+@Configuration
+public class MigrationAssistantImportConfiguration {
```
nanux: Would it be overreaching to call this `MigrationAssistantOsgiImportConfiguration` to make the purpose more explicit?

src/main/java/com/atlassian/migration/datacenter/core/exceptions/MigrationAlreadyExistsException.java
```diff
@@ -0,0 +1,7 @@
+package com.atlassian.migration.datacenter.core.exceptions;
+
+public class MigrationAlreadyExistsException extends Exception {
```
nanux: Right now we don't really use Exceptions to signal errors, do you think we should start adding Exceptions to each of the service that can fail?

pom.xml
theghostwhoforks: This is based in using [java config](https://developer.atlassian.com/server/framework/atlassian-sdk/spring-java-config/) to define bean definitions and import/export osgi services and expose them as spring beans for easy injection, all in code.

The eventual end state of `MigrationAssistantImportConfiguration` would contain (be importing) all relevant bean definitions required for the plugin to run.

This approach is being used by the auditing plugin and seems to be the new kid on the block for building plugins and interfacing with the OSGI layer

pom.xml
theghostwhoforks: This plugin is required to add the following imports in bnd

```     
org.springframework.cglib.core
org.springframework.cglib.proxy
org.springframework.cglib.reflect
```

[Alternate approach](https://developer.atlassian.com/server/framework/atlassian-sdk/spring-java-config-packages/) would be to explicitly import the above 3 packages, which is a bit brittle. 

There are no new exclusions that we need to add AFAIK.

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantImportConfiguration.java
theghostwhoforks: Yes! Will do

src/main/java/com/atlassian/migration/datacenter/core/exceptions/MigrationAlreadyExistsException.java
theghostwhoforks: We've started adding checked exceptions as part of the `MigrationService` refactoring and catching exceptions and handling errors explicitly in the the rest endpoints. (helps us return a relevant response code instead of 500).

I think it's a decent thing to start adding. It's either that or we use an Optional with an Error placeholder object. Returning just an Optional<Migration> would not be enough, in case we need to add behaviour based on the type of error being raised.

src/main/java/com/atlassian/migration/datacenter/api/MigrationEndpoint.java
```diff
@@ -19,9 +21,9 @@
 @Path("/migration")
 public class MigrationEndpoint {
 
-    private MigrationService migrationService;
+    private MigrationServiceV2 migrationService;
```
theghostwhoforks: V2 service will be renamed to migration service post review. This is retained for clarity

src/main/java/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.java
```diff
@@ -36,27 +34,29 @@ public CloudFormationEndpoint(MigrationService migrationService) {
...
+            return Response
+                    .status(Response.Status.CONFLICT)
+                    .entity(ImmutableMap.of("error", e.getMessage()))
+                    .build();
         }
     }
 
     @GET
-    @Path("/{stackId}/status")
+    @Path("/status")
```
theghostwhoforks: stackId will be retrieved from context now

src/main/java/com/atlassian/migration/datacenter/core/aws/auth/EncryptedCredentialsStorage.java
```diff
@@ -43,8 +42,8 @@
     private PluginSettings pluginSettings;
 
     @Inject
-    public EncryptedCredentialsStorage(@ComponentImport PluginSettingsFactory pluginSettingsFactory,
-                                       @ComponentImport JiraHome jiraHome) {
+    public EncryptedCredentialsStorage(PluginSettingsFactory pluginSettingsFactory,
```
theghostwhoforks: `ComponentImport` is no longer required

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -44,19 +51,23 @@
...
     private final MigrationServiceV2 migrationService;
+    private final SchedulerService schedulerService;
 
     private FileSystemMigrationReport report;
     private AtomicBoolean isDoneCrawling;
     private ConcurrentLinkedQueue<Path> uploadQueue;
-    private S3UploadConfig config;
+    private S3UploadConfig s3UploadConfig;
 
+    //TODO: Region Service and provider will be replaced by the S3 Client
```
theghostwhoforks: Will be resolved separately - This PR is already bloated

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -69,6 +80,49 @@ public FileSystemMigrationReport getReport() {
         return report;
     }
 
+    @Override
+    public Boolean scheduleMigration() {
```
theghostwhoforks: Code has been pulled in from the previous implementation in `AwsMigrationService` with a couple of additions (hooks into the `MigrationService` to call `transition` and `error`)

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -69,6 +80,49 @@ public FileSystemMigrationReport getReport() {
...
+        if (currentMigration.getStage() == NOT_STARTED) {
+            return false;
+        }
+
+        final JobRunnerKey runnerKey = JobRunnerKey.of(S3UploadJobRunner.KEY);
+        JobId jobId = JobId.of(S3UploadJobRunner.KEY + currentMigration.getID());
+        logger.info("Starting filesystem migration");
+
+        //TODO: Can the job runner be injected? It has no state
+        schedulerService.registerJobRunner(runnerKey, new S3UploadJobRunner(this));
```
theghostwhoforks: The `FileSystemService` is passed to the S3UploadJobRunner that will call `startMigration`

src/test/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationProgressEndpointTest.java
```diff
@@ -134,4 +135,28 @@ void shouldReturnBadRequestWhenNoReportExists() {
         assertEquals(Response.Status.BAD_REQUEST.getStatusCode(), response.getStatus());
         assertThat(response.getEntity().toString(), containsString("no file system migration exists"));
     }
+
```
theghostwhoforks: Added tests that now covers the runFileMigration method

src/test/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationServiceTest.java
```diff
@@ -54,130 +49,24 @@
...
+    public void shouldBeInNotStartedStageWhenNoMigrationsExist() {
         setupEntities();
-
-        MigrationStage initialStage = sut.getMigrationStage();
-
+        MigrationStage initialStage = sut.getCurrentStage();
         assertEquals(NOT_STARTED, initialStage);
     }
 
-    @Test
```
theghostwhoforks: Most of these tests have a new home in either -

1. `ApplicationDeploymentService`
2. `FileSystemMigrationService`

pom.xml
```diff
@@ -78,18 +78,29 @@
             <version>1.1.1</version>
             <scope>provided</scope>
         </dependency>
+        <dependency>
+            <groupId>org.apache.felix</groupId>
+            <artifactId>org.apache.felix.framework</artifactId>
+            <version>4.0.0</version>
```
nanux: Should we also move this to property?

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -69,6 +80,49 @@ public FileSystemMigrationReport getReport() {
         return report;
     }
 
+    @Override
+    public Boolean scheduleMigration() {
+        Migration currentMigration = this.migrationService.getCurrentMigration();
```
bpar476: I don't think we should be referencing the migration directly here. This should go through `MigrationService.getCurrentStage()`

src/test/java/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpointTest.java
```diff
@@ -29,34 +30,47 @@
     public void shouldAcceptRequestToProvisionCloudFormationStack() throws Exception {
         ProvisioningConfig provisioningConfig = new ProvisioningConfig("url", "stack-name", new HashMap<>());
 
-        String stackId = "stack-id";
-        when(this.migrationService.provisionInfrastructure(provisioningConfig)).thenReturn(stackId);
+        Mockito.doNothing().when(deploymentService).deployApplication(provisioningConfig.getStackName(), provisioningConfig.getParams());
```
bpar476: doNothing is the default behaviour for `void` methods of Mocks so this isn't required. Unless you feel the verbosity is necessary 🤔

src/test/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationServiceTest.java
```diff
@@ -20,23 +19,19 @@
...
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertNotNull;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.Mockito.never;
 import static org.mockito.Mockito.verify;
-import static org.mockito.Mockito.when;
+
+;
```
bpar476: Bonus semicolon?

src/main/java/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.java
```diff
@@ -36,27 +34,29 @@ public CloudFormationEndpoint(MigrationService migrationService) {
...
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
-    public Response getInfrastructureStatus(@PathParam("stackId") String stackId) {
-        Optional<String> status = migrationService.getInfrastructureProvisioningStatus(stackId);
-        if (status.isPresent()) {
-            return Response.status(Response.Status.OK).entity(status.get()).build();
+    public Response getInfrastructureStatus() {
+        try {
+            ApplicationDeploymentService.ApplicationDeploymentStatus status = deploymentService.getDeploymentStatus();
+            return Response.status(Response.Status.OK).entity(ImmutableMap.of("status", status)).build();
```
nanux: could be `Response.ok(entity).build()`

src/main/java/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.java
nanux: Will this be always the QS stack or we are planning to reuse this for migration stack as well?

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantOsgiImportConfiguration.java
```diff
@@ -0,0 +1,51 @@
...
+import com.atlassian.sal.api.pluginsettings.PluginSettingsFactory;
+import com.atlassian.scheduler.SchedulerService;
+import com.atlassian.soy.renderer.SoyTemplateRenderer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import static com.atlassian.plugins.osgi.javaconfig.OsgiServices.importOsgiService;
+
+@Configuration
+public class MigrationAssistantOsgiImportConfiguration {
```
nanux: Can we add javadoc explaining when to add to this `@Configuration` and link  to https://developer.atlassian.com/server/framework/atlassian-sdk/spring-java-config-spring-scanner/ ?

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
```diff
@@ -0,0 +1,22 @@
...
+public class MigrationAssistantBeanConfiguration {
+
+    @Bean
```
nanux: Was thinking wheater we need it eagerly or we can use `@Lazily` but it is premature optimisation.

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.java
```diff
@@ -40,20 +37,17 @@ public FileSystemMigrationEndpoint(MigrationService migrationService, Filesystem
...
                     .build();
-        } else {
-            boolean started = migrationService.startFilesystemMigration();
-            response = Response
-                    .status(Response.Status.ACCEPTED)
-                    .entity(ImmutableMap.of("migrationScheduled", started))
-                    .build();
         }
-        return response;
+        boolean started = fsMigrationService.scheduleMigration();
```
nanux: It would be good to rework the service so it throws an exception is something goes wrong (and we can print the error). I can do it later in my PR if that is to many changes in yours.

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantSpringConfiguration.java
```diff
@@ -0,0 +1,12 @@
+package com.atlassian.migration.datacenter.configuration;
+
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+@Configuration
+@Import({
```
nanux: Do we still need this when all the classes have `@Configuration` annotation and spring will scan this package?

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -69,6 +80,49 @@ public FileSystemMigrationReport getReport() {
...
+
+        final JobRunnerKey runnerKey = JobRunnerKey.of(S3UploadJobRunner.KEY);
+        JobId jobId = JobId.of(S3UploadJobRunner.KEY + currentMigration.getID());
+        logger.info("Starting filesystem migration");
+
+        //TODO: Can the job runner be injected? It has no state
+        schedulerService.registerJobRunner(runnerKey, new S3UploadJobRunner(this));
+        logger.info("Registered new job runner for S3");
+
+//      TODO: don't schedule when the job is running, there are few edge cases we need to cover
```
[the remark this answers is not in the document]
nanux: We can get rid of it, if we have a check for `FS_MIGRATION_COPY` stage, we shouldn't hit this code anyway.

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -69,6 +80,49 @@ public FileSystemMigrationReport getReport() {
...
+        JobConfig jobConfig = JobConfig.forJobRunnerKey(runnerKey)
+                .withSchedule(null) // run now
+                .withRunMode(RunMode.RUN_ONCE_PER_CLUSTER);
+        try {
+            logger.info("Scheduling new job for S3 upload runner");
+
+            this.migrationService.transition(MigrationStage.FS_MIGRATION_COPY, MigrationStage.WAIT_FS_MIGRATION_COPY);
+
+            schedulerService.scheduleJob(jobId, jobConfig);
+        } catch (SchedulerServiceException | InvalidMigrationStageError e) {
```
nanux: I think we can throw the `InvalidMigrationStageError` exception from here. We can also create `FilesystemMigrationException` to wrap any other errors errors.

src/main/java/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.java
bpar476: In my mind, there should be a separate endpoint for each stack. I don't want to bloat this PR with more changes as it's quite big already but I think there should be an `ApplicationDeploymentEndpoint` and a `MigrationHelperDeploymentEndpoint`. The caller shouldn't need to know any identifiers of the stacks.

src/test/java/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpointTest.java
```diff
@@ -29,34 +30,47 @@
...
+        assertEquals(provisioningConfig.getStackName(), response.getEntity());
     }
 
     @Test
     public void shouldBeConflictWhenCurrentMigrationStageIsNotValid() throws Exception {
         ProvisioningConfig provisioningConfig = new ProvisioningConfig("url", "stack-name", new HashMap<>());
 
-        when(this.migrationService.provisionInfrastructure(provisioningConfig)).thenThrow(new InvalidMigrationStageError("error"));
+        String errorMessage = "migration status is FUBAR";
+        Mockito.doThrow(new InvalidMigrationStageError(errorMessage)).when(deploymentService).deployApplication(provisioningConfig.getStackName(), provisioningConfig.getParams());
```
nanux: /nit for consistency we can use static import.

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
theghostwhoforks: Yeah - we need to build the S3 client and inject it into the spring container. There's a separate branch & PR for that piece of work

src/main/java/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.java
theghostwhoforks: To summarize, 3 options here -

1. Create 2 methods to handle the 2 types of stack that we have
2. Create 2 endpoints to handle the 2 types of stack
3. Have a generic endpoint, but the client passes in the stack ID.

I think it's reasonable to expect the client to store some state - either a migration identifier or a stack identifier. If the client stores the migration identifier, the rest path fragment could look like - /rest/1.0/dc-migration/migrations/<id>/aws/(deploymentStack|migrationStack)/status

Currently, the ID lookup is implicit as we load the migrations and associate the first found migration with the request.

Let's create a task in the migration template story to tackle this?

src/main/java/com/atlassian/migration/datacenter/core/aws/auth/EncryptedCredentialsStorage.java
theghostwhoforks: Yeah, we shouldn't. Let me test and update this PR

pom.xml
theghostwhoforks: Yeah - Will do

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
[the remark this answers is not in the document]
theghostwhoforks: will remove

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -69,6 +80,49 @@ public FileSystemMigrationReport getReport() {
...
+        //TODO: Review why this check is needed?
+        if (currentMigration.getStage() == NOT_STARTED) {
+            return false;
+        }
+
+        final JobRunnerKey runnerKey = JobRunnerKey.of(S3UploadJobRunner.KEY);
+        JobId jobId = JobId.of(S3UploadJobRunner.KEY + currentMigration.getID());
+        logger.info("Starting filesystem migration");
+
+        //TODO: Can the job runner be injected? It has no state
```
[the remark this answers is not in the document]
theghostwhoforks: From a spring perspective this isn't a circular deoendency - as the `FSMigrationService` constructs the `JobRunner` and passes itself to it.

While this isn't ideal (I think), changing this would require a new abstraction layer between the service and the jobRunner and wouldn't be too useful right now, IMO.

And yeah - If we were to inject the `S3UploadJobRunner`, this would become a circular dependency 🤦‍♂

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -69,6 +80,49 @@ public FileSystemMigrationReport getReport() {
         return report;
     }
 
+    @Override
+    public Boolean scheduleMigration() {
+        Migration currentMigration = this.migrationService.getCurrentMigration();
+        //TODO: Review why this check is needed?
+        if (currentMigration.getStage() == NOT_STARTED) {
```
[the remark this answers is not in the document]
theghostwhoforks: Yes, this check seemed out of place and will update it now!

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
theghostwhoforks: The only reason this is here is because the job ID requires the migration ID.

src/main/java/com/atlassian/migration/datacenter/core/aws/auth/EncryptedCredentialsStorage.java
theghostwhoforks: I'll make a note of this and remove this later. I won't be fixing this right away, but clean this up along with the component annotations

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.java
```diff
@@ -40,20 +37,17 @@ public FileSystemMigrationEndpoint(MigrationService migrationService, Filesystem
...
+        return Response
+                .status(Response.Status.ACCEPTED)
```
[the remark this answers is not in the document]
theghostwhoforks: It responds with accepted,  and sends the status in the entity response.

If this were to change, what response status should we use?

src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
```diff
@@ -39,157 +22,37 @@
 @Component
 public class AWSMigrationService implements MigrationService, MigrationServiceV2 {
```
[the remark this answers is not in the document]
theghostwhoforks: I'll remove this at the end of the review - Wanted to keep MigrationServiceV2  to show where the new APIs are being used

src/main/java/com/atlassian/migration/datacenter/spi/MigrationServiceV2.java
```diff
@@ -9,15 +10,23 @@
...
     /**
      * Gets the current stage of the migration
      */
     MigrationStage getCurrentStage();
 
+
+    /**
+     * Gets the Migration Object
+     */
+    Migration getCurrentMigration();
```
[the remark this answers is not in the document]
theghostwhoforks: It's used by the S3FileSystemMigrationService to retrieve the current migration. It uses the migration ID while scheduling the job.

src/test/java/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpointTest.java
theghostwhoforks: Yeah - I'll change this to a verify instead

src/main/java/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.java
bpar476: I disagree, I don't think the client should store the stack ID. It really is none of the client's business. I agree let's tackle it as a part of the migration stack.

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.java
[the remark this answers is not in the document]
bpar476: I would think `CONFLICT`. It's the pattern we've been going for so far when an illegal call is made. From a client perspective, I don't think you want a 200 level response when your intended action has failed.

src/main/java/com/atlassian/migration/datacenter/spi/MigrationServiceV2.java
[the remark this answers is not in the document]
bpar476: Is that really necessary? Can we use a random hash as the job ID? Having global access to the migration allows for setting the stage without going through the service which we should avoid if possible.

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantSpringConfiguration.java
theghostwhoforks: Eventually, the recommended approach is to get rid of the component scan and point the xml file to just one bean definition class file. 

You're right in spotting this; it's just that we'll need to get rid of the component scan soon.

src/main/java/com/atlassian/migration/datacenter/spi/MigrationServiceV2.java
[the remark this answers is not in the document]
theghostwhoforks: We can get the service to return just a DTO instead. Having a scheduler run a  job with a well known ID will be useful to tie the job run with a migration.

The DTO can store just the ID and the stage for now.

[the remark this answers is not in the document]
bpar476: Okay, let's have that as a follow-up ticket. Keen to push this through 😤

theghostwhoforks:
The migration service is dead. Long live the migration service!

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.java
theghostwhoforks: As discussed IRL, this will be done separately - Thanks!

src/main/java/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.java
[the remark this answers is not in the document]
theghostwhoforks: Done!

PR #55, opened 2020-03-10
Merged Beanstalk to ASG with LaunchConfig

stephensalim-aws:
Deleted Elastic Beanstalk
Replaced with ASG + LaunchConfig
Added CloudFormation Init Script to Bootstrap mounting of EFS.
Custom Resource to attach Helper Security Group to EFS Security Group.

PR #56, opened 2020-03-11
Chet 161 - S3 notifications to SQS for SSM

andrewpowell:
Setup of S3 notifications from S3 via SQS.

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/fs/restore/RestorationConfigurationManager.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/fs/restore/RestorationConfigurationManager.java
@@ -1,0 +1,42 @@
+package com.atlassian.migration.datacenter.core.fs.restore;
+
+import org.springframework.stereotype.Service;
+import software.amazon.awssdk.arns.Arn;
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.GetBucketNotificationConfigurationRequest;
+import software.amazon.awssdk.services.s3.model.GetBucketNotificationConfigurationResponse;
+import software.amazon.awssdk.services.s3.model.NotificationConfiguration;
+import software.amazon.awssdk.services.s3.model.QueueConfiguration;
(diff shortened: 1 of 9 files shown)


src/main/java/com/atlassian/migration/datacenter/core/fs/restore/RestorationConfigurationManager.java
```diff
@@ -0,0 +1,107 @@
...
+    private final SqsClient sqsClient;
+    private final IamClient iamClient;
+
+    public RestorationConfigurationManager(S3Client s3Client, SqsClient sqsClient, IamClient iamClient) {
+        this.s3Client = s3Client;
+        this.sqsClient = sqsClient;
+        this.iamClient = iamClient;
+    }
+
+    public void configureS3Bucket(String bucketName) {
```
nanux: I think this should be part of the migration template that is creating the S3 bucket.

nanux:
@andrewpowell as this is more of PoC, it would be good to raise this as a draft PR as we probably won't merge it for MVP. With this piece of work, we would like to explore the technical feasibility and potential cost required for the migration.

The meat of this PR is the `RestorationConfigurationManager` which configure S3 bucket and create IAM policies. I believe that should be part of the CloudFormation template for the migration stack instead of doing in Java. I think the same is possible to say about the newly introduced SQS queue. We have the migration stack here:
https://github.com/atlassian-labs/dc-migration-assistant/blob/master/templates/migration-stack/migration-helper.yml
and there is a pending PR:
https://github.com/atlassian-labs/dc-migration-assistant/pull/55/files
Resources created in this PR can go to the template.

As this is a spike, you don't necessarily need to create tests if it would be slowing you down.

/nit we can probably rename the branch to `spike-CHET-161-Restore-Files-To-EFS-via-SSM`

bpar476:
I agree on putting the creation of resources in CloudFormation and the main reason is that it makes it easy to tear down the resources after they are no longer required and that it is easy to share references to the resources via stack outputs

PR #57, opened 2020-03-12
CHET-156: Implement SSM script and surrounding CFN to download shared home from migration bucket

bpar476:
# Summary
* Create a Jira user on the migration stack
* Create an SSM document that will run the s3 sync
* Create log files to log the s3 sync progress


bpar476:
End result:
image

bpar476:
> Will this run before the migration starts so that the bucket is available for data?

Yep, the intention is to deploy the migration stack immediately after the application stack.

PR #58, opened 2020-03-12
CHET-145: Add database migration service

tarka:
This add the DatabaseMigrationService and testing. As discussed IRL, this is slightly messy as the S3Uploader, etc. has slightly FS-migration-specific; the next step is to refactor-out the common areas into common components, but this is better as a 2-step process.

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/application/DatabaseConfiguration.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/application/DatabaseConfiguration.java
@@ -5,19 +5,28 @@ package com.atlassian.migration.datacenter.core.application;
  */
 public class DatabaseConfiguration
 {
+    public enum DBType {
+        POSTGRESQL,
+        MYSQL,
+        SQLSERVER,
+        ORACLE
+    }
+
     private String host;
     private String name;
     private String username;
     private String password;
     private Integer port;
+    private DBType type;
 
-    public DatabaseConfiguration(String host, Integer port, String name, String username, String password)
+    public DatabaseConfiguration(DBType type, String host, Integer port, String name, String username, String password)
     {
         this.host = host;
         this.name = name;
         this.username = username;
         this.password = password;
         this.port = port;
+        this.type = type;
     }
 
     public String getHost()
@@ -43,5 +52,10 @@ public class DatabaseConfiguration
     public Integer getPort()
     {
         return port;
+    }
+
+    public DBType getType()
+    {
+        return type;
     }
 }
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
@@ -5,6 +5,7 @@ import com.atlassian.migration.datacenter.core.aws.region.RegionService;
 import com.atlassian.migration.datacenter.core.db.DatabaseExtractor;
 import com.atlassian.migration.datacenter.core.db.DatabaseExtractorFactory;
 import com.atlassian.migration.datacenter.core.exceptions.DatabaseMigrationFailure;
+import com.atlassian.migration.datacenter.core.fs.S3Uploader;
 import org.springframework.stereotype.Component;
 import software.amazon.awssdk.auth.credentials.AwsCredentialsProvider;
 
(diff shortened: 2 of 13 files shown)


src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -0,0 +1,139 @@
...
+        this.awsCredentialsProvider = awsCredentialsProvider;
+        this.regionService = regionService;
+        this.tempDirectory = tempDirectory;
+        this.endpointOverride = Optional.of(endointOverride);
+        this.setStatus(MigrationStatus.NOT_STARTED);
+    }
+
+    /**
+     * Start database dump and upload to S3 bucket. This is a blocking operation and should be started from ExecutorService
+     * or preferably from ScheduledJob. The status of the migration can be queried via
```
bpar476: Sad, forgotten, unfinished sentence?

src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -0,0 +1,139 @@
...
+
+        ConcurrentLinkedQueue<Path> uploadQueue = new ConcurrentLinkedQueue<>();
+        FileSystemMigrationProgress progress = new DefaultFilesystemMigrationProgress();
+        FileSystemMigrationErrorReport report = new DefaultFileSystemMigrationErrorReport();
+
+        S3AsyncClientBuilder builder = S3AsyncClient.builder()
+            .credentialsProvider(awsCredentialsProvider)
+            .region(Region.of(regionService.getRegion()));
+        if (endpointOverride.isPresent()) {
+            builder = builder.endpointOverride(endpointOverride.get());
```
bpar476: Is this mainly for the localstack tests?

src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -0,0 +1,139 @@
...
+
+        String bucket = System.getProperty("S3_TARGET_BUCKET_NAME", "trebuchet-testing");
+        S3UploadConfig config = new S3UploadConfig(bucket, client, target.getParent());
+
+        S3Uploader uploader = new S3Uploader(config, report, progress);
+        Crawler crawler = new DirectoryStreamCrawler(report, progress);
+
+        setStatus(MigrationStatus.UPLOAD_IN_PROGRESS);
+        try {
+            crawler.crawlDirectory(target, uploadQueue);
```
bpar476: Is target always going to be a directory? I was under the impression that the non-parallel pg_dump outputs a single file

src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -0,0 +1,139 @@
...
+        } catch (IOException e) {
+
```
bpar476: Should we fail the db migration if we get IOException here?

src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
tarka: Yeah, this is one of the things I'll refactor. However there is a case for including it as a configuration option as a lot of AWS competitors support the S3 API.

src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
tarka: For now it makes sense to use the same setup regardless; it save a bunch of special cases in the code.

src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
tarka: Oops, missed that.

src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
bpar476: Having a look further on I can see we're always using the directory dump method no matter what so this is probably fine 👍

src/main/java/com/atlassian/migration/datacenter/core/aws/db/MigrationStatus.java
```diff
@@ -0,0 +1,48 @@
+package com.atlassian.migration.datacenter.core.aws.db;
+
+import java.util.Optional;
+
+/**
+ * Copyright Atlassian: 11/03/2020
+ */
+public class MigrationStatus
```
theghostwhoforks: Note for later - Should use the state machine states for this instead? or do we see this as something compliementary?

src/main/java/com/atlassian/migration/datacenter/core/application/DatabaseConfiguration.java
```diff
@@ -5,19 +5,28 @@
  */
 public class DatabaseConfiguration
 {
+    public enum DBType {
+        POSTGRESQL,
+        MYSQL,
+        SQLSERVER,
```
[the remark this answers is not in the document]
tarka: SQLSERVER matches the DB URI format, so it's easier to convert: https://confluence.atlassian.com/adminjiraserver/connecting-jira-applications-to-sql-server-2012-938846858.html

PR #59, opened 2020-03-12
Chet 141 parallel tests

theghostwhoforks:
Run tests in parallel using the github permutation matrix.

Notes -

1. Github does not support yaml anchors
2. Profiles are being use to set or unset integration test flags
3. 4 jobs are run in parallel instead of 2 sequential iterations before. While this does not reduce the time by half, it's still faster than before


.github/workflows/mvn_test.yml
```diff
@@ -4,13 +4,13 @@ on: [push]
 
 jobs:
   run_tests:
-    name: Run tests with Java ${{ matrix.java }}
-
+    name: Java${{ matrix.java }} ${{ matrix.profile }} test
```
theghostwhoforks: Test name is rendered on screen as -
`Java8 integration test` and it fits in the tests pane on the left hand side of the actions page

theghostwhoforks:
Ignore the commit list. Just one file has been changed in this PR

theghostwhoforks:
I don't know how to get this merged in - GH says that the older tests that I've changed are required to merge this PR 🤷‍♂

.github/workflows/mvn_test.yml
```diff
@@ -31,6 +31,4 @@ jobs:
         with:
           java-version: ${{ matrix.java }}
       - name: Test with maven
-        run: mvn -B verify
-      - name: Integration tests
-        run: mvn -B clean verify -Pintegration
+        run: mvn -B clean verify -P ${{ matrix.profile}}
```
theghostwhoforks: @bpar476 @nanux - Why do we need clean here? Won't it undo the effect of caching maven deps?

bpar476: I think it was just leftover from the template action. I can't think of any reason you'd want to clean.

PR #60, opened 2020-03-13
CHET-141: Move to use spring config and ensure wireup completes succe…

theghostwhoforks:
…ssfully

Wire all beans using spring config and move away from annotations.

Step 2 will be to inject AWS sdk client.

Once this is done and merged, the state machine will be integrated


src/main/java/com/atlassian/migration/datacenter/core/aws/auth/ProbeAWSAuth.java
```diff
@@ -20,10 +21,10 @@
 public class ProbeAWSAuth {
```
bpar476: We can drop the `@component` here too right?

theghostwhoforks: Yeah - I'll clean this up as part of a different PR, along the FS changes to inject the S3 SDK, and raise a PR

PR #61, opened 2020-03-13
Spike chet 161

andrewpowell:
Adding in the approach to use a spring boot app that is configured by the CFN stack to read files from the SQS queue and write them to a location on the host machine (preferably an EFS mount)


filesystem-processor/pom.xml
```diff
@@ -0,0 +1,126 @@
...
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-starter-aws-messaging</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.springframework.cloud</groupId>
+            <artifactId>spring-cloud-starter-aws-parameter-store-config</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.amazonaws</groupId>
```
bpar476: Any particular reason to use the AWS SDK version 1?

filesystem-processor/src/main/java/com/atlassian/migration/datacenter/fs/processor/services/SQSMessageProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+    private final ObjectMapper objectMapper = new ObjectMapper();
+
+    public SQSMessageProcessor(AmazonS3 s3Client, ThreadPoolTaskExecutor threadPoolTaskExecutor, @Value("${app.jira.file.path}") String jiraHome) {
+        this.s3Client = s3Client;
+        this.threadPoolTaskExecutor = threadPoolTaskExecutor;
+        this.jiraHome = jiraHome;
+    }
+
+
+    @SqsListener(QUEUE_LOGICAL_NAME)
```
bpar476: Oh wow that's neat https://hello.atlassian.net/wiki/spaces/WPT/blog/2020/03/07/658091315/Workplace+Productivity+COVID-19+and+Working+From+Home#Computer-Accessories%2FInternet-Access-Expenses

bpar476: Oh wow that's neat https://cloud.spring.io/spring-cloud-static/spring-cloud-aws/1.2.3.RELEASE/multi/multi__messaging.html#_annotation_driven_listener_endpoints

filesystem-processor/src/main/java/com/atlassian/migration/datacenter/fs/processor/services/SQSMessageProcessor.java
```diff
@@ -0,0 +1,57 @@
...
+        log.debug("Received SQS message {}", s3EventNotificationRecord.toJson());
+        List<S3EventNotificationRecord> s3EventNotificationRecords = s3EventNotificationRecord.getRecords();
+        log.debug("Received ".concat(Integer.toString(s3EventNotificationRecords.size())).concat(" records from S3."));
+        File jiraHomePath = new File(this.jiraHome);
+        if (!jiraHomePath.exists()) {
+            if (jiraHomePath.mkdir()) {
+                log.debug("Created Jira Home path ".concat(jiraHomePath.getAbsolutePath()));
+            }
+        }
+        if (s3EventNotificationRecords.size() == 1) {
```
bpar476: What's the reason for not just using `forEach` in all cases?

migration-helper.yml
```diff
@@ -0,0 +1,390 @@
...
+        Statement:
+          - Effect: Allow
+            Principal:
+              AWS: '*'
+            Action:
+              - SQS:SendMessage
+            Resource: '*'
+            Condition:
+              ArnEquals:
+                aws:SourceArn: "arn:aws:s3:::*"
```
bpar476: Does this mean only S3 resources can put events into the queue?

src/test/java/com/atlassian/migration/datacenter/core/aws/infrastructure/LocalStackEnvironmentVariables.java
```diff
@@ -0,0 +1,12 @@
+package com.atlassian.migration.datacenter.core.aws.infrastructure;
+
+import cloud.localstack.docker.annotation.IEnvironmentVariableProvider;
+
+import java.util.Map;
+
+public class LocalStackEnvironmentVariables implements IEnvironmentVariableProvider {
+    @Override
+    public Map<String, String> getEnvironmentVariables() {
```
bpar476: What's the use case for providing these environment variables?

migration-helper.yml
```diff
@@ -0,0 +1,390 @@
+AWSTemplateFormatVersion: 2010-09-09
```
bpar476: Is this a duplication of templates/migration-stack/migration-helper.yml?

templates/migration-stack/migration-helper.yml
```diff
@@ -1,48 +1,121 @@
...
+    Properties:
+      Name: /config/migration-helper/app.s3.vpc.endpoints
+      Type: StringList
+      Value: !GetAtt EndpointS3.DnsEntries
+      Description: CloudFormation Stack Name
+
+  StackNameParameter:
+    Type: AWS::SSM::Parameter
+    Properties:
+      Name: /config/migration-helper/cloud.aws.stack.name
```
bpar476: I don't see any reference to this parameter in the spring boot app. Has it got an intended use?

templates/migration-stack/migration-helper.yml
```diff
@@ -1,48 +1,121 @@
...
+      Type: String
+      Value: !Ref AWS::StackName
+      Description: CloudFormation Stack Name
+
+  JiraFilePathParameter:
```
bpar476: Is the intention that this value can be set by a stack parameter? I think the path will always be `/efs/jira/shared` in the migration stack EC2 instance

templates/migration-stack/migration-helper.yml
```diff
@@ -1,48 +1,121 @@
...
+    Type: AWS::SQS::Queue
+    Properties:
+      DelaySeconds: 0
+      MaximumMessageSize: 262144
+      MessageRetentionPeriod: 864000
+      QueueName: !Sub 'atl-migration-queue-${AWS::StackName}'
+      ReceiveMessageWaitTimeSeconds: 0
+      VisibilityTimeout: 90
+
+  EndpointS3:
```
bpar476: Is the VPC endpoint for S3 a performance thing?

filesystem-processor/src/main/java/com/atlassian/migration/datacenter/fs/processor/services/SQSMessageProcessor.java
theghostwhoforks: wrong link @bpar476 ?

theghostwhoforks: will delete this  comment

bpar476:
Oh and one more question I just thought of:

How do we signal to the message consumer that there are no more files coming?

It's not enough for the queue to be empty as there could be files not uploaded from the source system yet. If we're purely populating the queue via S3 events then we would need to send a "canary" file as the signal.
If we allow other items to be sent on the queue we could send a specific message on the queue that indicates there are no more files being uploaded.

filesystem-processor/pom.xml
andrewpowell: This is what is used by the Spring Cloud AWS project.

filesystem-processor/src/main/java/com/atlassian/migration/datacenter/fs/processor/services/SQSMessageProcessor.java
andrewpowell: 👍

migration-helper.yml
andrewpowell: correct

src/test/java/com/atlassian/migration/datacenter/core/aws/infrastructure/LocalStackEnvironmentVariables.java
andrewpowell: I have a localstack pro license, it gets picked up in the environment variables on the host.

migration-helper.yml
andrewpowell: Yes

templates/migration-stack/migration-helper.yml
andrewpowell: Yes, that's the intention. We can use that path as a default.

templates/migration-stack/migration-helper.yml
andrewpowell: It's a security thing.  Make sure that data only goes over the VPC for S3 calls.  This is not needed now b/c it's included in the ASI

andrewpowell:
So this would need to be installed on the Helper Instance.  It's built as an executable JAR.  I think we could send metadata on the S3 transfer of the "canary file", or last file in the process.

migration-helper.yml
bpar476: Do we need to keep it?

andrewpowell:
Closing Draft

PR #62, opened 2020-03-15
no-issue: cache npm dependencies on CI

theghostwhoforks:
cache npm on CI for faster builds

### DIFF
--- a/.github/workflows/mvn_test.yml
+++ b/.github/workflows/mvn_test.yml
@@ -32,7 +32,7 @@ jobs:
           cache-name: cache-npm
         with:
           path: ~/.npm
-          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('package-lock.json') }}
           restore-keys: |
             ${{ runner.os }}-build-${{ env.cache-name }}-
             ${{ runner.os }}-build-
--- a/.github/workflows/mvn_test.yml
+++ b/.github/workflows/mvn_test.yml
@@ -21,7 +21,18 @@ jobs:
           cache-name: cache-m2
         with:
           path: ~/.m2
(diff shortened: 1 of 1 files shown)


.github/workflows/mvn_test.yml
```diff
@@ -26,6 +26,17 @@ jobs:
             ${{ runner.os }}-build-${{ env.cache-name }}-
             ${{ runner.os }}-build-
             ${{ runner.os }}-
+      - name: Cache npm dependencies
+        uses: actions/cache@v1
+        env:
+          cache-name: cache-npm
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/yarn.lock') }}
```
bpar476: You shouldn't need to cache the lock file as it's in VCS so will always be up to date. What needs to be cached is the `node_modules` directory.

bpar476: Whoops, I'm silly the yarn lock is just the hash file which is correct 👍 However, shouldn't the path be `node_modules`?

theghostwhoforks: https://help.github.com/en/actions/configuring-and-managing-workflows/caching-dependencies-to-speed-up-workflows#matching-a-cache-key

Yeah, it's just the hashfile!

I thought it should be `node_modules` too, but the doc asks to cache the `~/.npm` folder. Let me change this to the node_modules folder and see if it makes any difference in timing

theghostwhoforks: Screenshot 2020-03-16 at 12 45 27

Screenshot 2020-03-16 at 12 44 57

The cache size is the same (!) even while using the .npm or the node_modules folder

bpar476: That's very strange. I wonder if they have a weird config for storing node dependencies in the runner

bpar476: Looks like `~/.npm` is used to populate `node_modules` if it exists

theghostwhoforks: yeah. It stores the indexes and other cache data. Interesting.

PR #63, opened 2020-03-15
CHET-158: Script for status of s3 sync down

bpar476:
# Summary

* Python script which reads the last line of the s3 sync log.
* Checks if completed
* Checks for progress
* Checks if an error has occurred
* Prints a JSON object to stdout with a summary of the status
* Use Makefile to render the python script into the template

## Looking for feedback

Is there a better way to render the template with the python script? I wanted it to be external so that we can run it locally for testing and development. However, I don't like that you need to render the template and commit the source template and the rendered template to VCS. I think for the final release we should maybe upload the template to S3. That means we can have a GitHub action to render the template and upload it to S3 on commit to master. No more source controlling the generated file.


templates/migration-stack/Makefile
```diff
@@ -3,12 +3,18 @@ INSTANCE_TYPE ?= c5.large
...
+# SUBNET_ID: the IDs of the subnets to distribute the elastic beanstalk EC2 instance between (There is only 1 instance so should just be a single subnet)
 # EFS_ID: the file system ID of the EFS to sync the files to. Should be the EFS of the freshly deployed product stack
 # SG_ID: The ID of the security group attached to the EFS, required to allow the beanstalk instance to communicate with the EFS for the duration of the migration
 
-PARAMS := ParameterKey=HelperVpcId,ParameterValue=$(VPC_ID) ParameterKey=NetworkPrivateSubnet,ParameterValue=$(SUBNET_IDS) ParameterKey=EFSFileSystemId,ParameterValue=$(EFS_ID) ParameterKey=EFSSecurityGroup,ParameterValue=$(SG_ID) ParameterKey=HelperInstanceType,ParameterValue=$(INSTANCE_TYPE)
+PARAMS := ParameterKey=HelperVpcId,ParameterValue=$(VPC_ID) ParameterKey=NetworkPrivateSubnet,ParameterValue=$(SUBNET_ID) ParameterKey=EFSFileSystemId,ParameterValue=$(EFS_ID) ParameterKey=EFSSecurityGroup,ParameterValue=$(SG_ID) ParameterKey=HelperInstanceType,ParameterValue=$(INSTANCE_TYPE)
 
+rendertemplate:
+	mkdir render;
+	echo "/%{HOME_COPY_SCRIPT}%/{\n\ts/%{HOME_COPY_SCRIPT}%//g\n\tr render/tab_sync.py\n}" > render/sed
```
theghostwhoforks: https://www.gnu.org/software/make/manual/make.html#Recipe-Echoing
You may want to add an "@" prefix to make commands to suppress the default command echo

templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -0,0 +1,371 @@
...
+      Policies:
+        - PolicyName: MigrationBucketFullAccess
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Action: 's3:*'
+                Effect: Allow
+                Resource: !Sub ['arn:aws:s3:::${MigrationBucket}/*', { MigrationBucket: !Ref MigrationBucket
+           }]
+              - Action:
```
theghostwhoforks: Curious to understand - why the separate perms for list bucket and head bucket? To handle the case of the root object?

theghostwhoforks:
With respect to the feedback section, what we have is good for now. To scale it further, we could go down the ansible route and store the python code in the `dc-deployments-automation` repo and have a role for the migration stack that installs the scripts in a local bin that is executable by the jira/connie/bb user

The code can reside in a lib folder in the automation repo

templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -0,0 +1,371 @@
...
+                Action: "*"
+                Resource: "*"
+  OpenEFSAccessCustom:
+    Type: Custom::OpenEFSAccessCustom
+    Version: 1.0
+    Properties:
+      ServiceToken: !GetAtt OpenEFSAccess.Arn
+      HelperSG: !Ref HelperSecurityGroup
+      EFSSG: !Ref EFSSecurityGroup
+  OpenEFSAccess:
```
theghostwhoforks: How is this lambda func invoked?

templates/migration-stack/pkg/migration-helper.yml.src
bpar476: It's because the resource is a bucket for these operations so there needs to be a different pattern for the `Resource` section of the policy document. I may be wrong about this though 🤔

templates/migration-stack/pkg/migration-helper.yml.src
bpar476: It's invoked by `OpenEFSAccessCustom`
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-custom-resources.html

Stephen hooked this up

templates/migration-stack/Makefile
bpar476: Going off that doc
>  Typically you would use this for a command whose only effect is to print something
I'm going to leave it unsuppressed

PR #64, opened 2020-03-15
Re-Resolve yarn dependencies

bpar476:
This resolves the security vulnerability in `acorn@7.1.0`. I simply deleted the node_modules directory and the lockfile and regenerated it completely so we should have a more efficient resolution now


bpar476:
The snapshots had to be regenerated because the classes generated for the components changed after the dependency upgrade. Probably due to a different version of styled-components

PR #65, opened 2020-03-16
CHET-173: Refactor S3FilesystemMigrationService to be more testable, and expand integration tests

tarka:
This is separate from the service refactoring; this one ensures that we have a working baseline for the S3FilesystemMigrationService tests.

### DIFF
--- a/src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
+++ b/src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
@@ -1,7 +1,5 @@
 package com.atlassian.migration.datacenter.core.fs;
 
-import cloud.localstack.docker.LocalstackDockerExtension;
-import cloud.localstack.docker.annotation.LocalstackDockerProperties;
 import com.atlassian.jira.config.util.JiraHome;
 import com.atlassian.migration.datacenter.core.aws.auth.AtlassianPluginAWSCredentialsProvider;
 import com.atlassian.migration.datacenter.core.aws.region.RegionService;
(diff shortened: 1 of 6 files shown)


src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
```diff
@@ -19,54 +18,90 @@
...
-    private SchedulerService schedulerService;
+    AtlassianPluginAWSCredentialsProvider credentialsProvider;
 
     @BeforeEach
     void setup() throws Exception {
+        credentialsProvider = new AtlassianPluginAWSCredentialsProvider(new AwsCredentialsProviderShim(s3.getDefaultCredentialsProvider()));
+
+        s3AsyncClient = S3AsyncClient.builder()
+            .endpointOverride(new URI(s3.getEndpointConfiguration(S3).getServiceEndpoint()))
+            .credentialsProvider(new AwsCredentialsProviderShim(s3.getDefaultCredentialsProvider()))
```
theghostwhoforks: why do we need the shim? As opposed to using the defaultCredentialsProvider from S3?

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
```diff
@@ -19,54 +18,90 @@
...
         fsService.startMigration();
 
         Assertions.assertNotEquals(FilesystemMigrationStatus.FAILED, fsService.getReport().getStatus());
+
+        HeadObjectRequest req = HeadObjectRequest.builder()
+            .bucket(bucket)
+            .key(file.getFileName().toString())
+            .build();
+        HeadObjectResponse resp = s3AsyncClient.headObject(req).get();
+        assertTrue(resp.sdkHttpResponse().isSuccessful());
```
theghostwhoforks: Additionally, do we need to assert for content or size?

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
tarka: TestContainers/LocalStack provides v1 credentials, but we use the v2 API. It's just a quick wrapper around converting them.

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceIT.java
tarka: Ideally, but I'm being slack ;) Thats a bit more work and I need a baseline. I'll add some more test utils in the refactor.

PR #67, opened 2020-03-16
CHET-157: kick off efs hydration

bpar476:
# Summary

* Use SsmAPI to invoke the s3 sync command
* Ensure the command fires successfully
* Integrate with S3FileSystemMigrationService


src/test/java/com/atlassian/migration/datacenter/core/fs/S3SyncFileSystemDownloaderIT.java
```diff
@@ -0,0 +1,69 @@
...
+
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.fail;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.anyMap;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+@Tag("integration")
```
theghostwhoforks: Why is this an integration test. The SSM mocks are injected, stubbed and verified - looks like a unit test to me?

Given localstack limitations, we can't have an integration test for SSM components and that's fine for now, until the functest plugin is introduced to do a higher level of testing.

bpar476: The test takes 10 seconds due to the retry mechanism :( I could make the retry variable package-private and reset it from the test?

theghostwhoforks: The `MAX_COMMAND_STATUS_CHECK_RETRIES` variable?

Maybe making it a prop and setting it from the test may be a good thing to do

PR #70, opened 2020-03-17
Chet 174 fix osgi imports java config

theghostwhoforks:
This PR fixes the issue with the plugin not being able to consume OSGI services at start up.

From a conversation with auditing, this is an issue with atlassian-spring-java-config library. The imported OSGI services are not in a state to be used. 

To resolve this we -

1. Make the components that use the OSGI reference use a Lazily constructed object
2. Consumption of the injected service does not happen in the constructor. Rather it happens in a `@PostConstruct` method
3. PostConstruct annotations are moved to `javax.annotation-api` package, which has now been included


theghostwhoforks:
I recommend that the reviewers checkout this branch and start the plugin 

`mvn package amps:run -Dproduct=jira -DskipTests -Dmvn.skip.test`

and verify if the post construct annotation is working as expected

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
```diff
@@ -89,4 +100,28 @@ public SSMApi ssmApi(SsmClient client) {
...
+        return new AWSConfigurationService((WriteCredentialsService) awsCredentialsProvider, regionService, migrationService);
+    }
+
+    @Bean
+    public CfnApi cfnApi(AwsCredentialsProvider awsCredentialsProvider, RegionService regionService) {
+        return new CfnApi(awsCredentialsProvider, regionService);
+    }
+
+    @Bean
+    public MigrationService migrationService(ActiveObjects ao) {
```
nanux: Do we need to convert other components to beans as well? E.g. `S3FilesystemMigrationService`?

theghostwhoforks: When we inject the service into an endpoint, yes!

There weren't any usecases, so I didn't want to create beans for as yet un-used references

src/main/java/com/atlassian/migration/datacenter/core/aws/region/PluginSettingsRegionManager.java
```diff
@@ -3,35 +3,42 @@
...
+    private PluginSettings pluginSettings;
 
-    public PluginSettingsRegionManager(PluginSettingsFactory pluginSettingsFactory, GlobalInfrastructure globalInfrastructure) {
-        this.pluginSettingsFactory = pluginSettingsFactory;
+    public PluginSettingsRegionManager(Supplier<PluginSettingsFactory> pluginSettingsFactorySupplier, GlobalInfrastructure globalInfrastructure) {
+        this.pluginSettingsFactorySupplier = pluginSettingsFactorySupplier;
         this.globalInfrastructure = globalInfrastructure;
     }
 
+    @PostConstruct
```
nanux: This `@PostConstruct` method will be required for any class that works with `pluginSettingsFactory` instance?

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
```diff
@@ -25,7 +36,7 @@
 import java.nio.file.Paths;
 
 @Configuration
-@Import({MigrationAssistantOsgiImportConfiguration.class})
+@ComponentScan
```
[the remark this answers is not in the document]
theghostwhoforks: aha! component scan, without an explicit package name scans the immediate parent package.

Component scan picks up configuration classes - which is needed here (atleast for idea) for the configuration classes to be recognized. 

If you remove it, idea will go red as it's not able to find the bean definitions. However, spring will be fine. As it only scans classes under the `configuration` package, the number of classes scanned is small, which is why I've left this here

src/main/java/com/atlassian/migration/datacenter/configuration/SpringOsgiConfigurationUtil.java
```diff
@@ -0,0 +1,23 @@
...
+import static com.atlassian.plugins.osgi.javaconfig.OsgiServices.importOsgiService;
+
+/**
+ * Common methods used in Spring OSGI Configuration of the plugin
+ */
+public class SpringOsgiConfigurationUtil {
+    /**
+     * Works around the Spring Configuration problem when we try to import the class which is not available yet.
+     */
+    public static <T> Supplier<T> lazyImportOsgiService(Class<T> clazz){
```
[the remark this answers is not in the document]
theghostwhoforks: Yes, you are right. However, there appears to be a bug in the `java-osgi-config` that we encountered yesterday that brings in plugin components (in this case from `sal` that has not been fully constructed).

I'll reach out to the server platform and add a ticket reference here. Once that is resolved, this should not be necessary

src/main/java/com/atlassian/migration/datacenter/core/aws/auth/EncryptedCredentialsStorage.java
```diff
@@ -36,66 +32,37 @@
...
+    @PostConstruct
+    public void postConstruct() {
         assert this.jiraHome != null;
         String keyFilePath = this.jiraHome.getHome().getPath().concat("/").concat(ENCRYPTION_KEY_FILE_NAME);
         String saltFilePath = this.jiraHome.getHome().getPath().concat("/").concat(ENCRYPTION_SALT_FILE_NAME);
         String password = getEncryptionData(keyFilePath);
         String salt = getEncryptionData(saltFilePath);
         this.textEncryptor = Encryptors.text(password, salt);
-        this.pluginSettings = pluginSettingsFactory.createGlobalSettings();
+        this.pluginSettings = this.pluginSettingsFactorySupplier.get().createGlobalSettings();
```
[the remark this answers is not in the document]
theghostwhoforks: `pluginSettingsFactorySupplier.get()` returns an object of type `T`

`createGlobalSettings` will always return a `PluginSettings` object. (or atleast as per the javadoc)

src/main/java/com/atlassian/migration/datacenter/core/aws/region/PluginSettingsRegionManager.java
theghostwhoforks: Yes. Though, this was happening/or should happen in the `afterPropertiesSet` method of any `InitializingBean`.

From that perspective, it is equivalent. we shouldn't be assigning properties in the constructor as it _may_ error out.

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
```diff
@@ -89,4 +100,28 @@ public SSMApi ssmApi(SsmClient client) {
...
         return new S3SyncFileSystemDownloader(ssmApi);
     }
+
+    @Bean
+    public AvailabilityZoneManager availabilityZoneManager(AwsCredentialsProvider awsCredentialsProvider, GlobalInfrastructure globalInfrastructure) {
+        return new AvailabilityZoneManager(awsCredentialsProvider, globalInfrastructure);
+    }
+
+    public AWSConfigurationService awsConfigurationService(AwsCredentialsProvider awsCredentialsProvider, RegionService regionService, MigrationService migrationService) {
+        return new AWSConfigurationService((WriteCredentialsService) awsCredentialsProvider, regionService, migrationService);
```
bpar476: Is it common practice to cast like this? I get that the only credentials provider Bean is also a credentials service but why not make it explicit?

src/main/resources/META-INF/spring/plugin-context.xml
```diff
@@ -8,5 +8,5 @@
        http://www.springframework.org/schema/context/spring-context.xsd">
 
     <context:annotation-config/>
-    <context:component-scan base-package="com.atlassian.migration.datacenter" />
+    <context:component-scan base-package="com.atlassian.migration.datacenter.configuration" />
```
bpar476: Why do we need this in addition to the `@ComponentScan` annotation?

theghostwhoforks: I'll change this to just the root bean definition class reference

src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
theghostwhoforks: Only because we don't have a definition for `writeCredentialService`. If we do, then we need to make the `awscredentialsProvider` a factory bean object, so that the same singleton reference is returned.

The cast here is OK, IMO, as `AWSCredentailsProvider` implements both the `readCredentialsService` and `writeCredentialsService` interface

src/main/java/com/atlassian/migration/datacenter/core/aws/auth/EncryptedCredentialsStorage.java
```diff
@@ -52,66 +48,48 @@
...
-        } catch (IOException ex) {
-            logger.error(ex.getLocalizedMessage());
-        }
-        file.setWritable(false, true);
-        return keyString;
-    }
-
-    public void afterPropertiesSet() {
+    @PostConstruct
+    // FIXME: I do not work
```
nanux: What will happen in the application?

nanux: I see, we need to call the supplier every time when we need pluginSettings.

theghostwhoforks:
Yes.. also, this pattern is similar to what the auditing plugin uses right
now

PR #71, opened 2020-03-17
CHET-169: Add state machine hooks into database migration service

theghostwhoforks:
wanted to get @tarka's initial thoughts on database migration status and the need, if any, to co-exist with the state machine


theghostwhoforks:
I can't test if the beans are being injected without the java osgi refactor branch being merged in.

Once that is updated, I'll test and post the results here

src/test/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseArchivalServiceTest.java
```diff
@@ -0,0 +1,81 @@
...
+        service = new DatabaseArchivalService(migrationService, databaseExtractor);
+    }
+
+    @Test
+    void shouldArchiveDatabaseSuccessfully() throws Exception {
+        when(this.databaseExtractor.startDatabaseDump(tempDir.resolve("db.dump"))).thenReturn(process);
+        when(process.waitFor()).thenReturn(0);
+        Path target = service.archiveDatabase(tempDir);
+        assertTrue(target.endsWith("db.dump"));
+        //TODO: Should this stage to transition to be called pending export?
```
bpar476: I think the current nomenclature is fine.

src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -1,89 +1,70 @@
...
-            throw new DatabaseMigrationFailure(msg, e);
-        }
-        setStatus(MigrationStatus.DUMP_COMPLETE);
-
+    public FileSystemMigrationErrorReport performMigration() throws DatabaseMigrationFailure, InvalidMigrationStageError {
+        Path pathToDatabaseFile = databaseArchivalService.archiveDatabase(tempDirectory);
+        return uploadDatabaseArtifactToS3(pathToDatabaseFile);
+    }
 
+    //TODO: This should be decomposed as a standalone service for easier testing
```
bpar476: I think I agree with this. If the database dumping is being separated from the service then the storage probably should as well.

src/main/java/com/atlassian/migration/datacenter/spi/MigrationStage.java
```diff
@@ -8,13 +8,24 @@
...
+
+    WAIT_PROVISION_MIGRATION_STACK("wait_provision_migration"),
+    PROVISION_MIGRATION_STACK("provision_migration"),
+    WAIT_PROVISION_APPLICATION("wait_provision_app"),
+    PROVISION_APPLICATION("provision_app"),
     AUTHENTICATION("authentication"),
-    NOT_STARTED("");
+    NOT_STARTED("not_started");
 
     private String key;
```
tarka: What's the purpose of this?

src/main/java/com/atlassian/migration/datacenter/spi/MigrationStage.java
```diff
@@ -8,13 +8,24 @@
     FINISHED("finished"),
     CUTOVER("cutover"),
     VALIDATE("validate"),
-    DB_MIGRATION_IMPORT("db_migration_down"), DB_MIGRATION_EXPORT("db_migration_up"),
+
+    DB_MIGRATION_IMPORT("db_migration_down"),
```
tarka: I'm replace the DB migration state enum with this one. However I if we have an error we should include more information; I think a wrapper class that can encapsulate this information should work.

src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
theghostwhoforks: Cool. Will wrap this in a small service

src/main/java/com/atlassian/migration/datacenter/spi/MigrationStage.java
theghostwhoforks: This is here as part of the `DevelopEndpoint`. This was added to aid in deserialization of a `MigrationStage` object

cc: @nanux

src/main/java/com/atlassian/migration/datacenter/spi/MigrationStage.java
theghostwhoforks: I've updated all the enums to represent the state transitions.

/nit. I'd be a tad happier if we can make the currentStage return a callback for `next()` transition that accepts a migration service 

Something like this -
`getCurrentStage().next().apply(migrationService)`
can be accomplished easily.

I'll leave this for another refactoring.

theghostwhoforks:
Note to self:
This is pending -
1. Fix Develop endpoint 
2. Extract a utility function to a common builder

Once done, and tested, this PR will be made ready to merge

src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
```diff
@@ -60,7 +60,7 @@ public QuickstartDeploymentService(ActiveObjects ao, CfnApi cfnApi, MigrationSer
     @Override
     public void deployApplication(String deploymentId, Map<String, String> params) throws InvalidMigrationStageError {
         logger.info("received request to deploy application");
-        migrationService.transition(MigrationStage.PROVISION_APPLICATION, MigrationStage.WAIT_PROVISION_APPLICATION);
+        migrationService.transition(MigrationStage.PROVISION_APPLICATION_WAIT);
```
tarka: As we discussed in standup, we should probably pick one system. But we don't need to do it straight away if we want to see what approach works best.

theghostwhoforks:
All merge conflicts have now been resolved. They were all logical conflicts, something that couldn't be automatically resolved. 

Once the tests pass, this shall be merged in ASAP

PR #73, opened 2020-03-17
CHET-142: Dev endpoint to set stage

nanux:
The JSON payload is for example `"FS_MIGRATION_COPY"`.

Jackson is very picky with enums. Other option is to use `@PathParam` but I will push this forward and it is also availble from Postman.

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/api/develop/DevelopEndpoint.java
+++ b/src/main/java/com/atlassian/migration/datacenter/api/develop/DevelopEndpoint.java
@@ -1,0 +1,60 @@
+package com.atlassian.migration.datacenter.api.develop;
+
+import com.atlassian.migration.datacenter.core.exceptions.InvalidMigrationStageError;
+import com.atlassian.migration.datacenter.spi.MigrationService;
+import com.atlassian.migration.datacenter.spi.MigrationStage;
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.google.common.collect.ImmutableMap;
+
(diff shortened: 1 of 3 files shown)


src/main/java/com/atlassian/migration/datacenter/api/develop/DevelopEndpoint.java
```diff
@@ -0,0 +1,62 @@
...
+    public Response setMigrationStage(Stage stage) {
+        try {
+            final MigrationStage targetStage = MigrationStage.valueOf(stage.getStage());
+            final MigrationStage currentStage = migrationService.getCurrentStage();
+            migrationService.transition(currentStage, targetStage);
+            return Response
+                    .ok(ImmutableMap.of("stage", migrationService.getCurrentStage()))
+                    .build();
+        } catch (InvalidMigrationStageError | IllegalArgumentException e) {
+            return Response
```
bpar476: Should we surface the exception in logs here?

src/main/java/com/atlassian/migration/datacenter/api/develop/DevelopEndpoint.java
```diff
@@ -0,0 +1,62 @@
...
+                    .status(Response.Status.CONFLICT)
+                    .entity(ImmutableMap.of("error", String.format("Unable to transition migration to %s", stage.getStage())))
+                    .build();
+        }
+    }
+
+    @JsonAutoDetect
+    static class Stage {
```
bpar476: Why not just use the `MigrationStage` enum as the payload?

PR #75, opened 2020-03-19
Chet 169 db config parse

theghostwhoforks:
Fixes bean load on startup with a h2 database


theghostwhoforks:
The same code is present in `CHET-169-db-state-machine`. I pulled it here to allow for master to be pristine

src/main/java/com/atlassian/migration/datacenter/core/db/DatabaseExtractorFactory.java
```diff
@@ -19,16 +19,20 @@
...
     public static DatabaseExtractor getExtractor(ApplicationConfiguration config) throws DatabaseMigrationFailure
     {
         if (config.getDatabaseConfiguration().getType().equals(DBType.POSTGRESQL)) {
             return new PostgresExtractor(config);
-        } else {
-            throw new DatabaseMigrationFailure("Unsupported database type: " + config.getDatabaseConfiguration().getType());
         }
+
+        //Profile scoped perhaps?
+        if (config.getDatabaseConfiguration().getType().equals(DBType.H2)) {
```
nanux: With H2, I don't think we need to fail directly. We can just skip the dump and copy the database when we are running the filesystem transfer (the second one). I am not really fussed about it as H2 is not really a target platform so complicating the code is not necessary.

bpar476: I think I agree as testing locally with H2 makes it quite fast. However, we may want to introduce a run profile which starts jira with postgres in docker

theghostwhoforks: Yeah - we need to start in postgres by default going forward. Once we create a run profile, this is not necessary

theghostwhoforks: For now, this is necessary to start the plugin. Without this, the app start will fail as the database configuration throws an exception

PR #78, opened 2020-03-20
CHET-151: Abort filesystem migration REST endpoint

nanux:
* adds DELETE `/migration/fs/abort` endpoint that is is useful when testing FS and also eventually in the frontend
* consistent `InvalidMigrationStageError` from `scheduleMigration`

Had a bit of headache with git after the osgi/copyright/ssm changes, so I've redone the branch and also moved out cleanup bits out of the PR, I will create followup.


src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceTest.java
```diff
@@ -131,11 +132,41 @@ void shouldUnsetScheduledJobWhenSchedulerExceptionIsRaised() throws Exception {
         verify(migrationService).error();
     }
 
+    @Test
+    void shouldAbortRunningMigration() throws Exception {
+        mockJobDetailsAndMigration(MigrationStage.WAIT_FS_MIGRATION_COPY);
+
+        final FilesystemUploader uploader = mock(FilesystemUploader.class);
+        FieldSetter.setField(fsService, fsService.getClass().getDeclaredField("fsUploader"), uploader);
```
bpar476: Why use reflection instead of just creating the object?

nanux: `fsUploader` is lazily initialised in `scheduleMigration()` method which I don't want to call in this case.

theghostwhoforks: This probably requires a refactor in the future then to create an uploader upfront. The `FsUploader` object can be injected either by removing the constructor args or injecting a function instead.

It caused a bit of difficulty when testing the database service too. It's something we need to refactor anyway. Let's merge this in and tackle this going forward

src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -178,10 +185,33 @@ public void startMigration() throws InvalidMigrationStageError {
...
+    @Override
+    public void abortMigration() throws InvalidMigrationStageError {
+        // we always try to remove scheduled job if the system is in inconsistent state
+        if (schedulerService.getJobDetails(getScheduledJobId()) != null) {
+            schedulerService.unscheduleJob(getScheduledJobId());
+            logger.info("Removed scheduled filesystem migration job");
+        }
+
+        if (!isRunning() || fsUploader == null) {
+            throw new InvalidMigrationStageError(String.format("Invalid migration stage when cancelling filesystem migration: %s", migrationService.getCurrentStage()));
```
theghostwhoforks: Is this the correct error/message here?

nanux: Yes, the migration needs to be running to be able to cancel it and if either of the conditions is not fulfilled then the service is in an incorrect state. The first few lines of the method should ensure we are able to unschedule the job and run it again if we ended up in some inconsistent state - that is possible mostly in development but right now the state machine is quite raw so I rather left it in.

src/test/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationServiceTest.java
nanux: I've created https://aws-partner.atlassian.net/browse/CHET-199 to follow up.

PR #81, opened 2020-03-20
CHET-188: install psql client on migration stack

theghostwhoforks:
This installs postgres 9.6 client from amazon-linux-extras
Screenshot 2020-03-20 at 15 20 02


templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -105,6 +105,10 @@ Resources:
             commands:
               01_mount:
                 command: /opt/atlassian/dc-migration-assistant/mount-efs.sh
+          install_psql_client:
+            commands:
+              01_install_psql:
+                command: amazon-linux-extras install -y postgresql9.6
```
nanux: Should we just install `postgresql11`? I've just tested `amazon-linux-extras install -y postgresql11` and it works fine.
`psql` and `pg_dump` should be backwards compatible with older versions so it should give us more breathing room and higher flexibility for customer deployments.

theghostwhoforks: good idea. As long as they are compatible, we can add that in.

I can make the version configurable and default to 11 🤔

templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -105,6 +105,10 @@ Resources:
             commands:
               01_mount:
                 command: /opt/atlassian/dc-migration-assistant/mount-efs.sh
+          install_psql_client:
```
bpar476: Is there any reason we can't use the cfn-init `packages` property to install psql?

bpar476: Looks like `amazon-linux-extras` isn't a supported `cfn-init` package

theghostwhoforks: oh yeah! Sorry, I missed this comment last week

templates/migration-stack/pkg/migration-helper.yml.src
theghostwhoforks: Done

Screenshot 2020-03-30 at 12 21 58

theghostwhoforks: The default version is 11

Screenshot 2020-03-30 at 12 23 06

bpar476:
Isn't it still installing postgres 9.6? 🤔

theghostwhoforks:
> Isn't it still installing postgres 9.6? 🤔

Have you seen the screenshot above? With the default value of 11, I've verified that the psql client version is 11.x

bpar476:
> > Isn't it still installing postgres 9.6? 🤔
> 
> Have you seen the screenshot above? With the default value of 11, I've verified that the psql client version is 11.x

Even with `command: amazon-linux-extras install -y postgresql9.6`

theghostwhoforks:
> > > Isn't it still installing postgres 9.6? 🤔
> > 
> > 
> > Have you seen the screenshot above? With the default value of 11, I've verified that the psql client version is 11.x
> 
> Even with `command: amazon-linux-extras install -y postgresql9.6`?

For future reference, commits containing the latest changes weren't pushed to remote. This has been done now

PR #83, opened 2020-03-24
Iw kotlin

theghostwhoforks:
Draft PR for easy diffs


src/main/kotlin/com/atlassian/migration/datacenter/api/aws/AWSGlobalInfrastructureEndpoint.kt
```diff
@@ -13,45 +13,33 @@
...
-    public Response getRegions() {
-        List<String> regions = globalInfrastructure.getRegions();
-        if (regions == null) {
-            return Response
-                    .serverError()
-                    .build();
-        }
-
+    fun getRegions(): Response {
+        val regions = globalInfrastructure.regions
```
theghostwhoforks: funky syntax

src/main/kotlin/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.kt
```diff
@@ -0,0 +1,97 @@
...
+@Path("/migration/fs")
+class FileSystemMigrationEndpoint(private val fsMigrationService: FilesystemMigrationService) {
+    private val mapper: ObjectMapper = ObjectMapper()
+
+    @PUT
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/start")
+    fun runFileMigration(): Response {
+        return if (fsMigrationService.isRunning) {
```
theghostwhoforks: I like the `return` prefix - makes things clearer than what it is in Java with a return inside the `if` statement

PR #84, opened 2020-03-29
CHET-191: Orchestrate DB restore

bpar476:
# Summary

Creates SsmPsqlDbRestoreService which invokes an SSM document which will use PSQL to restore the database. This document hasn't been created yet but it is trivial to implement as it can be run synchronously.


tarka:
You might be missing a mock in the tests.

PR #85, opened 2020-03-30
Fix duplicate transition between db export and import

bpar476:
The `transitionToNext` of database archive was the same as `transitionToStart` of the database restore which would have caused an exception. I've refactored the callbacks so that they simply assert they are in the expected stage before starting instead of transitioning to the start stage. The transition to the start stage of any operation will be done by the `transitionToNext` of the previous operation.


theghostwhoforks:
testing with psql11

PR #86, opened 2020-03-30
CHET-202: harvest encryption

bpar476:
# Summary

Harvests the encryption component of the encryption manager into its own singleton so that it can be used to encrypt the new database credentials.

These credentials will be needed by the migration stack.

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
+++ b/src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
@@ -46,6 +46,7 @@ import com.atlassian.migration.datacenter.core.db.DatabaseExtractorFactory;
 import com.atlassian.migration.datacenter.core.fs.S3FilesystemMigrationService;
 import com.atlassian.migration.datacenter.core.fs.download.s3sync.S3SyncFileSystemDownloadManager;
 import com.atlassian.migration.datacenter.core.fs.download.s3sync.S3SyncFileSystemDownloader;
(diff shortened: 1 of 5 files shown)


theghostwhoforks:
How do we see the flow? 

1. The plugin takes a backup of the database
2. The plugin then encrypts the credentials (username & password)
3. The SSM play on the migration stack uses the backup file and credentials to perform the restore.

Open Questions - 

1. How does the decryption happen?
2. How is the database file identified in the S3 bucket? i.e. it will be downloaded by the s3 sync rehydration stage, but when the psql command has to run to restore, will to look for the artifact in a well known path? Or does the SSM playbook have to download the database file from S3 separately?

We can continue this discussion on slack or Jira if that's where this discussion should belong.

bpar476:
> How do we see the flow?
> 
>     1. The plugin takes a backup of the database
> 
>     2. The plugin then encrypts the credentials (username & password)
> 
>     3. The SSM play on the migration stack uses the backup file and credentials to perform the 

1. The credentials are encrypted when the application stack is provisioned.
2. The credentials are supplied as secure parameters to the migration stack when it is deployed
3. The migration stack stores these credentials in AWS Secrets Manager

bpar476:
> Open Questions -
> 
>     1. How does the decryption happen?
> 
>     2. How is the database file identified in the S3 bucket? i.e. it will be downloaded by the s3 sync rehydration stage, but when the psql command has to run to restore, will to look for the artifact in a well known path? Or does the SSM playbook have to download the database file from S3 separately?
> 
> 
> We can continue this discussion on slack or Jira if that's where this discussion should belong.

Probably best resolved on Slack or Zoom

PR #87, opened 2020-03-30
CHET-180: Refactor out the SchedulerService code from FS service for reuse in DB service.

tarka:


tarka:
> LGTM - There are a few commented lines that you can remove (references to the pre-refactoring service)

Good point, thanks.

PR #88, opened 2020-03-30
CHET-202: Securely store new db credentials

bpar476:
# Summary

- Stores database password encrypted in the migration context.
- Hardcodes the database user

## Fast Follow

- Refactor access to the migration context so it goes through the MigrationService
- Make the relationship between migration context and migration bidirectional

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
@@ -77,7 +77,7 @@ public class QuickstartDeploymentService implements ApplicationDeploymentService
     }
 
     private void storeDbCredentials(Map<String, String> params) {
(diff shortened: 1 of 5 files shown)


src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
```diff
@@ -69,6 +72,12 @@ public void deployApplication(String deploymentId, Map<String, String> params) t
         addDeploymentIdToMigrationContext(deploymentId);
 
         scheduleMigrationServiceTransition(deploymentId);
+
+        storeDbCredentials(params);
+    }
+
+    private void storeDbCredentials(Map<String, String> params) {
+        dbCredentialsStorageService.storeCredentials("atljira", params.get("DBPassword"));
```
theghostwhoforks: perhaps a constant or an env variable?

PR #90, opened 2020-03-31
CHET-190: pass target DB credentials to secretsmanager

bpar476:
# Summary

* Store database password in AWS secrets manager instead of migration context

## Reason

Originally we were going to use the migration context to provide the DB password as a parameter to the migration stack CloudFormation template. However, this is not recommended by AWS ([source](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/best-practices.html#creds)). Instead we store the credentials as a string with Secrets Manager under the name:
`com.atlassian.migration.db.target.[migration_deployment_id].applicationPassword`
where `migration_deployment_id` is the ApplicationDeploymentId of the current migration context

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/TargetDbCredentialsStorageService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/TargetDbCredentialsStorageService.java
@@ -60,7 +60,7 @@ public class TargetDbCredentialsStorageService {
 
         SecretsManagerClient client = clientFactory.get();
         CreateSecretRequest request = CreateSecretRequest.builder()
-                .name(String.format("com.atlassian.migration.db.target.%s.applicationPassword", context.getApplicationDeploymentId()))
+                .name(String.format("atl-%s-app-rds-password", context.getApplicationDeploymentId()))
                 .secretString(password)
                 .description("password for the application user in you new AWS deployment")
                 .build();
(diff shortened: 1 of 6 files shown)


src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/TargetDbCredentialsStorageService.java
```diff
@@ -31,44 +37,42 @@
...
-        this.ao = ao;
-        this.encryptionManager = encryptionManager;
     }
 
     /**
-     * Stores the given database password in the {@link com.atlassian.migration.datacenter.dto.MigrationContext}
-     * to be used later to restore the database. The password will be encrypted before storage.
+     * Stores the given database password in AWS <a href="https://aws.amazon.com/secrets-manager/">Secrets Manager</a>
+     * to be used later to restore the database. Will be stored under the key:
+     *          com.atlassian.migration.db.target.[migration_deployment_id].applicationPassword
```
theghostwhoforks: Reference - https://github.com/atlassian-labs/dc-migration-assistant/blob/c0c00e3d3528d17a2d86bc3a09840893eccd653f/templates/migration-stack/migration-helper.yml#L331

1. What is an example of a migration_deployment_id? i.e. Will we have access to form it at the time of template provisioning? 
2. Can we make it smaller to make the secret key  the secret access policy (linked above). Perhaps, calling it just `atl-${migration-deployment-id}-app-rds-password` may suffice? (I don't know if that IAM ARN reference has a limit on the key size :) )

bpar476: The migration deployment id is the name of the cloudformation stack. We can make a smaller name, AWS is known for weird character-size limits.

bpar476: Done!

PR #91, opened 2020-03-31
CHET-180: DB Migration Endpoint

tarka:


src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -18,18 +18,53 @@
...
+
 /**
  * Copyright Atlassian: 10/03/2020
  */
-public class DatabaseMigrationService {
+public class DatabaseMigrationService
+{
+    private static Logger logger = LoggerFactory.getLogger(DatabaseMigrationService.class);
+
+    // FIXME: Should really be delegated to the state machine.
```
bpar476: Is there any reason this can't be delegated already? We have similar status in the `MigrationStage` enum

tarka: You're right, I've missed a stash pop somewhere I think.

src/main/java/com/atlassian/migration/datacenter/spi/MigrationStage.java
```diff
@@ -68,6 +68,11 @@ public boolean isValidTransition(MigrationStage to) {
                 .orElse(true);
     }
 
+    public boolean isDBPhase() {
+        // Hacky, but OK for now.
+        return this.toString().startsWith("DB_");
```
theghostwhoforks, line 73: should this be any db_phase or anything more specific like export only?

PR #92, opened 2020-03-31
Chet 189 ssm rds restore

theghostwhoforks:
Just retrieves secret from secretsstore and syncs bucket containing the database dump.

Next steps (separate PR)
- Accept RDS endpoint as part of migration template
- exec psql restore from EFS to RDS 
- Monitor command output (?)
-- Follow pattern similar to FS restore
-- Check if psql has a progress option that can we could use?

### DIFF
--- a/templates/migration-stack/migration-helper.yml
+++ b/templates/migration-stack/migration-helper.yml
@@ -123,7 +123,7 @@ Resources:
                     #!/bin/bash
                     echo "this script will pull the secret from secrets manager"
                     echo "A RDS endpoint needs to be passed into the template"
-                    aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region}
+                    aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"
                     aws s3 ls s3://${MigrationBucket}/db.dump/ --region ${AWS::Region}
                     echo "${RDSEndPoint}"
                   - { RDSEndPoint: "rds-dns-address", SecretIdentifier: "atl-qs-rds-password"}
--- a/templates/migration-stack/migration-helper.yml
+++ b/templates/migration-stack/migration-helper.yml
@@ -44,7 +44,7 @@ Resources:
     Metadata:
         AWS::CloudFormation::Init:
           configSets:
-            install_prerequisites: [create_efs_script,mount_efs,prepare_efs_sync_script,install_psql_client]
+            install_prerequisites: [create_efs_script,mount_efs,prepare_efs_sync_script,install_psql_client, create_rds_restore_script]
           create_efs_script:
             files:
               /opt/atlassian/dc-migration-assistant/mount-efs.sh:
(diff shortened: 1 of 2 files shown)


templates/migration-stack/migration-helper.yml
```diff
@@ -115,6 +115,21 @@ Resources:
...
+                content: !Sub
+                  - |
+                    #!/bin/bash
+                    echo "this script will pull the secret from secrets manager"
+                    echo "A RDS endpoint needs to be passed into the template"
+                    DATABASE_DOWNLOAD_DIR="/efs/downloads/db.dump"
+                    mkdir -p $DATABASE_DOWNLOAD_DIR
+                    aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"
+                    aws s3 sync s3://${MigrationBucket}/db.dump/ $DATABASE_DOWNLOAD_DIR --region ${AWS::Region}
+                    echo "${RDSEndPoint}"
```
bpar476: Are we going to invoke `pg_restore` here?

templates/migration-stack/migration-helper.yml
```diff
@@ -115,6 +115,21 @@ Resources:
...
                   - "amazon-linux-extras install -y postgresql${PSQLClientVersion}"
                   - { PSQLClientVersion: !Ref PSQLVersion }
+          create_rds_restore_script:
+            files:
+              /opt/atlassian/dc-migration-assistant/restore-db-to-rds.sh:
+                content: !Sub
+                  - |
+                    #!/bin/bash
+                    echo "this script will pull the secret from secrets manager"
+                    echo "A RDS endpoint needs to be passed into the template"
```
bpar476: Should we create a string parameter for this?

templates/migration-stack/migration-helper.yml
```diff
@@ -301,13 +316,22 @@ Resources:
...
-           }]
+                }]
+        - PolicyName: MigrationInstanceSecretReadAccess
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Action:
+                  - 'secretsmanager:GetSecretValue'
+                Effect: Allow
+                Resource: !Sub ['arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:${SecretIdentifier}-??????', { SecretIdentifier: "atl-qs-rds-password"
```
bpar476: What's the action to get rid of the questionmarks?

templates/migration-stack/migration-helper.yml
```diff
@@ -450,6 +475,23 @@ Resources:
...
+      Content:
+        schemaVersion: "2.2"
+        description: "This document is used by the Atlassian DC Migration Assistant to restore the database backup in S3 to a provisioned RDS instance"
+        mainSteps:
+        - action: "aws:runShellScript"
+          name: "restoreDatabaseBackupToRDS"
+          inputs:
+            runCommand:
+            - "#!/bin/bash"
+            - runuser -l jira -c 'at -f /opt/atlassian/dc-migration-assistant/restore-db-to-rds.sh now'
```
bpar476: We probably don't need to use `at` to run this command as the db restore typically takes less than the maximum ssm timeout of 48 hours. I would just invoke the script directly. This means the status of the SSM command is equal to the status of the database restore.

templates/migration-stack/migration-helper.yml
theghostwhoforks: yeah. I'll follow that with another PR. I just wanted to exec the aws secretsmanager and get the secret out in this one

templates/migration-stack/migration-helper.yml
theghostwhoforks: yes. Will do as a quick follow

templates/migration-stack/migration-helper.yml
theghostwhoforks: So, the secretsmanager ARN  looks like this - 
`arn:aws:secretsmanager:us-east-2:887764444972:secret:atl-qs-rds-password-AVoCx9`

This policy matches the key that we specify and the random 6 digit string suffix that secretsmanager adds when a secret is stored. This policy is tightly locked to retrieve just one key from the secrets store

https://docs.aws.amazon.com/secretsmanager/latest/userguide/auth-and-access_identity-based-policies.html#available-managed-policies

templates/migration-stack/migration-helper.yml
theghostwhoforks: True. Copy pasta will be corrected!

templates/migration-stack/migration-helper.yml
bpar476: Ah I see. Why do they add that suffix 🤔

PR #94, opened 2020-04-01
CHET-140: Add abort incomplete multipart upload lifecycle policy

bpar476:
# Summary

* Adds bucket lifecycle policy to abort incomplete multipart uploads after 7 days

## Screenshots for proof
Screen Shot 2020-04-01 at 11 10 16 am
Screen Shot 2020-04-01 at 11 10 22 am

### DIFF
--- a/templates/migration-stack/pkg/migration-helper.yml.src
+++ b/templates/migration-stack/pkg/migration-helper.yml.src
@@ -43,7 +43,7 @@ Resources:
       LifecycleConfiguration:
         Rules:
           - AbortIncompleteMultipartUpload:
-              DaysAfterInitiation: 7
+              DaysAfterInitiation: 4
             Status: "Enabled"
     DeletionPolicy: Delete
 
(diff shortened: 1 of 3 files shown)


templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -40,6 +40,11 @@ Resources:
         ServerSideEncryptionConfiguration:
           - ServerSideEncryptionByDefault:
               SSEAlgorithm: AES256
+      LifecycleConfiguration:
+        Rules:
+          - AbortIncompleteMultipartUpload:
+              DaysAfterInitiation: 7
```
theghostwhoforks, line 46: Why `7`?

bpar476: I wanted to make sure it wouldn't kill an in-progress multipart upload. It's very unlikely any individual object would take the full 7 days to upload, maybe I can trim it down to 3 days

PR #95, opened 2020-04-01
Chet 189 ssm rds restore

theghostwhoforks:
1. restores a database dump present in the S3 bucket.
2. Contains slight refactoring of the template (duplication in securrity group custom resource lambda)
3. Changes delete policy of S3 bucket to retain - Until a new delete hook is added to clean up the bucket on stack delete

Successful run of the latest version of the command with an automatically created security group, credentials retrieved from secretsmanager

Screenshot 2020-04-01 at 16 56 07

### DIFF
--- a/templates/migration-stack/migration-helper.yml
+++ b/templates/migration-stack/migration-helper.yml
@@ -403,7 +403,7 @@ Resources:
                   rdssg =  event['ResourceProperties']['RDSSG']
                   helper_sg = event['ResourceProperties']['HelperSG']
                   if event['RequestType'] == 'Delete':
-                      for security_group_id in [efssg, rdssg]:
+                      for security_group_id in list(filter(None, [efssg, rdssg])):
                         response = ec2client.revoke_security_group_ingress(
                             GroupId=security_group_id,
                             IpPermissions=[
@@ -417,7 +417,7 @@ Resources:
                       responseData = {'Delete': 'SUCCESS'}
                       cfnresponse.send(event, context, cfnresponse.SUCCESS, responseData)
                   if event['RequestType'] == 'Create':
-                      for security_group_id in [efssg, rdssg]:
+                      for security_group_id in list(filter(None, [efssg, rdssg])):
                         response = ec2client.authorize_security_group_ingress(
                             GroupId=security_group_id,
                             IpPermissions=[
--- a/templates/migration-stack/migration-helper.yml
+++ b/templates/migration-stack/migration-helper.yml
@@ -123,7 +123,7 @@ Resources:
                     #!/bin/bash
                     echo "this script will pull the secret from secrets manager"
                     echo "A RDS endpoint needs to be passed into the template"
-                    aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region}
+                    aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"
                     aws s3 ls s3://${MigrationBucket}/db.dump/ --region ${AWS::Region}
                     echo "${RDSEndPoint}"
                   - { RDSEndPoint: "rds-dns-address", SecretIdentifier: "atl-qs-rds-password"}
--- a/templates/migration-stack/migration-helper.yml
+++ b/templates/migration-stack/migration-helper.yml
@@ -407,40 +407,40 @@ Resources:
           ec2client = boto3.client('ec2')
           def lambda_handler(event, context):
             try:
-                  efssg =  {'sg': event['ResourceProperties']['EFSSG'], 'from': 2049, 'to': 2049 }
-                  rdssg =  {'sg': event['ResourceProperties']['RDSSG'], 'from': 5432, 'to': 5432 }
-                  helper_sg = event['ResourceProperties']['HelperSG']
-                  if event['RequestType'] == 'Delete':
-                      for security_group_props in [efssg, rdssg]:
-                        response = ec2client.revoke_security_group_ingress(
-                            GroupId=security_group_props['sg'],
-                            IpPermissions=[
-                                {'IpProtocol': 'tcp',
-                                'FromPort': security_group_props['from'],
-                                'ToPort': security_group_props['to'],
-                                'UserIdGroupPairs': [{'GroupId': helper_sg}]}
-                            ]
-                        )
-                        print(response)
-                      responseData = {'Delete': 'SUCCESS'}
-                      cfnresponse.send(event, context, cfnresponse.SUCCESS, responseData)
-                  if event['RequestType'] == 'Create':
-                      for security_group_id in [efssg, rdssg]:
-                        response = ec2client.authorize_security_group_ingress(
-                            GroupId=security_group_props['sg'],
-                            IpPermissions=[
-                                {'IpProtocol': 'tcp',
-                                'FromPort': security_group_props['from'],
-                                'ToPort': security_group_props['to'],
-                                'UserIdGroupPairs': [{'GroupId': helper_sg}]}
-                            ]
-                        )
-                        print(response)
-                      responseData = {'Create': 'SUCCESS'}
-                      cfnresponse.send(event, context, cfnresponse.SUCCESS, responseData)
-                  if event['RequestType'] == 'Update':
-                      responseData = {'Update': 'SUCCESS'}
-                      cfnresponse.send(event, context, cfnresponse.SUCCESS, responseData)
+              efssg =  {'sg': event['ResourceProperties']['EFSSG'], 'from': 2049, 'to': 2049 }
+              rdssg =  {'sg': event['ResourceProperties']['RDSSG'], 'from': 5432, 'to': 5432 }
+              helper_sg = event['ResourceProperties']['HelperSG']
+              if event['RequestType'] == 'Delete':
+                for security_group_props in [efssg, rdssg]:
+                  response = ec2client.revoke_security_group_ingress(
+                      GroupId=security_group_props['sg'],
+                      IpPermissions=[
+                          {'IpProtocol': 'tcp',
+                          'FromPort': security_group_props['from'],
+                          'ToPort': security_group_props['to'],
+                          'UserIdGroupPairs': [{'GroupId': helper_sg}]}
+                      ]
+                  )
+                  print(response)
+                responseData = {'Delete': 'SUCCESS'}
+                cfnresponse.send(event, context, cfnresponse.SUCCESS, responseData)
+              if event['RequestType'] == 'Create':
+                for security_group_id in [efssg, rdssg]:
+                  response = ec2client.authorize_security_group_ingress(
+                      GroupId=security_group_props['sg'],
+                      IpPermissions=[
+                          {'IpProtocol': 'tcp',
+                          'FromPort': security_group_props['from'],
+                          'ToPort': security_group_props['to'],
+                          'UserIdGroupPairs': [{'GroupId': helper_sg}]}
+                      ]
+                  )
+                  print(response)
+                responseData = {'Create': 'SUCCESS'}
+                cfnresponse.send(event, context, cfnresponse.SUCCESS, responseData)
+              if event['RequestType'] == 'Update':
+                responseData = {'Update': 'SUCCESS'}
+                cfnresponse.send(event, context, cfnresponse.SUCCESS, responseData)
             except Exception as e:
                 responseData = {'Error': str(e)}
                 cfnresponse.send(event, context, cfnresponse.FAILED, responseData)
--- a/templates/migration-stack/migration-helper.yml
+++ b/templates/migration-stack/migration-helper.yml
@@ -399,17 +399,17 @@ Resources:
           ec2client = boto3.client('ec2')
           def lambda_handler(event, context):
             try:
-                  efssg =  event['ResourceProperties']['EFSSG']
-                  rdssg =  event['ResourceProperties']['RDSSG']
+                  efssg =  {'sg': event['ResourceProperties']['EFSSG'], 'from': 2049, 'to': 2049 }
+                  rdssg =  {'sg': event['ResourceProperties']['RDSSG'], 'from': 5432, 'to': 5432 }
                   helper_sg = event['ResourceProperties']['HelperSG']
                   if event['RequestType'] == 'Delete':
-                      for security_group_id in set(filter(None, [efssg, rdssg])):
+                      for security_group_props in [efssg, rdssg]:
                         response = ec2client.revoke_security_group_ingress(
-                            GroupId=security_group_id,
+                            GroupId=security_group_props['sg'],
                             IpPermissions=[
                                 {'IpProtocol': 'tcp',
-                                'FromPort': 2049,
-                                'ToPort': 2049,
+                                'FromPort': security_group_props['from'],
+                                'ToPort': security_group_props['to'],
                                 'UserIdGroupPairs': [{'GroupId': helper_sg}]}
                             ]
                         )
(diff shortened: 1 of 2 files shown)


templates/migration-stack/migration-helper.yml
```diff
@@ -118,6 +132,29 @@ Resources:
...
+                    DATABASE_DOWNLOAD_DIR="/efs/downloads/db.dump"
+                    mkdir -p $DATABASE_DOWNLOAD_DIR
+                    DB_DUMP_LOG_FILE="/var/atlassian/dc-migration-assistant/pg_dump-log.txt"
+                    SECRET_PASSWORD=`aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"`
+                    aws s3 sync s3://${MigrationBucket}/db.dump/ $DATABASE_DOWNLOAD_DIR --region ${AWS::Region} | tee $DB_DUMP_LOG_FILE
+                    echo "Restoring database from $DATABASE_DOWNLOAD_DIR to ${DBHost}:${DBPort}/$DBName" | tee $DB_DUMP_LOG_FILE
+                    PGPASSWORD=$SECRET_PASSWORD pg_restore -h ${DBHost} -U ${DBUser} -d ${DBName} -p ${DBPort} -F d -j 6 --verbose $DATABASE_DOWNLOAD_DIR 2>&1 | tee $DB_DUMP_LOG_FILE
+                    echo "Monitor the logfile at $DB_DUMP_LOG_FILE for the term 'pg_restore: finished main parallel loop' to appear to signal end of restore process"
+                  - {
+                    SecretIdentifier: !Sub "atl-${AWS::StackId}-app-rds-password",
```
theghostwhoforks, line 150: I'm assuming that the `context.getApplicationDeploymentId()` returns the stackId. I think it'll return the name, in which case, we'll need to replace this with ${AWS::StackName}

cc: @bpar476

templates/migration-stack/migration-helper.yml
```diff
@@ -304,13 +341,32 @@ Resources:
...
-           }]
+                }]
+        - PolicyName: MigrationInstanceSecretReadAccess
+          PolicyDocument:
+            Version: 2012-10-17
+            Statement:
+              - Action:
+                  - 'secretsmanager:GetSecretValue'
+                Effect: Allow
+                Resource: !Sub ['arn:aws:secretsmanager:${AWS::Region}:${AWS::AccountId}:secret:${SecretIdentifier}-??????', { SecretIdentifier: "atl-qs-rds-password"
```
theghostwhoforks, line 358: This needs to change to incorporate the stackId in the secret identifier.

templates/migration-stack/migration-helper.yml
```diff
@@ -346,49 +402,33 @@ Resources:
...
-                          GroupId=rdssg,
-                          IpPermissions=[
-                              {'IpProtocol': 'tcp',
-                              'FromPort': 5432,
-                              'ToPort': 5432,
-                              'UserIdGroupPairs': [{'GroupId': evsg}]}
-                          ]
-                      )
-                      print(response)
+                      for security_group_id in [efssg, rdssg]:
```
theghostwhoforks, line 407: Will verify if this indentation is correct 😬

templates/migration-stack/migration-helper.yml
theghostwhoforks: This has to be stack name as the key with stack Id looks like this  -

`atl-arn:aws:cloudformation:<region>:<account>:stack/migration-helper-rds-002/50b0dd90-73b3-11ea-9ba8-027927fac094-app-rds-password`

which is invalid

bpar476: The application deployment ID will be the stack name of the quick start

templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -254,49 +309,33 @@ Resources:
...
-                          GroupId=rdssg,
-                          IpPermissions=[
-                              {'IpProtocol': 'tcp',
-                              'FromPort': 5432,
-                              'ToPort': 5432,
-                              'UserIdGroupPairs': [{'GroupId': evsg}]}
-                          ]
-                      )
-                      print(response)
+                      for security_group_id in set(filter(None, [efssg, rdssg])):
```
theghostwhoforks, line 314: In quickstart stacks, the EFS and RDS security groups are the same. I've verified this with a jira and confluence stack independently.

In light of this, we should either -

1. Get rid of one of the 2 security group parameters and replace with one with a generic name
2. or accept empty values for one of the security groups and filter it before invoking the ec2 APIs
3. or accept duplicate and remove the duplicate value by adding the values to a set and ensure the API call happens with the unique security group

templates/migration-stack/migration-helper.yml
```diff
@@ -118,6 +132,29 @@ Resources:
...
+                content: !Sub
+                  - |
+                    #!/bin/bash
+                    DATABASE_DOWNLOAD_DIR="/efs/downloads/db.dump"
+                    mkdir -p $DATABASE_DOWNLOAD_DIR
+                    DB_DUMP_LOG_FILE="/var/atlassian/dc-migration-assistant/pg_dump-log.txt"
+                    SECRET_PASSWORD=`aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"`
+                    aws s3 sync s3://${MigrationBucket}/db.dump/ $DATABASE_DOWNLOAD_DIR --region ${AWS::Region} | tee $DB_DUMP_LOG_FILE
+                    echo "Restoring database from $DATABASE_DOWNLOAD_DIR to ${DBHost}:${DBPort}/$DBName" | tee $DB_DUMP_LOG_FILE
+                    PGPASSWORD=$SECRET_PASSWORD pg_restore -h ${DBHost} -U ${DBUser} -d ${DBName} -p ${DBPort} -F d -j 6 --verbose $DATABASE_DOWNLOAD_DIR 2>&1 | tee $DB_DUMP_LOG_FILE
```
bpar476, line 147: Where does the number of jobs come from? I think we probably don't want to have more jobs than CPUs do we?

templates/migration-stack/migration-helper.yml
```diff
@@ -440,6 +479,42 @@ Resources:
...
+        schemaVersion: "2.2"
+        description: "This document is used by the Atlassian DC Migration Assistant to restore the database backup in S3 to a provisioned RDS instance"
+        mainSteps:
+        - action: "aws:runShellScript"
+          name: "restoreDatabaseBackupToRDS"
+          inputs:
+            runCommand:
+            - "#!/bin/bash"
+            - runuser -l jira -c '/opt/atlassian/dc-migration-assistant/restore-db-to-rds.sh'
+            timeoutSeconds: "60" #TODO: What is the correct value for timeout seconds here?
```
bpar476, line 495: The timeout should be long enough to support a large database restore. I would set it to at least 12 hours.

templates/migration-stack/migration-helper.yml
```diff
@@ -346,49 +401,33 @@ Resources:
...
-                          GroupId=rdssg,
-                          IpPermissions=[
-                              {'IpProtocol': 'tcp',
-                              'FromPort': 5432,
-                              'ToPort': 5432,
-                              'UserIdGroupPairs': [{'GroupId': evsg}]}
-                          ]
-                      )
-                      print(response)
+                      for security_group_id in set(filter(None, [efssg, rdssg])):
```
bpar476, line 406: Nice! Why do we need `set(filter(...` can't we just do `for ... in [efssg, rdssg]`?

templates/migration-stack/migration-helper.yml
```diff
@@ -118,6 +132,29 @@ Resources:
...
+                  - |
+                    #!/bin/bash
+                    DATABASE_DOWNLOAD_DIR="/efs/downloads/db.dump"
+                    mkdir -p $DATABASE_DOWNLOAD_DIR
+                    DB_DUMP_LOG_FILE="/var/atlassian/dc-migration-assistant/pg_dump-log.txt"
+                    SECRET_PASSWORD=`aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"`
+                    aws s3 sync s3://${MigrationBucket}/db.dump/ $DATABASE_DOWNLOAD_DIR --region ${AWS::Region} | tee $DB_DUMP_LOG_FILE
+                    echo "Restoring database from $DATABASE_DOWNLOAD_DIR to ${DBHost}:${DBPort}/$DBName" | tee $DB_DUMP_LOG_FILE
+                    PGPASSWORD=$SECRET_PASSWORD pg_restore -h ${DBHost} -U ${DBUser} -d ${DBName} -p ${DBPort} -F d -j 6 --verbose $DATABASE_DOWNLOAD_DIR 2>&1 | tee $DB_DUMP_LOG_FILE
+                    echo "Monitor the logfile at $DB_DUMP_LOG_FILE for the term 'pg_restore: finished main parallel loop' to appear to signal end of restore process"
```
bpar476, line 148: Isn't the restore synchronous so we can just wait or the script to finish to signal the end of the restore?

templates/migration-stack/migration-helper.yml
```diff
@@ -440,6 +479,42 @@ Resources:
...
+          name: "restoreDatabaseBackupToRDS"
+          inputs:
+            runCommand:
+            - "#!/bin/bash"
+            - runuser -l jira -c '/opt/atlassian/dc-migration-assistant/restore-db-to-rds.sh'
+            timeoutSeconds: "60" #TODO: What is the correct value for timeout seconds here?
+            workingDirectory: "/opt/atlassian/dc-migration-assistant/"
+      DocumentType: "Command"
+
+  RdsRestoreStatusDocument:
```
bpar476, line 499: Why do we need this when we can just use the return value of `pg_restore` to see if there were errors?

templates/migration-stack/pkg/migration-helper.yml.src
theghostwhoforks: In the haste of refactoring, I've missed the DB port. Fixing that

bpar476: I think there's no harm in keeping the extra parameter. It keeps it open if we allow skipping the quick start deployment and it's not difficult to supply an extra parameter to the cfn template.

templates/migration-stack/migration-helper.yml
theghostwhoforks: i just set j to 6. I can tune it down to 1 or look at `/proc/cpuinfo`

I'll remove the parameter and let pg_restore do it's default thing

templates/migration-stack/migration-helper.yml
theghostwhoforks: We could. If it runs for a long time, then we can just get the error/success count from the file.

Which is better than creating a new command

templates/migration-stack/migration-helper.yml
theghostwhoforks: When it failed, due to table colissions, the return value was still 0, if I recollect correctly.

Anyway, this document will be removed as it's more of a hassle to write. a multi-line bash script in a SSM command 🤦‍♂

templates/migration-stack/migration-helper.yml
theghostwhoforks: I've made the change to ensure we add both the EFS + RDS port into the security group

bpar476: why does it matter if it's in a `set` or just a `[]`

templates/migration-stack/migration-helper.yml
bpar476: If we want to report errors, we could just read the file in this command and print them to standard out to be parsed by the app.

templates/migration-stack/migration-helper.yml
```diff
@@ -130,14 +143,25 @@ Resources:
...
+                    aws s3 sync s3://${MigrationBucket}/db.dump/ $DATABASE_DOWNLOAD_DIR --region ${AWS::Region} | tee $DB_DUMP_LOG_FILE
+                    echo "Restoring database from $DATABASE_DOWNLOAD_DIR to ${DBHost}:${DBPort}/$DBName" | tee $DB_DUMP_LOG_FILE
+                    PGPASSWORD=$SECRET_PASSWORD pg_restore -h ${DBHost} -U ${DBUser} -d ${DBName} -p ${DBPort} -F d -j 6 --verbose $DATABASE_DOWNLOAD_DIR 2>&1 | tee $DB_DUMP_LOG_FILE
+                    PG_RESTORE_EXIT_CODE=$?
+                    ERRORS_EXIST=`grep -qiE 'error|warning' /var/atlassian/dc-migration-assistant/pg_dump-log.txt && echo 'true' || echo 'false'`
+                    RESTORE_COMPLETE=`grep -qiE 'pg_restore: finished main parallel loop' /var/atlassian/dc-migration-assistant/pg_dump-log.txt && echo 'true' || echo 'false'`
+                    echo -e "{\"is_restore_complete\":\""$RESTORE_COMPLETE"\",\"is_error_present\":\""$ERRORS_EXIST"\", "return_code":\""$PG_RESTORE_EXIT_CODE"\"}"
+                  - {
+                    SecretIdentifier: !Sub "atl-${AWS::StackName}-app-rds-password",
```
bpar476, line 158: I think you'll need a parameter for this because the database secret is stored with the quick start stack name as the prefix, not the migration stack name.

templates/migration-stack/migration-helper.yml
theghostwhoforks: I've updated the code just now and pushed. Can you verify if you're still seeing the set invocation?

templates/migration-stack/migration-helper.yml
theghostwhoforks: Done. Update the restore command to emit a json doc at the end

templates/migration-stack/migration-helper.yml
theghostwhoforks: But that can be changed on the server side, correct? 

Because we control the name of the parameter that goes in the migration stack variable?

theghostwhoforks: I just wanted to avoid a parameter, but if you reckon it's not possible, then I'll add a parameter

templates/migration-stack/migration-helper.yml
bpar476: Oh I was looking at the `outdated` preview 🤦‍♂

templates/migration-stack/migration-helper.yml
bpar476: That would mean we need to cache the database password on the app before uploading it to secretsmanager as we discover the db password when we deploy the application stack but don't know the migration stack name until later when we deploy the migration stack 🤔

templates/migration-stack/migration-helper.yml
theghostwhoforks: No issues!

The main issue that was trying to resolve was that the security group for EFS and RDS is the same, and I wanted to understand why we have an additional parameter to accept RDS SG, when in the case of quickstarts, it will always be equal to the EFSSG? 🤔 

Right now, I take both SGs in a dict and iterate over them. But for someone entering this migration stack form manually, they'll have to add the same security group twice, into the EFS SG and RDS SG parameter

templates/migration-stack/migration-helper.yml
theghostwhoforks: Don't we always know the migration stack name upfront? or do we generate that dynamically?

Also, can we table this additional parameter discussion and resolve it later? I'd like to get this PR merged before EOD 🙏

bpar476: Okey dokey 👍

templates/migration-stack/migration-helper.yml
bpar476: Noone should have to use the form. Devs can use the makefile and the app will use the API. If it does add significant overhead we can cut the parameter but I think keeping them separate is a good separation of concerns. In the future we may want to support bring-your-own-infra where this assumption will no-longer be correct and I don't think it costs us a lot to keep the parameter there.

templates/migration-stack/migration-helper.yml
bpar476: The migration stack name could be known up-front but it still has to be unique in case of a failure. However, I think it's undesirable to create a semantic coupling between the secrets manager key name and however we generate the migration stack name.

PR #96, opened 2020-04-01
CHET-115: deploy migration stack service

bpar476:
# Summary

* Takes logic from quick start deployment and hoists it into a superclass
* Quickstart and migration template deployment inherit from this class
* Use template method to handle migration stage transitions

## Follow-ups
* Integrating migration stack deployment with state machine

### DIFF
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
@@ -16,6 +16,7 @@
 
 package com.atlassian.migration.datacenter.core.aws.infrastructure;
 
+import com.atlassian.migration.datacenter.core.aws.CfnApi;
 import com.atlassian.migration.datacenter.core.exceptions.InvalidMigrationStageError;
 import com.atlassian.migration.datacenter.spi.infrastructure.InfrastructureDeploymentStatus;
 import com.atlassian.migration.datacenter.spi.infrastructure.MigrationInfrastructureDeploymentService;
@@ -27,9 +28,18 @@ import java.util.Map;
  * application deployment with data.
  */
 public class AWSMigrationHelperDeploymentService implements MigrationInfrastructureDeploymentService {
+
+    private static final String MIGRATION_HELPER_TEMPLATE_URL = "https://trebuchet-aws-resources.s3.amazonaws.com/migration-helper.yml";
+
+    private final CfnApi cfnApi;
+
+    public AWSMigrationHelperDeploymentService(CfnApi cfnApi) {
+        this.cfnApi = cfnApi;
+    }
+
     @Override
     public void deployMigrationInfrastructure(String deploymentId, Map<String, String> params) throws InvalidMigrationStageError {
-
+        cfnApi.provisionStack(MIGRATION_HELPER_TEMPLATE_URL, deploymentId, params);
     }
 
     @Override
--- a/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
+++ b/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
@@ -18,6 +18,7 @@ package com.atlassian.migration.datacenter.core.aws.infrastructure;
 
 import com.atlassian.migration.datacenter.core.aws.CfnApi;
 import com.atlassian.migration.datacenter.core.exceptions.InvalidMigrationStageError;
+import com.atlassian.migration.datacenter.dto.MigrationContext;
 import com.atlassian.migration.datacenter.spi.MigrationService;
 import com.atlassian.migration.datacenter.spi.infrastructure.InfrastructureDeploymentStatus;
 import com.atlassian.migration.datacenter.spi.infrastructure.MigrationInfrastructureDeploymentService;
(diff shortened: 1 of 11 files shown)


src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
```diff
@@ -0,0 +1,66 @@
...
+        super.deployCloudformationStack(MIGRATION_HELPER_TEMPLATE_URL, deploymentId, params);
+
+        final MigrationContext context = migrationService.getCurrentContext();
+        context.setHelperStackDeploymentId(deploymentId);
+        context.save();
+    }
+
+    @Override
+    protected void handleSuccessfulDeployment() {
+
```
theghostwhoforks, line 54: I'm assuming this is where the state machine hooks will be added later?

src/main/java/com/atlassian/migration/datacenter/spi/infrastructure/MigrationInfrastructureDeploymentService.java
```diff
@@ -0,0 +1,33 @@
...
+
+import com.atlassian.migration.datacenter.core.exceptions.InvalidMigrationStageError;
+
+import java.util.Map;
+
+/**
+ * This service should be implemented if a migration requires any additional infrastructure to facilitate
+ * the migration.
+ */
+public interface MigrationInfrastructureDeploymentService {
```
theghostwhoforks, line 27: This looks identical to `ApplicationDeploymentService`. Do we need this to be called out separately or collapsed into say, `InfrastructureDeploymentService` interface?

src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
```diff
@@ -0,0 +1,66 @@
...
+    private final MigrationService migrationService;
+
+    public AWSMigrationHelperDeploymentService(CfnApi cfnApi, MigrationService migrationService) {
+        super(cfnApi);
+        this.migrationService = migrationService;
+    }
+
+    @Override
+    public void deployMigrationInfrastructure(String deploymentId, Map<String, String> params) throws InvalidMigrationStageError {
```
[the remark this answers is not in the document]
bpar476: There is a mild risk that we hit cloudformation stack name length limits.

src/main/java/com/atlassian/migration/datacenter/spi/infrastructure/MigrationInfrastructureDeploymentService.java
bpar476: I thought it was important to have them as separate interfaces. I think each "component" of migration having its own interface makes it easier to plug different backends into each step without affecting others. Do we gain anything by consolidating the interfaces?

src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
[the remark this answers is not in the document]
theghostwhoforks: Perhaps, yeah. The 128 character limit, right? 🤔

I think we can kick the can down the road on this one?

src/main/java/com/atlassian/migration/datacenter/spi/infrastructure/MigrationInfrastructureDeploymentService.java
theghostwhoforks: If we are using this as a marker interface, which is OK!, then we can make one extend the other, instead of defining the methods again.

src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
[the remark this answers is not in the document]
bpar476: Agreed. I'll make the change to use the suffix and we can handle the character limit later

src/main/java/com/atlassian/migration/datacenter/spi/infrastructure/MigrationInfrastructureDeploymentService.java
bpar476: Then the methods have to have the same name 🤔

bpar476: And even though the code is similar, they are responsible for different things. Again, what's the **advantage** of consolidating the code?

theghostwhoforks: I don't see any new behaviour that deploy migration service interface is adding over the deploy application stack interface. I get why we need to have 2 different implementations, just not the interfaces. If we make a change to the interface that impacts both service implementations, consolidating to use the same method parameters may be useful.

If the migration service interface did something different when compared with the ApplicationDeploymentService interface, then I'd be in favour of keeping them apart. I understand the need to have marker interfaces to help in dependency injection. Which is why I suggested extending one interface.

Do you see any need to keep them separate (barring the method name bit that you mentioned above)?

bpar476: Seeing as the name is static now, the interface is different so this discussion is moot now. 

There is no code difference in the interface. The difference is purely semantic but I think that's important.

PR #98, opened 2020-04-02
RFC: Modularise all the things

tarka:
This adds the ability to componentise code into maven modules:

* core, frontend, plugin and SPI functionality have been split out.
* Spring/OSGI wiring is now part of the plugin module.
* The project group namespace has been clarified, but could use discussion.

The following modules should probably be created, but I wanted to time-box this today:

* The REST API.
* AWS-specific code should be separated; it's currently [dug in like a tick](https://gfycat.com/groundedcookedalaskankleekai).
* We should probably have a "cloud-spi" module to abstract cloud backends.

This should be considered an RFC; the code passes tests and generates the required OBR, but hasn't been extensively tested in-app.


theghostwhoforks:
I approve solely based on the archer reference

plugin/pom.xml
```diff
@@ -0,0 +1,154 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
theghostwhoforks, line 1: I'd call this jira-plugin right away

spi/pom.xml
```diff
@@ -0,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
theghostwhoforks, line 1: I'm confused with the `spi` naming convention and the contents of the package here (this is prior to this refactoring). The term itself is a bit overloaded [1] and ties in with Java SPI definition [2][3]

In our case, provision infrastructure is an interface that will be implemented by `AWS` and `Azure` differently.

Examples of Java SPI implementations, like `XmlParserFactory`, have an interface definition,registration and lookup mechanism that work to decouple interface definition and implementation, even when those implementations are not packaged in the same jar. [3]

With these in mind, I suggest that - 

1. `dto` and `exception` packages should not reside in the `SPI` module
2. Concrete implementations should not reside in `SPI`. i.e. `fs.FailedFileMigration`, should not be there
3. If we don't see ourselves using the service registration and discovery mechanism, then we should rename spi to `api` or `contract` or <enter better suggestion here> as it is can lead to more confusion if left as is.

I'm open to hearing different points of view on this that may help clarify some of the points raised above!

[1]. Our own [Atlassian-audit plugin has a SPI module](https://bitbucket.org/atlassian/atlassian-audit/src/master/atlassian-audit-spi/) that defines the contract between the plugin and the product. i.e. plugin interfaces that must be implemeted by the product.
[Wikipedia definition](https://en.wikipedia.org/wiki/Service_provider_interface) of SPI refers to service providers (like a specific XML parser implementation say) that is made available on the classpath 
[2] > A service provider framework is a system in which multiple service providers implement a service, and the system makes the implementations available to its clients, decoupling them from the implementations.
[3]https://www.baeldung.com/java-spi

plugin/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+            <version>1.0.1b</version>
+            <scope>provided</scope>
+        </dependency>
+
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>com.atlassian.maven.plugins</groupId>
```
theghostwhoforks, line 103: I built the project and it created a an artifact called `plugin-1.0.0.obr` - can we name the plugin module to plugin-jira and customize the name to have dc-migration-assistant prefix to it? 

i,e, `dc-migration-assistant-jira-plugin/app`

theghostwhoforks:
I've reviewed the project in idea and built and run tests locally

bpar476:
I love having all that frontend code tidied up. Having separate modules for everything is probably ideal to make sure we're disciplined with respecing SPI/core boundaries.

Just one nitpick.
* The template servlet is defined in `core` but the template it renders is in `plugin`. It think the soy template can probably be in the `code` module as it is just a container for the React frontend. Unlikely to have product-specific code.

If this doesn't require a code freeze on master, I'm a fan of moving to the multi-module architecture.

PR #100, opened 2020-04-02
CHET-116: integrate migration with migration stack

bpar476:
# Summary

* Modify existing migration services to pull the values for the migration infrastructure from the AWSMigrationInfrastructureDeploymentService

## NB

This pr is dependent on #99 It will be marked ready and re-targeted at master once that PR has been reviewed


core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -33,17 +34,13 @@
 import java.nio.file.Path;
 import java.util.concurrent.atomic.AtomicBoolean;
 
-import static com.atlassian.migration.datacenter.spi.fs.reporting.FilesystemMigrationStatus.FAILED;
-
 /**
```
nanux, line 37: /nit I believe this is incorrect copyright (not really your fault)

core/src/main/java/com/atlassian/migration/datacenter/core/fs/download/s3sync/S3SyncFileSystemDownloader.java
```diff
@@ -28,31 +29,28 @@
...
+    S3SyncFileSystemDownloader(SSMApi ssmApi, AWSMigrationHelperDeploymentService migrationHelperDeploymentService, int maxCommandStatusRetries) {
         this.ssmApi = ssmApi;
         this.maxCommandStatusRetries = maxCommandStatusRetries;
+        this.migrationHelperDeploymentService = migrationHelperDeploymentService;
     }
 
     public void initiateFileSystemDownload() throws CannotLaunchCommandException {
-        // FIXME: Reload the migration stack instance ID in case instance has gone down during migration
-        String commandID = ssmApi.runSSMDocument(SSM_PLAYBOOK, MIGRATION_STACK_INSTANCE, Collections.emptyMap());
+        String fsRestoreDocument = System.getProperty("com.atlassian.migration.s3sync.documentName", migrationHelperDeploymentService.getFsRestoreDocument());
```
nanux, line 48: Wouldn't be cleaner to encapsulate all `System.getProperty()` calls in the services into `migrationHelperDeploymentService`?
E.g. change `com.atlassian.migration.datacenter.core.aws.infrastructure.AWSMigrationHelperDeploymentService#getFsRestoreDocument` to:
```
public String getFsRestoreDocument() {
        final String documentOverride = System.getProperty("com.atlassian.migration.s3sync.documentName");
        if (documentOverride != null) {
            return documentOverride;
        } else {
            ensureStackOutputsAreSet();
            return fsRestoreDocument;
        }
    }
```

As this is repeated for few properties in the `AWSMigrationHelperDeploymentService`, it is possible to generify a private method that will either return the override or the property (while calling `ensureStackOutputsAreSet()`).
We would be also able to put all the property overrides in one file and extract into constant so they are not scattered across the code.

core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -62,12 +63,13 @@ public S3FilesystemMigrationService(Supplier<S3AsyncClient> s3AsyncClientSupplie
                                         JiraHome jiraHome,
                                         S3SyncFileSystemDownloadManager fileSystemDownloadManager,
                                         MigrationService migrationService,
-                                        MigrationRunner migrationRunner) {
+                                        MigrationRunner migrationRunner, AWSMigrationHelperDeploymentService migrationHelperDeploymentService) {
```
nanux, line 66: /nit should be new line

core/src/main/java/com/atlassian/migration/datacenter/core/fs/download/s3sync/S3SyncFileSystemDownloader.java
bpar476: I like this suggestion :)

PR #101, opened 2020-04-02
CHET-204: Add Kotlin support

nanux:
* Add support for Kotlin language to maven `pom`
* Converted `api` package (production + test code) to Kotlin
* Added `MockK` library for Kotlin style mocking and rewrote the tests

I have merged the latest `master` and converted new classes since the innovation. I have dropped all other work that I did in the innovation week - that is still living in `iw-kotlin` branch. The PR is looking massive but it is not really complicated as it shouldn't really change the functionality.

I am of two minds about moving all the code to `src/(main|test)/kotlin` directory. I think it is a cleaner solution and when using `package` view in IntelliJ IDEA you can see all classes in the package from Java and Kotlin. Right now I've kept it in the `java` folder (as it complicates the PR diff even further) but I am happy to move it.

Have a read and let me know if it is readable and in good quality to merge.


pom.xml
```diff
@@ -454,6 +476,50 @@
...
+                            <goal>test-compile</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <jvmTarget>1.8</jvmTarget>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
```
nanux, line 504: We need this because we compile `java` and `kotlin`.

src/test/java/com/atlassian/migration/datacenter/api/MigrationEndpointTest.kt
```diff
@@ -0,0 +1,100 @@
...
+@ExtendWith(MockKExtension::class)
+class MigrationEndpointTest {
+    @MockK
+    lateinit var migrationService: MigrationService
+
+    @InjectMockKs
+    lateinit var sut: MigrationEndpoint
+
+    @BeforeEach
+    fun init() = MockKAnnotations.init(this)
```
nanux, line 50: This with a combination of `lateinit var` was the recommended way how to do mocking with `MockK`

src/test/java/com/atlassian/migration/datacenter/api/develop/DevelopEndpointTest.kt
```diff
@@ -0,0 +1,82 @@
...
+        MockKAnnotations.init(this)
+    }
+
+    @Test
+    @Throws(Exception::class)
+    fun shouldSetStageWithCorrectTargetStageWhenProfileIsEnabled() {
+        val migrationStage = objectMapper.readValue("\"FS_MIGRATION_COPY\"", MigrationStage::class.java)
+        every { environment.activeProfiles } returns arrayOf(DevelopEndpoint.ALLOW_ANY_TRANSITION_PROFILE)
+        every { migrationService.currentStage } returns migrationStage
+        every { migrationService.transition(MigrationStage.FS_MIGRATION_COPY) } just Runs
```
nanux, line 60: It is necessary to mock interactions for Unit methods (`void`) or the mock needs to be relaxed with `@RelaxedMockK`

pom.xml
```diff
@@ -454,6 +476,50 @@
...
+                    <execution>
+                        <id>test-compile</id>
+                        <phase>test-compile</phase>
+                        <goals>
+                            <goal>test-compile</goal>
+                        </goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <jvmTarget>1.8</jvmTarget>
```
bpar476, line 500: Will this still work with Java 11?

src/test/java/com/atlassian/migration/datacenter/api/develop/DevelopEndpointTest.kt
theghostwhoforks: I like making interaction expectations explicit

nanux:
> As long as we can target java 8 and java 11 I'm happy with this.
> 
> 
> 
> Regarding `src/main/java` vs `src/main/kotlin`. I think it's fine to move this all to `src/main/kotlin` It makes it clearer where devs can expect to find Kotlin code in the future. Also, the entire package is in kotlin so it's not like the package is split across multiple directories.
> 
> 
> 
> I think this will be a diffifcult PR to merge along with the multi-module PR. Let's discuss this in stand-up.

I haven’t tested this java 11 but I will compile with it and test it. 

I will also move the code to kotlin directory.

nanux:
Moved all `kotlin` code to `kotlin` directory. Also tested with Java 11 (OpenJDK Runtime Environment AdoptOpenJDK (build 11.0.6+10))

PR #102, opened 2020-04-02
Fix name of db password secret in secrets manager

bpar476:
It will now be named 'atl-<migration-stack-name>-app-rds-password' so that the migration stack can pull it down easier.


src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/TargetDbCredentialsStorageService.java
```diff
@@ -60,7 +61,13 @@ public void storeCredentials(String password) {
 
         SecretsManagerClient client = clientFactory.get();
         CreateSecretRequest request = CreateSecretRequest.builder()
-                .name(String.format("atl-%s-app-rds-password", context.getApplicationDeploymentId()))
+                /*
+                The secret is named like this because it makes it easier for the migration stack to download
+                the secret. The migration stack name is static so we use it here even though it hasn't been
+                deployed yet. this allows the migration stack to use "atl-${AWS::StackName}-app-rds-password" as
```
nanux, line 67: /nit _. This_

src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/TargetDbCredentialsStorageService.java
```diff
@@ -60,7 +61,13 @@ public void storeCredentials(String password) {
...
+                the secret name.
+                */
+                .name(String.format("atl-%s-migration-app-rds-password", context.getApplicationDeploymentId()))
```
nanux, line 70: It does contain -migration- string in the name which is not documented. Is that expected?

bpar476: It is expected. `<application-stack-name>-migration` is the naming scheme for the migration stack. I'll document it in the API for the `AWSMigrationHelperDeploymentService`

PR #103, opened 2020-04-03
no-issue:

theghostwhoforks:
1. Move soy templates into front end
2. Remove plugin references from core
3. Move spring plugin xml to the plugin module
4. Update reference in template servlet to point to plugin 'GA' while rendering the template


theghostwhoforks:
Moved the soy template to front end
Removed atlassian plugin XML from core. This file is retained in the plugin module
Updated references in template servlet

PR #105, opened 2020-04-05
CHET-206: Break project into modules

tarka:
This is a re-do of the previous PR to pull in the Kotlin changes. This is the main structural work; there are other related changes required, but these can be fixed-up in-tree:
* README updates
* Clean-up dependencies, move version management to parent.
* AWS module split.


bpar476:
This works, I was able to run the UI standalone and within the plugin. However, Jira Software isn't installed when running amps 🤔

PR #106, opened 2020-04-06
CHET-223: Fix kotlin test maven configuration

nanux:
* Fix the `sourceDirs` for kotlin tests - this is then propagated to child modules
* Removed `.java` qualifier for tests
* Tested also `-Pintegration` profile - even though interestingly we don't have an integration test in `api` package (I had created dummy one for the test)
* Added kotlin incremental compilation - https://kotlinlang.org/docs/reference/using-maven.html#incremental-compilation

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -252,8 +252,10 @@
                             <goal>compile</goal>
                         </goals>
                         <configuration>
-                            <sourceDirs>${project.basedir}/src/main/java</sourceDirs>
-                            <sourceDirs>${project.basedir}/src/main/kotlin</sourceDirs>
+                            <sourceDirs>
+                                <sourceDir>${project.basedir}/src/main/java</sourceDir>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -263,8 +263,8 @@
                             <goal>test-compile</goal>
                         </goals>
                         <configuration>
-                            <sourceDirs>${project.basedir}/src/main/java</sourceDirs>
-                            <sourceDirs>${project.basedir}/src/main/kotlin</sourceDirs>
+                            <sourceDirs>${project.basedir}/src/test/java</sourceDirs>
```
theghostwhoforks, line 266: Whoa - how does it allow 2 `sourceDirs` definition here? would it accept a regex like syntax `(java|kotlin)` instead?

nanux: Interestingly `maven` didn't complain and actually pick it up correctly but it is not the correct format. I've updated it to make it consistent.

PR #108, opened 2020-04-06
CHET-117: migration stack state machine

bpar476:
# Summary
* Deploy migration stack after quick start
* Read application stack to create migration stack inputs
* Call migration service transitions and assertions as part of migration stack deployment

## Draft

This is a draft until #100 is merged


core/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
```diff
@@ -132,6 +191,7 @@ public StackStatus getStatus(String stackName) {
             Stack stack = response.stacks().get(0);
             return Optional.ofNullable(stack);
         } catch (CompletionException | CancellationException e) {
+            logger.error("Error getting stack {}", stackName, e);
             return Optional.empty();
```
tarka, line 195: Not part of this PR, but shouldn't this return an error with a message so that the UI can present the information?

bpar476: We have a ticket open to address this method. I think because we're using exceptions in most places for error stages, we should just throw an exception. https://aws-partner.atlassian.net/browse/CHET-219

I didn't want to tackle it here because it's kind of an extra thing. I agree though it's not very good traceability into what's going on.

PR #109, opened 2020-04-06
CHET-206: add api dependency to plugin

theghostwhoforks:
add api dependency to plugin

### DIFF
--- a/jira-plugin/pom.xml
+++ b/jira-plugin/pom.xml
@@ -23,6 +23,11 @@
         </dependency>
         <dependency>
             <groupId>com.atlassian.migration.datacenter</groupId>
+            <artifactId>api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
+        <dependency>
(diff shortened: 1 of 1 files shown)


theghostwhoforks:
`SPI` is not added as a dependency as `core` and `api` depends on the package

Verified: The obr contains the `spi` package

jira-plugin/pom.xml
```diff
@@ -21,6 +21,11 @@
             <artifactId>core</artifactId>
             <version>${project.version}</version>
         </dependency>
+        <dependency>
+            <groupId>com.atlassian.migration.datacenter</groupId>
+            <artifactId>api</artifactId>
+            <version>${project.version}</version>
+        </dependency>
```
tarka, line 28: SPI should probably be added too.

tarka: Actually, never mind, it's pulled in by core.

PR #111, opened 2020-04-06
CHET-183: Update dogfooding on each merge to master

nanux:
I've cleaned up the history of this branch, it was a lot of mess with Github Actions trial and error.

* We need to upload an artifact to share it across jobs
* It needs to be uploaded only from one step from the strategy matrix. I've picked unit tests on Java 11
* The variables are managed from Github secrets
* So far we are doing just a basic check whether the plugin was installed, more extensive tests are coming.
* Branch is now configurable via DOGFOODING_REF github secrets (`refs/heads/master` by default)

### DIFF
--- a/.github/workflows/mvn_test.yml
+++ b/.github/workflows/mvn_test.yml
@@ -43,3 +43,38 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: Test with maven
         run: mvn -B clean verify -P ${{ matrix.profile}}
+      - name: Upload
+        if: github.ref == 'refs/heads/CHET-183-update-dogfooding'
+        uses: actions/upload-artifact@v1
+        with:
+          name: plugin-jar
+          path: jira-plugin/target/jira-plugin-1.0.0.jar
+
+
+  deploy:
+    name: Deploy to dogfooding instance
+    if: github.ref == 'refs/heads/CHET-183-update-dogfooding'
+    runs-on: ubuntu-latest
+    needs: run_tests
+
+    steps:
+      - name: Download plugin jar file from the previous job
+        uses: actions/download-artifact@v1
+        with:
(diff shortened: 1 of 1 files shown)


bpar476:
Looks great! is the 'dogfooding ref' part of the configurable branch strategy?

nanux:
> Looks great! is the 'dogfooding ref' part of the configurable branch strategy?

Not sure what you mean by configurable branch strategy. Branch is now configurable via `DOGFOODING_REF` github secrets variable (refs/heads/master by default)

.github/workflows/mvn_test.yml
```diff
@@ -43,3 +46,39 @@ jobs:
           java-version: ${{ matrix.java }}
       - name: Test with maven
         run: mvn -B clean verify -P ${{ matrix.profile}}
+      - name: Upload
+        if: github.ref == env.DOGFOODING_REF && matrix.profile == '!integration' && matrix.java == 11
```
theghostwhoforks, line 50: Is this to ensure that the artifact is just loaded once?

.github/workflows/mvn_test.yml
```diff
@@ -2,6 +2,9 @@ name: Tests
 
 on: [push]
 
+env:
+  DOGFOODING_REF: ${{ secrets.DOGFOODING_REF }}
```
theghostwhoforks, line 6: is this usually `master` correct? Why is it a secret?

theghostwhoforks:
ah - already merged 👍

.github/workflows/mvn_test.yml
nanux: It is useful for testing - for example, this branch before merging but potentially in future we can test integration branch or stable tag without changing the code.

PR #112, opened 2020-04-07
CHET-217: add ability to pass in AZ list instead of string

theghostwhoforks:


theghostwhoforks:
The alternative to this is to use a custom deserializer - I have this implemented locally, but is a bit of an overkill for just one field, I reckon? If this solution isn't great, I can bring that in

Thoughts?

bpar476:
What's the reasoning for accepting it as an array in the first place instead of changing the UI to send a string and maintainging the same contract as cloudformation?

theghostwhoforks:
> What's the reasoning for accepting it as an array in the first place instead of changing the UI to send a string and maintainging the same contract as cloudformation?

The UI can send in a valid JSON object (i.e. an array of AZs, which is what a multi-option type would send anyway) without needing to handle the input field any differently. The backend converts it to what AWS expects, which is a comma separated list

bpar476:
> > What's the reasoning for accepting it as an array in the first place instead of changing the UI to send a string and maintainging the same contract as cloudformation?
> 
> The UI can send in a valid JSON object (i.e. an array of AZs, which is what a multi-option type would send anyway) without needing to handle the input field any differently. The backend converts it to what AWS expects, which is a comma separated list

I mean why is this an advantage instead of delegating to the client to send the comma separated list.

theghostwhoforks:
> > > What's the reasoning for accepting it as an array in the first place instead of changing the UI to send a string and maintainging the same contract as cloudformation?
> > 
> > 
> > The UI can send in a valid JSON object (i.e. an array of AZs, which is what a multi-option type would send anyway) without needing to handle the input field any differently. The backend converts it to what AWS expects, which is a comma separated list
> 
> I mean why is this an advantage instead of delegating to the client to send the comma separated list.

I found it advantageous to just send the AZs as a well formed json object, instead of handling this case. i,e in postman/curl when the. API had to be tested, I just found this to be more natural.

When you raised the bug, did you mean to fix this in the front end side (where we currently send a JSON string-ified array)?

bpar476:
I would have just done it in the UI but your logic is sound too. We don't have any other list parameters so I don't see any risk

PR #113, opened 2020-04-07
CHET-184: Verify REST endpoints

nanux:
Can be run with `mvn verify -pl func-tests -Prest` (REST tests are not running by default.

* Introduces `/develop/migration/reset` REST endpoint https://github.com/atlassian-labs/dc-migration-assistant/pull/117
* Add 2 `GET` REST tests with Rest Assured framework
* Add the func test stage to Github Actions.

Right now we have just 2 ReadOnly endpoint tests and we should expand in future but for testing our dogfooding instance whether the plugin is up, it is functional.

Apologies for the messy Github timeline - I was cleaning up commit history with force pushes.


func-tests/pom.xml
```diff
@@ -0,0 +1,95 @@
...
+        <artifactId>parent</artifactId>
+        <groupId>com.atlassian.migration.datacenter</groupId>
+        <version>1.0.0</version>
+    </parent>
+    <modelVersion>4.0.0</modelVersion>
+
+    <artifactId>func-tests</artifactId>
+
+    <build>
+        <sourceDirectory>${project.basedir}/src/main/kotlin</sourceDirectory>
```
theghostwhoforks, line 31: src/main/kotlin does not exist. Perhaps we don't need this to be passed in?

func-tests/src/test/kotlin/com/atlassian/migration/api/aws/AWSGlobalInfrastructureEndpointRestTest.kt
```diff
@@ -0,0 +1,49 @@
...
+    init {
+        RestAssured.authentication = basic(username, password)
+        RestAssured.baseURI = System.getProperty("JIRA_BASE_URL", "http://localhost:2990/jira")
+        RestAssured.basePath = System.getProperty("JIRA_BASE_PATH", "/rest/dc-migration/1.0")
+    }
+
+    @Test
+    fun `GET AWS Regions should respond 200 HTTP`() {
+        Given {
+            log().all()
```
theghostwhoforks, line 41: Just wanted to share - we can use specifications in restassured to create request/response specifications across tests like this.


```
     val requestSpec = RequestSpecBuilder()
                .log(LogDetail.ALL)
                .addParam("os_authType", "basic")
                .build()
    }

    @Test
    fun `GET AWS Regions should respond 200 HTTP`() {
        Given {
            spec(requestSpec)
        } When {
            get("/aws/global-infrastructure/regions")
        } Then {
            statusCode(200)
        }
    }
```
Found this to be pretty cool as the test codebase grows

func-tests/pom.xml
nanux: I've added `.gitkeep` so the structure is there if we start adding some supporting classes for the functional tests (e.g. page objects), so people are guided that they should be `kotlin` as well.

func-tests/src/test/kotlin/com/atlassian/migration/api/aws/AWSGlobalInfrastructureEndpointRestTest.kt
nanux: Awesome, changed the code to use `RequestSpecBuilder`.

nanux:
The builds are passing now, @theghostwhoforks can you review if we can resolve your comment?

PR #117, opened 2020-04-08
CHET-228: Reset REST endpoint

nanux:


nanux:
I will just merge this to the CHET-184-verify-endpoints branch as I need to work with it there and we can review it in the same PR.

PR #119, opened 2020-04-09
Chet 229 authentication page

theghostwhoforks:
This is what the error page looks like on the dev server

Screenshots -

What a user sees, when she lands on the AWS auth page
Screenshot 2020-04-16 at 12 04 32

Error flag, when credentials API returns a 4xx or 5xx
Screenshot 2020-04-16 at 12 56 50

Button loading, while awaiting response from API
Screenshot 2020-04-16 at 12 04 46

On success, a user is routed to Quickstart deploy page -
Screenshot 2020-04-16 at 12 58 02


frontend/src/main/dc-migration-assistant-fe/components/aws/auth/AuthenticateAWS.tsx
```diff
@@ -49,6 +49,11 @@ const CredsSubmitButton = styled(Button)`
     margin-top: 10px;
 `;
 
+const CancelMigrationButton = styled(Button)`
```
bpar476, line 52: I had to downgrade the version of styled components as Atlaskit has a peer dependency on it :( We can't use the `styled()` function on existing components.

frontend/src/main/resources/i18n/dc-migration-assistant.properties
```diff
@@ -30,13 +30,16 @@ atlassian.migration.datacenter.home.synopsis=Use this assistant to securely migr
 atlassian.migration.datacenter.home.explore.migration=Explore Migration
 
 atlassian.migration.datacenter.overview.title=Migration Overview
+atlassian.migration.datacenter.step.phrase=Step 1 of 6: Connect to AWS
```
bpar476, line 33: Should this string key be related to the authenticate stage? e.g. `atlassian.migration.datacenter.step.authenticate.phrase`

frontend/src/main/dc-migration-assistant-fe/components/aws/auth/AuthenticateAWS.tsx
theghostwhoforks: So, do we need to change this to something else? What alternatives are we uing?

bpar476: Is this compiling on your end if you do a `yarn`?

bpar476: I would try this approach https://styled-components.com/docs/advanced#referring-to-other-components

bpar476: Alernatively I would try to use a flexbox container instead of manually setting margins

bpar476: Also the `Button` has a `style` prop which you can use which may just be the cleanest way

theghostwhoforks: Yeah - It's compiling. 

I've been using the dev server without any issues.

The `Button` with the `CSSProperties` that you've used in the other PR may be the right way to go ahead with this

bpar476: I would do a `rm -rf node_modules` and then `yarn` as it shouldn't compile when using the correct styledcomponents version 🤔

theghostwhoforks:
For unknown reasons, the catch block is being called even when the response code is 204. Needs to be investigated

theghostwhoforks:
> For unknown reasons, the catch block is being called even when the response code is 204. Needs to be investigated

So, fetch, when it encounters a 204, errors out as it cannot parse an empty body 🤦 

https://github.com/whatwg/fetch/issues/113

frontend/src/main/dc-migration-assistant-fe/components/aws/auth/AuthenticateAWS.tsx
theghostwhoforks: yeah - This is the only thing pending on the PR now. Will update on screenshots and  move away from `styled()` invocations

theghostwhoforks: this has been resolved now

api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/AWSConfigureEndpoint.kt
```diff
@@ -31,21 +32,26 @@ class AWSConfigureEndpoint(private val awsConfigurationService: AWSConfiguration
...
-                .build()
+            Response.noContent().build()
+        } catch (e: InvalidMigrationStageError) {
+            Response
+                    .status(Response.Status.CONFLICT)
+                    .entity(mapOf("message" to e.message))
+                    .build()
+        } catch (e: InvalidCredentialsException) {
+            Response
+                    .status(Response.Status.BAD_REQUEST)
```
theghostwhoforks, line 51: changed this to a bad request instead of a conflict

core/src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/DefaultAwsCloudCredentialsValidator.java
```diff
@@ -0,0 +1,49 @@
+package com.atlassian.migration.datacenter.core.aws.cloud;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.auth.credentials.AwsBasicCredentials;
+import software.amazon.awssdk.services.sts.StsClient;
+import software.amazon.awssdk.services.sts.model.GetCallerIdentityResponse;
+import software.amazon.awssdk.utils.Md5Utils;
+
+public class DefaultAwsCloudCredentialsValidator implements CloudCredentialsValidator {
```
theghostwhoforks, line 10: Reviewed as part of a previous PR

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/__snapshots__/quickstartToAtlaskit.test.tsx.snap
```diff
@@ -33,7 +33,7 @@ exports[`Quick Start to Atlaskit Input Converter Should create a select input wh
   autocorrect="off"
   id="react-select-2-input"
   spellcheck="false"
-  style="box-sizing: content-box; width: 2px; border: 0px; font-size: inherit; opacity: 1; outline: 0; padding: 0px;"
+  style="box-sizing: content-box; width: 2px; border: 0px; opacity: 1; outline: 0; padding: 0px;"
```
theghostwhoforks, line 36: don't know why this changed - perhaps I ran an updateSnapshots sometime?

frontend/src/main/dc-migration-assistant-fe/components/db/DatabaseMigration.tsx
```diff
@@ -17,8 +17,8 @@
 import React, { FunctionComponent } from 'react';
 
 import { I18n } from '@atlassian/wrm-react-i18n';
-import { MigrationTransferProps, MigrationTransferPage } from '../shared/MigrationTransferPage';
 import moment from 'moment';
+import { MigrationTransferProps, MigrationTransferPage } from '../shared/MigrationTransferPage';
```
theghostwhoforks, line 21: `yarn lint --fix` did this

PR #122, opened 2020-04-14
CHET-229: Add a credentials validator to check if Auth API is called …

theghostwhoforks:
Invalid request -
Screenshot 2020-04-14 at 15 06 30

Valid request -
Screenshot 2020-04-14 at 15 06 56

…with correct credentials

### DIFF
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/AWSConfigureEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/AWSConfigureEndpoint.kt
@@ -16,6 +16,7 @@
 package com.atlassian.migration.datacenter.api.aws
 
 import com.atlassian.migration.datacenter.core.aws.cloud.AWSConfigurationService
+import com.atlassian.migration.datacenter.spi.exceptions.InvalidCredentialsException
 import com.atlassian.migration.datacenter.spi.exceptions.InvalidMigrationStageError
 import com.fasterxml.jackson.annotation.JsonAutoDetect
 import javax.ws.rs.Consumes
@@ -31,21 +32,26 @@ class AWSConfigureEndpoint(private val awsConfigurationService: AWSConfiguration
     @Consumes(MediaType.APPLICATION_JSON)
     @Produces(MediaType.APPLICATION_JSON)
     fun storeAWSCredentials(configure: AWSConfigureWebObject?): Response {
-        try {
+        return try {
             configure?.let {
                 awsConfigurationService.configureCloudProvider(
-                    configure.accessKeyId,
-                    configure.secretAccessKey,
-                    configure.region
+                        configure.accessKeyId,
+                        configure.secretAccessKey,
+                        configure.region
                 )
             }
-        } catch (invalidMigrationStageError: InvalidMigrationStageError) {
-            return Response
-                .status(Response.Status.CONFLICT)
-                .entity(invalidMigrationStageError.message)
-                .build()
+            Response.noContent().build()
+        } catch (e: InvalidMigrationStageError) {
+            Response
+                    .status(Response.Status.CONFLICT)
+                    .entity(e.message)
+                    .build()
+        } catch (e: InvalidCredentialsException) {
+            Response
+                    .status(Response.Status.CONFLICT)
+                    .entity(e.message)
+                    .build()
         }
-        return Response.noContent().build()
     }
 
     @JsonAutoDetect
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/DefaultAwsCloudCredentialsValidator.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/DefaultAwsCloudCredentialsValidator.java
@@ -41,7 +41,7 @@ public class DefaultAwsCloudCredentialsValidator implements CloudCredentialsVali
         try {
             String userId = callerIdentity.userId();
             String userIdBase64 = Md5Utils.md5AsBase64(userId.getBytes());
-            logger.debug("Credentials validation completed successfully for user - {}", userIdBase64);
+            logger.debug("Credentials validation completed successfully for user (MD5, base64 encoded) - {}", userIdBase64);
         } catch (Exception e) {
             logger.debug("Error while computing md5 of user ID. Ignoring this error", e);
         }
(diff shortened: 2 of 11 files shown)


theghostwhoforks:
Will update PR with postman screenshots

core/src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/DefaultAwsCloudCredentialsValidator.java
```diff
@@ -0,0 +1,29 @@
...
+     * @param accessKeyId an identifier for the entity accessing the cloud provider e.g. AWS access key ID
+     * @param secretAccessKey the secret to authenticate the entity to the cloud provider e.g. AWS secret access key
+     * @return true when supplied keys are valid. false when the any one of the keys, or both, are not valid.
+     */
+//    TODO: Evaluate if we can use the simulate policy API ( This does not validate IAM policy as defined here {@link https://docs.aws.amazon.com/IAM/latest/APIReference/API_SimulatePrincipalPolicy.html} as invoking that). Invoking this API requires passing in a required `policySourceARN` that points to the user/group/role for which the simulation has to run for. We do not accept this as an input from the user yet.
+    @Override
+    public Boolean validate(String accessKeyId, String secretAccessKey) {
+        try {
+            AwsBasicCredentials awsBasicCredentials = AwsBasicCredentials.create(accessKeyId, secretAccessKey);
+            CloudFormationAsyncClient.builder().credentialsProvider(() -> awsBasicCredentials).build();
```
bpar476, line 21: Does this actually fail if the credentials are invalid? I thought you only get denied when you actually make a request.

theghostwhoforks: testing this out. It looks like it, will update upon verification

theghostwhoforks: Using this now - https://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html

api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/AWSConfigureEndpoint.kt
```diff
@@ -31,21 +32,26 @@ class AWSConfigureEndpoint(private val awsConfigurationService: AWSConfiguration
...
+                        configure.secretAccessKey,
+                        configure.region
                 )
             }
-        } catch (invalidMigrationStageError: InvalidMigrationStageError) {
-            return Response
-                .status(Response.Status.CONFLICT)
-                .entity(invalidMigrationStageError.message)
-                .build()
+            Response.noContent().build()
```
nanux, line 43: If `configure` is null, does it mean Jackson wasn't able to parse the data? Should we return `400` code? Or maybe even better https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422

api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/AWSConfigureEndpoint.kt
```diff
@@ -31,21 +32,26 @@ class AWSConfigureEndpoint(private val awsConfigurationService: AWSConfiguration
...
+        } catch (e: InvalidMigrationStageError) {
+            Response
+                    .status(Response.Status.CONFLICT)
+                    .entity(e.message)
+                    .build()
+        } catch (e: InvalidCredentialsException) {
+            Response
+                    .status(Response.Status.CONFLICT)
```
nanux, line 51: I would say this is probably a `400` HTTP status.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/cloud/DefaultAwsCloudCredentialsValidator.java
```diff
@@ -0,0 +1,49 @@
...
+     * @param accessKeyId     an identifier for the entity accessing the cloud provider e.g. AWS access key ID
+     * @param secretAccessKey the secret to authenticate the entity to the cloud provider e.g. AWS secret access key
+     * @return true when supplied keys are valid. false when the any one of the keys, or both, are not valid.
+     */
+//    TODO: Evaluate if we can use the simulate policy API ( This does not validate IAM policy as defined here {@link https://docs.aws.amazon.com/IAM/latest/APIReference/API_SimulatePrincipalPolicy.html} as invoking that). Invoking this API requires passing in a required `policySourceARN` that points to the user/group/role for which the simulation has to run for. We do not accept this as an input from the user yet.
+//    TODO: getCallerIdentity may not work for users using MFA and/or when the role does not have privileges to call getUser. Verify this
+    @Override
+    public Boolean validate(String accessKeyId, String secretAccessKey) {
+        try {
+            AwsBasicCredentials awsBasicCredentials = AwsBasicCredentials.create(accessKeyId, secretAccessKey);
```
nanux, line 25: Note, this validates authentication but doesn't validate authorisation for services that we need for the migration and new stack. I am not sure if we can make easy validation for permissions. I had a quick look into `iam` client but not sure if you can get an effective list of applicable policies. Ideally, we would get a list of permissions that the user has available and validate it against the our list of expected permissions.

Dry runs are unfortunately not supported in sdk v2
https://github.com/aws/aws-sdk-java-v2/issues/76

theghostwhoforks: Thank you for taking the time to review the AWS client documentation in detail. Much appreciated!

I went around in circles on the dry run support - The API docs suggest that it's available only to realize that it isn't there in the v2 SDK.

I'm not too happy with the current implementation. I looked at [simulate principal policy API](https://docs.aws.amazon.com/IAM/latest/APIReference/API_SimulatePrincipalPolicy.html), but I'm not convinced that this will work? (if someone has a very specific handwritten policy as opposed to using a AWS managed policy - we can't expect a `cloudformation::*` when admins can give a role access to `cloudformation::Create*` and not give access to `cloudformation::List*` - which will technically work).

I'd be happy to hear your recommendations on this! 

For now, if we allow admins to retry a step in the migration plan, we will be able to allow them to recover by giving access to resources when they notice that a step has failed.

PR #123, opened 2020-04-15
CHET-242: Correct "migrated files" count being inflated

bpar476:
The queue was not being emptied after "acknowledging" all the responses so we were repeating ourselves many times.


theghostwhoforks:
Was the reason for the inflated count because we weren't calling polling/removing items from the queue  every time we handled the response from S3?

nanux:
> Was the reason for the inflated count because we weren't calling polling/removing items from the queue every time we handled the response from S3?

That is correct, we never polled from the queue.

PR #126, opened 2020-04-15
CHET-236: expose fs migration meta status

bpar476:
* When uploading to S3 set status of FS migration to "UPLOADING"
* When downloading on remote host, set status to "DOWNLOADING".
* This allows us to expose more state in the UI.


core/src/main/java/com/atlassian/migration/datacenter/core/fs/download/s3sync/S3SyncFileSystemDownloadManager.java
```diff
@@ -55,7 +55,7 @@ public void downloadFileSystem(FileSystemMigrationProgress progress) throws S3Sy
                 logger.debug("file system download is complete");
                 syncCompleteFuture.complete(null);
             }
-        }, 0, 5, TimeUnit.MINUTES);
+        }, 0, 10, TimeUnit.SECONDS);
```
nanux, line 58: note: this was changed after discussion that 5mins report on filesDownloaded (that is computed in this block) wouldn't be enough to give meaningful feedback for frontend.

PR #127, opened 2020-04-15
CHET-244: expose when fs crawl is complete

bpar476:
We expose when FS crawl is complete because at that point you can determine the exact progress of the migration by checking the `filesFound` of the fs migration status.

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
@@ -133,6 +133,16 @@ public class DefaultFileSystemMigrationReport implements FileSystemMigrationRepo
     }
 
     @Override
+    public boolean allFilesFound() {
+        return progress.allFilesFound();
+    }
+
+    @Override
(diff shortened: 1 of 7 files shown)


core/src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
```diff
@@ -132,6 +132,16 @@ public void reportFileFound() {
         progress.reportFileFound();
     }
 
+    @Override
+    public boolean allFilesFound() {
```
nanux, line 136: I would call it `isCrawlingFinished` just because I can immeditally see from the method name that it is a boolean.

core/src/main/java/com/atlassian/migration/datacenter/core/fs/reporting/DefaultFileSystemMigrationReport.java
```diff
@@ -132,6 +132,16 @@ public void reportFileFound() {
...
+        return progress.allFilesFound();
+    }
+
+    @Override
+    public void reportAllFilesFound() {
```
nanux, line 141: `reportCrawlingFinished`

PR #130, opened 2020-04-19
CHET-245: Remove FS progress strings; increase poll rate of FS progress

bpar476:
This is after getting feedack from Hanna. Having the file progress going from a large number to zero then counting up again is confusing. We will just use the progress bar.

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/fs/FileSystemTransfer.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/fs/FileSystemTransfer.tsx
@@ -37,7 +37,7 @@ const getFsMigrationProgress = (): Promise<Progress> => {
             if (result.status === 'UPLOADING') {
                 const progress: Progress = {
                     phase: 'Uploading files to AWS',
-                    progress: `${result.uploadedFiles} files uploaded`,
+                    progress: '',
                 };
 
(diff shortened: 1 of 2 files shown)


frontend/src/main/dc-migration-assistant-fe/components/fs/FileSystemTransfer.tsx
```diff
@@ -69,13 +69,13 @@ const getFsMigrationProgress = (): Promise<Progress> => {
...
                 return {
                     phase: 'Preparing to migrate files',
-                    progress: '...',
+                    progress: '',
                 };
             }
             return {
                 phase: 'error',
                 completeness: 0,
-                progress: 'error',
```
nanux, line 78: How we are going to surface the error?

bpar476:
> What if that would be on two lines? Like
> 
> Step 1: 23,329 files uploaded (COMPLETE)
> Step 2: 13,392 files downloaded (57%)

I'd run that past Hanna :)

frontend/src/main/dc-migration-assistant-fe/components/fs/FileSystemTransfer.tsx
bpar476: We probably need to do more non-happy path development but probably show an error section message.

PR #136, opened 2020-04-21
CHET-248: "Start" fs migration experience

bpar476:
# Summary

* Check if the fs migration has started
* Initiate fs migration
* Handle initiationg of migration failing.

image


bpar476:
I need to fix the "start" string and reduce the margin between those buttons.

PR #139, opened 2020-04-22
CHET-256: surface fs page errors

bpar476:
# Summary

* Get errors from `failedFiles` of result
* Show number of failed files
* Show all unique reasons for file failure
* Render in error section message

image


bpar476:
Something seems to have gone very wrong with my branching and this includes the "start fs migration" code as well 🤦

PR #140, opened 2020-04-22
CHET-251: Add DB screen progress callbacks to API

tarka:


bpar476:
This may be broken by the latest updates to `MigrationTransfer.tsx` as I think it it will require a function to start it. However I reckon just put no-ops in the missing bits for now, track them, and fill them out as a follow up.

PR #142, opened 2020-04-23
Chet 271 validation page

theghostwhoforks:
Mock Screenshot of validation

Screenshot 2020-04-28 at 10 08 00

(Due to stub data, we don't see a instance URL here.)

There will be a quick follow task after this PR to add a small validation to ensure that the stage of the migration state machine is in `VALIDATION` stage

On button click, a user is redirected to home


api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
```diff
@@ -77,4 +77,28 @@ class MigrationEndpoint(private val migrationService: MigrationService) {
                 .build()
         }
     }
+
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    @GET
+    @Path("/context")
+    fun getMigrationContext(): Response {
```
bpar476, line 85: Should we call this something different e.g. `/summary` as we're not exposing the entire migration context to the user (and we probably shouldn't)

frontend/src/main/dc-migration-assistant-fe/stage/Validation.tsx
```diff
@@ -0,0 +1,157 @@
...
+const SectionMessageContainer = styled.div`
+    margin-top: 15px;
+    width: 100%;
+`;
+
+const MigrationSummaryActionCallout = (): ReactElement => {
+    return (
+        <MigrationSummaryContainer>
+            <div>
+                <h4>Actions required on your side</h4>
```
bpar476, line 52: /nit Internationalisation

frontend/src/main/dc-migration-assistant-fe/stage/Validation.tsx
```diff
@@ -0,0 +1,157 @@
...
+            </div>
+            <div>
+                <ul>
+                    <li>
+                        {I18n.getText('atlassian.migration.datacenter.validation.post.action.1')}
```
bpar476, line 57: I think the i18n keys should be a bit more meaningful than `action.1`, `action.2` etc. It's hard to tell what their purpose is otherwise.

api/pom.xml
```diff
@@ -23,6 +23,11 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
```
bpar476, line 26: Is this actually needed in the `api` module? I don't see any references?

frontend/src/main/dc-migration-assistant-fe/stage/Validation.tsx
```diff
@@ -0,0 +1,157 @@
...
+                    .filter(x => x.key === INSTANCE_URL_MIGRATION_CONTEXT_KEY)
+                    .map(summary => {
+                        return (
+                            <Row key={`migration-summary-${summary.key}`} hasChildren={false}>
+                                <Cell width={400} singleLine>
+                                    {I18n.getText(
+                                        'atlassian.migration.datacenter.validation.summary.phrase.instanceUrl'
+                                    )}
+                                </Cell>
+                                <Cell width={400}>{summary.value}</Cell>
```
bpar476, line 118: Should this have a `Link` component?

api/pom.xml
theghostwhoforks: <I thought I wrote why this was needed in the PR, but I didn't 🤦 >

So, the reason this is required is because of 

> See questions. I think the service URL needs to be wrapped in a `<Link>`

Yes! Will do

theghostwhoforks: I'm testing this outside of idea to understand if this was an idea issue or something else

theghostwhoforks: Compilation fails without AO from CLI as well - kotlin needs a ref to AO `entity` when we invoke fields on the `migrationContext`

bpar476: Oh I see, it can probably be `provided` scope right?

theghostwhoforks:
> See questions. I think the service URL needs to be wrapped in a `<Link>`

All PR feedback has been resolved now

frontend/src/main/dc-migration-assistant-fe/stage/Validation.tsx
```diff
@@ -115,7 +125,9 @@ const MigrationSummary: FunctionComponent = () => {
                                         'atlassian.migration.datacenter.validation.summary.phrase.instanceUrl'
                                     )}
                                 </Cell>
-                                <Cell width={400}>{summary.value}</Cell>
+                                <Cell width={400}>
+                                    <Link to={summary.value} />
```
bpar476, line 129: Does this still render the URL? I thought you need to enclose it like this `<Link to={url}>{url}</Link`

theghostwhoforks: Silly me - I'm rebasing with master. Will update that soon

api/pom.xml
theghostwhoforks: Yeah - Will call it out explicitly

PR #145, opened 2020-04-26
CHET-263: Fix DB upload + some other bugs

nanux:
* Fix `S3AsyncClient` initialisation in `DatabaseArtifactS3UploadService` - the client was null because our `@PostConstruct` methods are not working so for now we need to init the client manually
* I've removed stored `CfnApi` client as it is storing a state that is possible to change and the client would be stale.
* Add tests to cover the case (+some more)

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
@@ -31,7 +31,6 @@ import software.amazon.awssdk.services.cloudformation.model.DescribeStackResourc
 import software.amazon.awssdk.services.cloudformation.model.DescribeStackResourcesResponse;
 import software.amazon.awssdk.services.cloudformation.model.DescribeStacksRequest;
 import software.amazon.awssdk.services.cloudformation.model.DescribeStacksResponse;
-import software.amazon.awssdk.services.cloudformation.model.ListExportsRequest;
 import software.amazon.awssdk.services.cloudformation.model.ListExportsResponse;
 import software.amazon.awssdk.services.cloudformation.model.Parameter;
 import software.amazon.awssdk.services.cloudformation.model.Stack;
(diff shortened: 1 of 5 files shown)


core/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
```diff
@@ -75,20 +78,15 @@ public CfnApi(AwsCredentialsProvider credentialsProvider, RegionService regionMa
...
 
     /**
-     * Lazily create a CFN client; should only be called after necessary AWS information has been provided.
+     * Return a client should only be called after necessary AWS information has been provided.
      */
     private CloudFormationAsyncClient getClient() {
-        if (client.isPresent()) {
-            return client.get();
-        }
+        return client.orElseGet(
```
nanux, line 84: This is a fix for CHET-260.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseArtifactS3UploadService.java
```diff
@@ -42,6 +42,7 @@ public void postConstruct() {
     }
 
     public FileSystemMigrationReport upload(Path target, String targetBucketName, DatabaseUploadStageTransitionCallback callback) throws InvalidMigrationStageError, FilesystemUploader.FileUploadException {
+        s3AsyncClient = s3AsyncClientSupplier.get();
```
nanux, line 45: This is the fix for the DB upload problem.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -112,7 +108,7 @@ public FileSystemMigrationErrorReport performMigration() throws DatabaseMigratio
             migrationService.error(e);
             throw new DatabaseMigrationFailure("Error when restoring database", e);
         }
-        migrationService.transition(MigrationStage.DB_MIGRATION_UPLOAD_WAIT);
+        migrationService.transition(MigrationStage.VALIDATE);
```
nanux, line 111: Unrelated fix but we were ending up in incorrect stage.

PR #147, opened 2020-04-27
CHET-283: Provide region when validating AWS creds.

tarka:


theghostwhoforks:
Also, `com.atlassian.migration.datacenter.core.aws.region.PluginSettingsRegionManager#isValidRegion` validates the region before storing it. 

We check this in `AWSConfigurationService` before persisting region and other credentials.
So, this may not be needed? What am I missing here?

bpar476:
> Also, `com.atlassian.migration.datacenter.core.aws.region.PluginSettingsRegionManager#isValidRegion` validates the region before storing it.
> 
> We check this in `AWSConfigurationService` before persisting region and other credentials.
> So, this may not be needed? What am I missing here?

The issue is the STS client may not be authenticated to the correct region, although the IAM user should exist globally, so maybe this isn't needed 🤔

bpar476:
@tarka Is this maybe a red herring and the credentials you were using were invalid?

nanux:
> @tarka Is this maybe a red herring and the credentials you were using were invalid?

I think it definitely looks suspicious. From the documentation - https://docs.aws.amazon.com/STS/latest/APIReference/Welcome.html :

> By default, AWS Security Token Service (STS) is available as a global service, and all AWS STS requests go to a single endpoint at https://sts.amazonaws.com. Global requests map to the US East (N. Virginia) region. 

_Therefore I believe the problem wasn't the missing region._

In the docs there is also this:

> AWS recommends using Regional AWS STS endpoints instead of the global endpoint to reduce latency, build in redundancy, and increase session token validity.

So there is a case to make about using the region with the client but I think our one-time-use case doesn't really require it.

tarka:
> So there is a case to make about using the region with the client but I think our one-time-use case doesn't really require it.

I've tried the valid credentials without the region, and it's doesn't work (for me).

PR #148, opened 2020-04-27
CHET-284: start migration

bpar476:
## Summary

* Call API to create migration when "start" is clicked
* Render error flag when creation fails

Note: Does not redirect to next page as that is being left to CHET-278 so that it can be done in a consistent manner.

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/shared/ErrorFlag.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/shared/ErrorFlag.tsx
@@ -1,0 +1,66 @@
+import styled, { keyframes } from 'styled-components';
+import React, { FunctionComponent, ReactElement } from 'react';
+import Flag from '@atlaskit/flag';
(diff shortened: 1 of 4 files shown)


frontend/src/main/dc-migration-assistant-fe/components/shared/ErrorFlag.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+const ErrorContainer = styled.div`
+    position: fixed;
+    top: 70px;
+    left: 75%;
+    right: 1%;
+    overflow: inherit;
+`;
+
+export const ErrorFlag: FunctionComponent<AuthenticationErrorProps> = ({
```
theghostwhoforks, line 39: Nice. Reusable

bpar476: You built it, I just moved it ;)

PR #150, opened 2020-04-28
CHET-272: Add a error flag to highlight out of sequence migration step

theghostwhoforks:
Add a validation error flag component.

Refactor to use the migration api

2 stages of the component -

* When stage is validation
 
Screenshot 2020-04-28 at 14 43 26

* When stage is not validation

Screenshot 2020-04-28 at 14 43 47

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/stage/Validation.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/stage/Validation.tsx
@@ -21,8 +21,9 @@ import TableTree, { Cell, Row } from '@atlaskit/table-tree';
 import { Button } from '@atlaskit/button/dist/esm/components/Button';
 import styled from 'styled-components';
 import { Link, useHistory } from 'react-router-dom';
-import { callAppRest, RestApiPathConstants } from '../utils/api';
 import { homePath } from '../utils/RoutePaths';
+import { migration, MigrationStage } from '../api/migration';
+import { ErrorFlag } from '../components/shared/ErrorFlag';
 
 const MigrationSummaryContainer = styled.div`
     display: grid;
@@ -94,8 +95,8 @@ const MigrationSummary: FunctionComponent = () => {
     >([]);
 
     useEffect(() => {
-        callAppRest('GET', RestApiPathConstants.migrationSummaryRestPath)
-            .then(response => response.json())
+        migration
+            .getMigrationSummary()
             .then(data => {
                 setSummaryData(
                     Object.entries(data).map(summary => {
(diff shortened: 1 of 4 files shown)


bpar476:
Is this meant to be for CHET-272? I thought we were going to do a redirect 🤔

theghostwhoforks:
> Is this meant to be for CHET-272? I thought we were going to do a redirect 🤔

Yes, This is for CHET-272

frontend/src/main/resources/i18n/dc-migration-assistant.properties
```diff
@@ -100,6 +100,9 @@ atlassian.migration.datacenter.db.storage.size=Database storage size (GB)
 atlassian.migration.datacenter.db.storage.type=Database storge type
 
 atlassian.migration.datacenter.step.validation.phrase=Step 7 of 7: Validation
+atlassian.migration.datacenter.step.validation.incorrect.stage.error.title=Out of sequence migration step
```
theghostwhoforks: @ddomingorh - Can you please check if this text is accurate and depicts the intent for. a user clearly? 

A user will see this when they navigate to the validation page directly without completing the previous steps in the migration sequence

frontend/src/main/dc-migration-assistant-fe/stage/Validation.tsx
```diff
@@ -186,10 +160,54 @@ export const ValidateStagePage: FunctionComponent = () => {
                 style={{
                     marginTop: '15px',
                 }}
-                onClick={redirectUserToHome}
+                onClick={onClick}
```
bpar476, line 163: Why use an `onClick` and the `history` API when you could just use a `<Link>` component?

theghostwhoforks: Makes sense - Will change this to use a `href` and `Link` instead of the `onClick` Action

frontend/src/main/dc-migration-assistant-fe/stage/Validation.tsx
```diff
@@ -161,11 +157,57 @@ export const ValidateStagePage: FunctionComponent = () => {
...
             >
                 {I18n.getText('atlassian.migration.datacenter.validation.next.button')}
             </Button>
         </>
     );
 };
+
+const InvalidMigrationStageErrorMessage = (): ReactElement => (
+    <SectionMessage appearance="error">
```
theghostwhoforks, line 172: I realize now that we can pass in actions to the section message. Will use that instead of `links` in a paragraph

PR #151, opened 2020-04-28
CHET-266: Add ability to continue with a migration after started.

tarka:
This is a short-term workaround to allow us to continue with an in-progress migration.This will aid end-to-end testing.


bpar476:
I think next spring we're planning on tackling CHET-279 which kind of handles a similar case. Do we want to incorporate this PR even though it will probably be implemented diferently next week? 🤔

tarka:
> I think next spring we're planning on tackling CHET-279 which kind of handles a similar case. Do we want to incorporate this PR even though it will probably be implemented diferently next week?

As I mentioned, this is to help with testing right now. If it gets torn out that's fine.

PR #155, opened 2020-04-29
CHET-286: fix SecretIdentifier

nanux:


templates/migration-stack/migration-helper.yml
```diff
@@ -148,11 +148,11 @@ Resources:
                     DB_DUMP_LOG_FILE="/var/atlassian/dc-migration-assistant/pg_dump-log.txt"
                     SECRET_PASSWORD=`aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"`
                     if [ $? != 0]; then
-                      echo "Error when fetching db password from secret manage"
+                      echo "Error when fetching DB password from secret manager"
                       exit 1
                     fi
                     aws s3 sync s3://${MigrationBucket}/db.dump/ $DATABASE_DOWNLOAD_DIR --region ${AWS::Region} | tee $DB_DUMP_LOG_FILE
-                    if [ $? != 0]; then
+                    if [ ${PIPESTATUS[0]} != 0]; then
```
theghostwhoforks, line 155: why did  this change @nanux ?

templates/migration-stack/migration-helper.yml
```diff
@@ -170,6 +170,7 @@ Resources:
                             exit 1
                     fi
                   - {
+                    SecretIdentifier: !Sub "atl-${AWS::StackName}-app-rds-password",
```
theghostwhoforks, line 173: 👍

theghostwhoforks:
I don't see the `PIPESTATUS` code in the `.src` file in the commit - https://github.com/atlassian-labs/dc-migration-assistant/blob/f8b422b4b18fb072d7b412ffe37c8628b03bcb6f/templates/migration-stack/pkg/migration-helper.yml.src
Am I missing something? I expected to see the changes in both the files

templates/migration-stack/migration-helper.yml
nanux: The previous implementation wasn't really doing what it should. It was using the exit code of the `tee` command rather than the `s3 sync` which should be indicative whether the command failed or not.

We should fail in case `s3 sync` failed and to get the result for a command in `pipe`, we can use `${PIPESTATUS[0]}` - https://www.shellscript.sh/tips/pipestatus/ This works for us because the restore script runs in `bash`

You can test this in your shell:
```
#!/bin/bash

echo "Using pipeline"
false | tee "x"
echo ${PIPESTATUS[0]}

echo "Using \$? code"
false | tee "x"
echo $?
```

nanux:
> I don't see the `PIPESTATUS` code in the `.src` file in the commit - https://github.com/atlassian-labs/dc-migration-assistant/blob/f8b422b4b18fb072d7b412ffe37c8628b03bcb6f/templates/migration-stack/pkg/migration-helper.yml.src
> Am I missing something? I expected to see the changes in both the files

I am missing something :) Fixed.

templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -143,32 +143,32 @@ Resources:
...
-                     fi
-                     aws s3 sync s3://${MigrationBucket}/db.dump/ $DATABASE_DOWNLOAD_DIR --region ${AWS::Region} | tee $DB_DUMP_LOG_FILE
-                     if [ $? != 0]; then
-                       echo "Error when synchronising with the S3 bucket"
-                       exit 1
-                     fi
-                     echo "Restoring database from $DATABASE_DOWNLOAD_DIR to ${DBHost}:${DBPort}/$DBName" | tee $DB_DUMP_LOG_FILE
-                     PGPASSWORD=$SECRET_PASSWORD pg_restore --no-owner --no-acl --clean --if-exists -h ${DBHost} -U ${DBUser} -d ${DBName} -p ${DBPort} -F d --verbose $DATABASE_DOWNLOAD_DIR 2>&1 | tee $DB_DUMP_LOG_FILE
-                     ERROR_OUTPUT=$(grep -iE 'error|warning' $DB_DUMP_LOG_FILE)
+                    DATABASE_DOWNLOAD_DIR="/efs/downloads/db.dump"
```
nanux, line 146: This is just aligning the indentation with source.

test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+echo "Using pipeline"
+false | tee "x"
+echo ${PIPESTATUS[0]}
```
bpar476, line 5: Is this meant to be commited?

nanux: 😥 got new `git` aliases and forgot it will add all files in the folder.

PR #158, opened 2020-04-29
fix the REST test running against dogfooding

nanux:
This is a fix for now but it is not ideal.

Testing against dogfooding instance that is used will be quite tricky. I think it would be good to introduce a testing beacon that we can ping to verify state of the plugin and verify basic diagnostics.

I've created:
https://aws-partner.atlassian.net/browse/CHET-307


theghostwhoforks:
this should resolve the failing actions 
https://github.com/atlassian-labs/dc-migration-assistant/actions

PR #159, opened 2020-04-29
CHET-no-issue: Do not call getCurrentSevice multiple times in the sam…

theghostwhoforks:
…e controller method

AFAIK, these calls aren't memoized. So, we make multiple DB calls in the method, when just one would suffice


bpar476:
It should be hitting the AO cache and not the actual DB but there's no reason not to do this.

PR #161, opened 2020-04-30
Chet 311 parameterise cfn template url

theghostwhoforks:
OK, so this PR adds the ability to parameterise template URLs both in the FE and backend code flows

On the server side, if we want to specify a template URL for deploy, we should specify a system property like so - `-Dquickstart.template.url=https://dcd-slinghost-templates.s3-ap-southeast-2.amazonaws.com/trebuchet/tests/existing-vpc/quickstart-jira-dc.template.yaml ` to specify the DC template without VPC.

On the front end, things get a bit more interesting. Obviously, the FE app can't read environment variables at runtime. We can inject the template URLs at build time. To do this, we create a webpack plugin that reads from a `.env` file (located in the project root) and creates a dictionary with keys having a prefix (`process.env.`) and suffixed by the key value defined in the `.env` file. 

In this case, the dictionary would be `{process.env.REACT_APP_QUICKSTART_PARAMETERS_URL=https://trebuchet-public-resources.s3.amazonaws.com/quickstart-jira-dc-with-vpc.template.parameters.yaml}`

After this, when we use `process.env.REACT_APP_QUICKSTART_PARAMETERS_URL` in any `.tsx` file, it will be replaced with the value above.

When we run this using `yarn:dev-server`, we can additionally specify overrides in a file called `.env.development` (this has been `.gitignored`) and further override the value in the `.env` file, which is checked in to source control.

When we compile the FE project using `mvn install`, the .env file is read and the default, production value of the URL is used during the compilation flow. Subsequently, this is packaged and installed as a plugin, we can see the form rendered with the values from the template file present in `.env`

If we need to override this with the, say, existing VPC template, we should create an `.env.development` file (or modify the .env file for local testing only) and update the URL to the without VPC parameters file (`https://dcd-slinghost-templates.s3-ap-southeast-2.amazonaws.com/trebuchet/tests/existing-vpc/quickstart-jira-dc.template.yaml`)

We can then build the FE project by running `mvn clean install -DskipTests -Dyarn.cmd.build=build:dev
` in development mode to pick these changes in the `.env.development` file and when we package & install the plugin, we should be able to see the parameters being based on the URL in the `.env.development` file

It's imperative that the .env file remains checked-in and has the prroduction parameters file URL in it at all times. This perhaps requires a build time test that I can write.

Do try this out and let me know if anything needs to be changed/modified.

All template and the parameters file (publicly accessible) can be found under - 
`https://s3.console.aws.amazon.com/s3/buckets/dcd-slinghost-templates/trebuchet/tests/existing-vpc/?region=ap-southeast-2&tab=overview`

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
@@ -44,6 +44,8 @@ public class QuickstartDeploymentService extends CloudformationDeploymentService
     private final Logger logger = LoggerFactory.getLogger(QuickstartDeploymentService.class);
     private static final String QUICKSTART_TEMPLATE_URL = "https://aws-quickstart.s3.amazonaws.com/quickstart-atlassian-jira/templates/quickstart-jira-dc-with-vpc.template.yaml";
 
+    private static final String templateUrl = System.getProperty("quickstart.template.url", QUICKSTART_TEMPLATE_URL);
+
     private final MigrationService migrationService;
     private final TargetDbCredentialsStorageService dbCredentialsStorageService;
     private final AWSMigrationHelperDeploymentService migrationHelperDeploymentService;
@@ -74,7 +76,7 @@ public class QuickstartDeploymentService extends CloudformationDeploymentService
         migrationService.transition(MigrationStage.PROVISION_APPLICATION_WAIT);
 
         logger.info("deploying application stack");
-        super.deployCloudformationStack(QUICKSTART_TEMPLATE_URL, deploymentId, params);
+        super.deployCloudformationStack(templateUrl, deploymentId, params);
 
         addDeploymentIdToMigrationContext(deploymentId);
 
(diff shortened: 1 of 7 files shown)


theghostwhoforks:
To make e2e testing even faster, suppose we know that the VPC template is working as expected and we want to test different flows in FS and DB migration steps.

We can create a template that just emits the input parameters as outputs. Let's say, we emit all the values required by the state machine (deployment ID, endpoint addresses etc..), with the above approach, we can get a straight through provisioning case, while still integrating with CFn (for both the application template and the migration-helper template) and storing values in the migration context. 

This will give us a case to pass through to the next stage of the state machine without having to add a no-op implementation (where will we get the values to enter in the context?) test FS and DB restore easily

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
```diff
@@ -44,6 +44,8 @@
     private final Logger logger = LoggerFactory.getLogger(QuickstartDeploymentService.class);
     private static final String QUICKSTART_TEMPLATE_URL = "https://aws-quickstart.s3.amazonaws.com/quickstart-atlassian-jira/templates/quickstart-jira-dc-with-vpc.template.yaml";
 
+    private static final String templateUrl = System.getProperty("quickstart.template.url", QUICKSTART_TEMPLATE_URL);
```
dylanrathbone, line 47: convention is to UPPERCASE constants?

theghostwhoforks: The `QUICKSTART_TEMPLATE_URL` is a constant.

The `templateUrl`, like the `logger` above is a final field.

dylanrathbone: Ahh! Read it in the wrong context!

PR #162, opened 2020-04-30
CHET-315: Add warning page

nanux:
Modal approach for comparison - https://github.com/atlassian-labs/dc-migration-assistant/compare/CHET-312-display-modal

Screenshot
image

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/warning/WarningStage.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/warning/WarningStage.tsx
@@ -1,0 +1,45 @@
+import React, { FunctionComponent, useState } from 'react';
+import SectionMessage from '@atlaskit/section-message';
+import { Checkbox } from '@atlaskit/checkbox';
+import Button from '@atlaskit/button';
+import { useHistory } from 'react-router-dom';
+import { dbPath } from '../../utils/RoutePaths';
+
+export const WarningStagePage: FunctionComponent = () => {
+    const history = useHistory();
+
+    const [agreed, setAgreed] = useState<boolean>(false);
+
+    const handleClick = () => {
+        history.push(dbPath);
+    };
+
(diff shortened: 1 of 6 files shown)


bpar476:
Can we get screenshots? :)

frontend/src/main/dc-migration-assistant-fe/components/warning/WarningStage.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { useHistory } from 'react-router-dom';
+import { dbPath } from '../../utils/RoutePaths';
+
+export const WarningStagePage: FunctionComponent = () => {
+    const history = useHistory();
+
+    const [agreed, setAgreed] = useState<boolean>(false);
+
+    const handleClick = () => {
+        history.push(dbPath);
```
bpar476, line 14: Through the rest of the code-base, we've been using `<Redirect />` components instead of the `history` API. `Redirect` is the preffered method for programmatic routing as it aligns with React's goal of being declarative.

The pattern is to use component state to track if there should be a redirect, if that state is true, return:
`<Redirect to={dbPath} push />`

frontend/src/main/dc-migration-assistant-fe/components/warning/WarningStage.tsx
```diff
@@ -0,0 +1,45 @@
...
+    };
+
+    const agreeOnClick = (event: any) => {
+        setAgreed(event.target.checked);
+    };
+    return (
+        <div>
+            <h1>Step 5 of 7: Redirect users</h1>
```
bpar476, line 22: The step numbers are a bit wonky because we don't have an ASI page. This probably be "Step 4 of 6". The other pages will need to have their strings updated to "Step X of 6" as well.

bpar476:
Clean approach :)

bpar476:
Definitely more fond of this approach than the modal

PR #164, opened 2020-05-01
Chet 99 learn more

theghostwhoforks:
This generates a learn more link in 
1. Validation
2. the shared migration transfer page.

For 2, I think we need to make it simple and link to a single section in the CAC article with a description of how to recover from errors. The error text in the section comes from the API, the cost of creating logic to handle which learn more link to render based on the calling component is higher than having a single error/recovery section in the CAC page. 
cc: @ddomingorh - Thoughts?

Attaching screen shots for reference -

Screenshot 2020-05-01 at 12 56 35

Common learn more link in the transfer page component

Screenshot 2020-05-04 at 09 03 12

Error page on CAC

Screenshot 2020-05-04 at 09 03 25


bpar476:
The design has a learn more link for the description of the "Copy content" page as well @ddomingorh do we still need that?

bpar476:
And what is it meant to link to?

ddomingorh:
> The design has a learn more link for the description of the "Copy content" page as well @ddomingorh do we still need that?

@theghostwhoforks and I discussed this last Friday, IIRC we decided to cut the Learn more from this page.

The Learn more should only appear if there's an error, and it should link to the CAC section in the screenshot above.

PR #166, opened 2020-05-01
CHET-320: Create common CancelButton component and update the design

nanux:
* Internationalization for Warning page
* Extract common `CancelButton`
* The next page after FS migration is the Warning page now
* Add cancel button on the warning page
* Rename the submit button from `Continue` to `Next`

image

image

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/shared/CancelButton.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/shared/CancelButton.tsx
@@ -4,17 +4,14 @@ import Button from '@atlaskit/button';
 import { overviewPath } from '../../utils/RoutePaths';
 import { I18n } from '../../atlassian/mocks/@atlassian/wrm-react-i18n';
 
-export const defaultButtonStyle = {
-    padding: '5px',
-};
-export const marginButtonStyle = {
-    ...defaultButtonStyle,
-    marginRight: '20px',
+export const cancelButtonStyle = {
+    paddingLeft: '5px',
+    marginLeft: '20px',
 };
 
 export const CancelButton: FunctionComponent = () => (
     <Link to={overviewPath}>
-        <Button style={marginButtonStyle}>
+        <Button style={cancelButtonStyle}>
             {I18n.getText('atlassian.migration.datacenter.generic.cancel')}
         </Button>
     </Link>
--- a/frontend/src/main/dc-migration-assistant-fe/components/shared/CancelButton.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/shared/CancelButton.tsx
@@ -1,0 +1,21 @@
+import React, { FunctionComponent } from 'react';
+import { Link } from 'react-router-dom';
+import Button from '@atlaskit/button';
+import { overviewPath } from '../../utils/RoutePaths';
+import { I18n } from '../../atlassian/mocks/@atlassian/wrm-react-i18n';
+
+export const defaultButtonStyle = {
+    padding: '5px',
+};
+export const marginButtonStyle = {
+    ...defaultButtonStyle,
+    marginRight: '20px',
+};
+
+export const CancelButton: FunctionComponent = () => (
+    <Link to={overviewPath}>
+        <Button style={marginButtonStyle}>
+            {I18n.getText('atlassian.migration.datacenter.generic.cancel')}
+        </Button>
+    </Link>
+);
--- a/frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPageActions.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPageActions.tsx
@@ -1,7 +1,7 @@
 import React, { FunctionComponent } from 'react';
 import Button from '@atlaskit/button';
 import { Link } from 'react-router-dom';
-import { CancelButton, marginButtonStyle, defaultButtonStyle } from './CancelButton';
+import { CancelButton } from './CancelButton';
 
 export type MigrationTransferActionsProps = {
     /**
@@ -51,6 +51,14 @@ export const MigrationTransferActions: FunctionComponent<MigrationTransferAction
     started,
     loading,
 }) => {
+    const defaultButtonStyle = {
+        padding: '5px',
+    };
+
+    const marginButtonStyle = {
+        ...defaultButtonStyle,
+        marginRight: '20px',
+    };
     const StartButton = (
         <Button
             style={marginButtonStyle}
--- a/frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPageActions.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPageActions.tsx
@@ -1,8 +1,7 @@
 import React, { FunctionComponent } from 'react';
 import Button from '@atlaskit/button';
 import { Link } from 'react-router-dom';
-import { I18n } from '@atlassian/wrm-react-i18n';
-import { overviewPath } from '../../utils/RoutePaths';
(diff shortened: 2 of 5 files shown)


frontend/src/main/dc-migration-assistant-fe/components/shared/CancelButton.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export const defaultButtonStyle = {
+    padding: '5px',
+};
+export const marginButtonStyle = {
+    ...defaultButtonStyle,
+    marginRight: '20px',
+};
+
+export const CancelButton: FunctionComponent = () => (
```
nanux, line 15: I've extracted this to its own component as I am expecting this will be used in multiple places (2 at the moment). Please let me know if that is the right React pattern.

frontend/src/main/dc-migration-assistant-fe/components/warning/WarningStage.tsx
```diff
@@ -27,25 +25,52 @@ export const WarningStagePage: FunctionComponent = () => {
...
+            <div style={checkboxSectionStyle}>
+                <Checkbox
+                    value="agree"
+                    label="I'm ready for the next step"
+                    onChange={agreeOnClick}
+                    name="agree"
+                />
+            </div>
+            {NextButton}
+            <CancelButton css={marginButtonStyle} />
```
nanux, line 73: is `css` the "right" way how to style componts?

theghostwhoforks: we've been using with a style attribute like say, `<Button style={{ marginLeft: '20px',  paddingLeft: '5px' }}....` add styles to the button

theghostwhoforks: Which is what you're using inside the component 👍 

That said, I'm not too sure if the `css` attribute is right key here. I've seen existing atlaskit components using the css prop, so it may be correct?

frontend/src/main/dc-migration-assistant-fe/components/shared/CancelButton.tsx
theghostwhoforks: This is good. Eventually, clicking on cancel should take the user back to the landing page and render an error flag like so - 
https://app.zeplin.io/project/5e9ea63418e4a27a9986919e/screen/5ea94a3eb51a9023c917c2e5

This would be a good place to add the click handler as well in a future strory

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPageActions.tsx
```diff
@@ -1,8 +1,7 @@
 import React, { FunctionComponent } from 'react';
 import Button from '@atlaskit/button';
 import { Link } from 'react-router-dom';
-import { I18n } from '@atlassian/wrm-react-i18n';
-import { overviewPath } from '../../utils/RoutePaths';
+import { CancelButton, marginButtonStyle, defaultButtonStyle } from './CancelButton';
```
theghostwhoforks, line 4: do we need to import the styles here?

frontend/src/main/dc-migration-assistant-fe/components/warning/WarningStage.tsx
bpar476: I believe the `style` prop is what we've been using for buttons. What I would do is add the `Button` props to the `CancelButton` and pass them down.

bpar476:
> The `Section` component is too close to the paragraph element before it. It needs to breathe.
> 
> Do we have a pattern for that? Also, I assume we will tackle the container width - to make it more readable - later and apply it across the codebase. Is this assumption correct?

I would just add some `margin-bottom` to the paragraph using `styled-components`

```
const Paragraph = styled.p`
    margin-bottom: 10px;
`
```

frontend/src/main/dc-migration-assistant-fe/components/warning/WarningStage.tsx
```diff
@@ -27,25 +25,52 @@ export const WarningStagePage: FunctionComponent = () => {
...
-            <Checkbox
-                value="agree"
-                label="I'm ready for the next step"
-                onChange={agreeOnClick}
-                name="agree"
-            />
-            <Button isDisabled={!agreed} onClick={handleClick}>
-                Continue
-            </Button>
+            <div style={checkboxSectionStyle}>
```
bpar476, line 64: When using a raw div we should be using styled components to set the style:

```
const CheckBoxContainer = styled.div`
    margin: 20px;
`
```

bpar476: Then you can replace the `<div>` with `<CheckBoxContainer>`

frontend/src/main/dc-migration-assistant-fe/components/shared/CancelButton.tsx
```diff
@@ -0,0 +1,21 @@
+import React, { FunctionComponent } from 'react';
+import { Link } from 'react-router-dom';
+import Button from '@atlaskit/button';
+import { overviewPath } from '../../utils/RoutePaths';
+import { I18n } from '../../atlassian/mocks/@atlassian/wrm-react-i18n';
+
+export const defaultButtonStyle = {
```
bpar476, line 7: Why are the styles exported? 🤔

frontend/src/main/dc-migration-assistant-fe/components/warning/WarningStage.tsx
bpar476: You can actually do something quite tricky with typescript where you say the props for the `CancelButton` is equal to whatever you want, plus the props of a button (minus the ones you use).

```
type ForbiddenCancelButtonProps = {
    onClick: () => void;
    ... any other props of the button you don't want overriden
}

type CancelButtonProps = Exclude<DOMAttributes<HTMLButtonElement>, ForbiddenCancelButtonProps>


```
this means that the cancel button props will have all the props of a `Button` minus the props you put in `ForbiddenCancelButtonProps`

frontend/src/main/dc-migration-assistant-fe/components/shared/CancelButton.tsx
nanux: They are still used in the same place as they were before:
https://github.com/atlassian-labs/dc-migration-assistant/pull/166/files#diff-10671ce1316648976e39ae6854690534R4

It doesn't look right but we are reusing the style there as well, what would be a better way?

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPageActions.tsx
bpar476: I think it would be better to have a component `<MarginButton>` and `<DefaultButton>` defined in separate files instead of having the `CancelButton` also define styles for other kinds of button, if you know what I mean 😅

frontend/src/main/dc-migration-assistant-fe/components/shared/CancelButton.tsx
bpar476: If we use that style for _every_ button then we would create a component as you've done for the cancel button. However, when it comes to CSS I tend to worry less about DRY as you won't always want to change properties like margin, width, or padding across your entiner application.

We can leave this for now though :)

nanux: I found out I was too eager with DRY anyway and introduced a style change - Cancel button had : `style={{ marginLeft: '20px', paddingLeft: '5px' }}` and I've tried to  merge it with rest of the buttons. I've fixed it and I put back the css styles for the buttons in the `MigrateTransferPageActions` component.

PR #167, opened 2020-05-04
CHET-273: UI test automation

tarka:
Interim checkin; convert to Cypress and add ability to reset migration state on test start.


frontend/src/main/dc-migration-assistant-fe/api/migration.ts
```diff
@@ -98,4 +99,20 @@ export const migration = {
             res.json()
         );
     },
+    resetMigration: (): Promise<void> => {
```
bpar476, line 102: This is a dev endpoint right? Should we throw an error if this is called in production?

jira-plugin/pom.xml
```diff
@@ -63,7 +63,7 @@
                 <version>${amps.version}</version>
                 <extensions>true</extensions>
                 <configuration>
-                    <jvmArgs>${amps.jvm.args}</jvmArgs>
+                    <jvmArgs>${amps.jvm.args} -Dspring.profiles.active=allowAnyTransition</jvmArgs>
```
bpar476, line 66: We don't want this for every build do we?

frontend/src/main/dc-migration-assistant-fe/api/migration.ts
tarka: It will 404, as the endpoint won't exist if the profile isn't enabled.

jira-plugin/pom.xml
tarka: Unless I'm mistaken this is the `amps:run` configuration; hence `enableQuickReload`.

tarka:
WIP ATM; I'll add some docs once I've fleshed out some more of the page tests.

frontend/src/main/dc-migration-assistant-fe/api/migration.ts
nanux: After this is https://github.com/atlassian-labs/dc-migration-assistant/pull/169 merged, it is production endpoint but it requires the `error` stage.

jira-plugin/pom.xml
nanux: I think currently this just allow any transition (we are moving the reset endpoint to production) and I think it is safe to have this as part of `amps:run` and I hope `amps:debug` will pick it up as well.

tarka:
> This is really neat! Could we add something to the `README` on how to run these for automated testing or for speeding up manual testing?

I'll add testing notes when I'm a bit further through, I'm still working out some details.

frontend/src/main/dc-migration-assistant-fe/api/migration.ts
```diff
@@ -20,6 +20,7 @@ import { homePath, quickstartStatusPath, awsAuthPath } from '../utils/RoutePaths
 enum RestApiPathConstants {
     migrationRestPath = `migration`,
     migrationSummaryRestPath = `migration/summary`,
+    migrationResetPath = `develop/migration/reset`,
```
theghostwhoforks, line 23: As @nanux  mentioned, this needs to be just `migration/reset` now

The state of the state machine should be in error state for reset to work.

jira-plugin/pom.xml
bpar476: Ah okay :)

frontend/src/main/dc-migration-assistant-fe/api/migration.ts
tarka: If there's no objections I'll re-add the develop reset endpoint; for testing we need a fast-reset function.

PR #168, opened 2020-05-04
CHET-279: load current migration step

bpar476:
## Summary

* Create wrapper component which renders:
  * All routes
  * A component which gets the migration and redirects to the corresponding page
* Adds a mapping from migration stage to page
* Cleans up some old routes and pages


frontend/src/main/dc-migration-assistant-fe/components/MigrationStageRenderer.tsx
```diff
@@ -0,0 +1,24 @@
...
+    const [loading, setLoading] = useState<boolean>(true);
+    const [currentStage, setStage] = useState<MigrationStage>();
+
+    useEffect(() => {
+        setLoading(true);
+        migration.getMigrationStage().then(stage => {
+            setStage(stage);
+            setLoading(false);
+        });
+    }, []);
```
nanux, line 23: Just to make sure that I understand the `[]` correctly - this means that that the `useEffect()` is called only once and it is independent on the props of the component. We can do this because we check for the stage once and then we call the redirect - so we don't keep this component rendered?

https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

frontend/src/main/dc-migration-assistant-fe/components/Routes.tsx
```diff
@@ -19,41 +19,36 @@ import { BrowserRouter as Router, Route, Switch } from 'react-router-dom';
...
-            </Route>
             <Route path={awsBasePath}>
                 <AWSRoutes />
             </Route>
-            <Route path={fsPath}>
+            <Route exact path={fsPath}>
                 <FileSystemTransferPage />
             </Route>
-            <Route path={warningPath}>
+            <Route exactpath={warningPath}>
```
nanux: Should there be a space: `exact path`?

frontend/src/main/dc-migration-assistant-fe/components/MigrationStageRenderer.tsx
bpar476: Correct! After the redirect the component will be unmounted.

PR #169, opened 2020-05-04
Chet 330 reset migrations

theghostwhoforks:
Reset a migration only when in error state to prevent accidental reset of migration


api/src/test/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpointTest.kt
```diff
@@ -122,4 +116,31 @@ class MigrationEndpointTest {
...
+    fun shouldNotResetMigrationWhenCurrentMigrationStageIsNotError() {
+        every { migrationService.currentStage } returns MigrationStage.AUTHENTICATION
+
+        val response = sut.resetMigration()
+
+        assertEquals(
+                Response.Status.CONFLICT.statusCode,
+                response.status
+        )
+        verify(exactly = 0) { migrationService.deleteMigrations() }
```
theghostwhoforks, line 144: is this the best way to do a verify never?

nanux:
`com/atlassian/migration/datacenter/spi/MigrationService.java:64` has this javadoc:

```
/**
     * Deletes all migrations and associated contexts. It should be used only in developer testing.
     */

```

We can remove the `It should be used only in developer testing.` part.

api/src/test/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpointTest.kt
nanux: Seems legit:

---
Specifying exactly = 0 you can verify call did not happen at all:
```
verify(exactly = 0) { 
  mock1.call(5)
}
```

Sometimes you want to check if mocked DOC was not called at all. This can be achieved with special wasNot Called construct. It is equivalent to verifyZeroInteractions in Mockito.

```
verify { 
  mock1 wasNot Called 
}
```
https://blog.kotlin-academy.com/mocking-is-not-rocket-science-expected-behavior-and-behavior-verification-3862dd0e0f03#7506

api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
```diff
@@ -20,11 +20,7 @@ import com.atlassian.migration.datacenter.spi.MigrationService
 import com.atlassian.migration.datacenter.spi.MigrationStage
 import com.atlassian.migration.datacenter.spi.exceptions.InvalidMigrationStageError
 import com.atlassian.migration.datacenter.spi.exceptions.MigrationAlreadyExistsException
-import javax.ws.rs.Consumes
-import javax.ws.rs.GET
-import javax.ws.rs.POST
-import javax.ws.rs.Path
-import javax.ws.rs.Produces
+import javax.ws.rs.*
```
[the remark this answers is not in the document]
bpar476: I have the same setting

api/src/test/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpointTest.kt
theghostwhoforks: 👍

api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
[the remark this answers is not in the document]
theghostwhoforks: Screenshot 2020-05-04 at 14 01 56


Should this be anything else? (i.e. name count is set to 3)

PR #171, opened 2020-05-04
CHET-332: Meaningful errors

bpar476:
## Summary

* Add message parameter to error transition
* Store error message in migration context when transitioning
* Add some meaningful error messages in all error transitions
* Expose error message in migration summary API


bpar476:
Lots of files changed because of many error messages 😢

core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -138,7 +138,7 @@ public void abortMigration() throws InvalidMigrationStageError {
 
         logger.warn("Aborting running filesystem migration");
 
-        migrationService.error();
+        migrationService.error("file system migration was aborted");
```
nanux, line 141: /nit `File`

core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/DatabaseRestoreStageTransitionCallback.java
```diff
@@ -46,7 +46,7 @@ public void transitionToServiceNextStage() throws InvalidMigrationStageError {
     }
 
     @Override
-    public void transitionToServiceErrorStage() {
-        migrationService.error();
+    public void transitionToServiceErrorStage(String error) {
```
nanux, line 49: /nit - we can be probably consistent across all the `transitionToServiceErrorStage` and call the parameter `errorMessage` or just `error` (not `message`)

core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/SsmPsqlDatabaseRestoreService.java
```diff
@@ -58,9 +58,9 @@ public void restoreDatabase(DatabaseRestoreStageTransitionCallback restoreStageT
             consumer.handleCommandOutput(maxCommandRetries);
             restoreStageTransitionCallback.transitionToServiceNextStage();
         } catch (SuccessfulSSMCommandConsumer.UnsuccessfulSSMCommandInvocationException | SuccessfulSSMCommandConsumer.SSMCommandInvocationProcessingError e) {
-            restoreStageTransitionCallback.transitionToServiceErrorStage();
-            throw new DatabaseMigrationFailure("Unable to invoke database download command", e);
+            final String errorMessage = "Unable to invoke database download command";
```
nanux, line 61: Couldn't this also mean that the database restore failed? E.g. the `pg_restore` script failed?

core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -109,7 +109,7 @@ public Boolean scheduleMigration() throws InvalidMigrationStageError
         boolean result = migrationRunner.runMigration(jobId, jobRunner);
 
         if (!result) {
-            migrationService.error();
+            migrationService.error("Error starting filesystem migration job. Check application logs for details");
```
nanux, line 112: I wonder if we shouldn't send people always to the `application logs` I would probably make the message part of the error page by default.

core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -171,7 +171,7 @@ public void abortMigration() throws InvalidMigrationStageError {
         fsUploader.abort();
         report.setStatus(FAILED);
 
-        migrationService.error();
+        migrationService.error("file system migration was aborted");
```
nanux, line 174: /nit `File`  - I think we should be consistent with the text if we are going to display it on he page.

core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
bpar476: Sounds good to me :)

core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/SsmPsqlDatabaseRestoreService.java
bpar476: Correct! I'll update the string.

bpar476: We could use the SSM command output to generate a more specific error message in future.

PR #173, opened 2020-05-04
CHET-23: Add a skeleton error page with routing

theghostwhoforks:
/migration-error renders a skeleton error page like so - 

Screenshot 2020-05-04 at 15 27 17


frontend/src/main/dc-migration-assistant-fe/utils/RoutePaths.ts
```diff
@@ -41,3 +41,5 @@ export const dbPath = `${routePrefix}/db`;
 export const warningPath = `${routePrefix}/warning`;
 
 export const validationPath = `${routePrefix}/validation`;
+
+export const migrationErrorPath = `${routePrefix}/migration-error`;
```
nanux, line 45: https://github.com/atlassian-labs/dc-migration-assistant/blob/d78e038361c3350899490caadf7ff057840f96e7/frontend/src/main/dc-migration-assistant-fe/utils/migration-stage-to-path.ts#L4

You can also change the:

```
case 'error':
            throw new Error('page error not implemented');
```

to load correct error page.

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
```diff
@@ -0,0 +1,53 @@
...
+ * limitations under the License.
+ */
+
+import React, { ReactElement, useEffect, useState } from 'react';
+import styled from 'styled-components';
+import { I18n } from '@atlassian/wrm-react-i18n';
+
+import { migration, MigrationStage } from '../api/migration';
+
+const MigrationErrorContainer = styled.div`
```
nanux, line 23: Just general FE question - we have this style all over the app. What is the best approach to unify this in future? Creating something like `<PageContainer>` component and wrap all pages in it?

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
```diff
@@ -0,0 +1,53 @@
...
+    display: flex;
+    flex-direction: column;
+    align-items: center;
+`;
+
+export const MigrationError = (): ReactElement => {
+    useEffect(() => {
+        migration.getMigrationStage().then((stage: string) => {
+            if (stage !== MigrationStage.ERROR.valueOf()) {
+                console.log('Stage is not in error stage');
```
nanux, line 33: Should we redirect to some other page when there is no error?

bpar476: I think we should redirect to the home page 👍  But maybe not a blocker for just getting the page in place

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
theghostwhoforks: Sounds like a good shareable component to create!

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
theghostwhoforks: Redirecting to home is an option for now. Rendering static text when no error can be found is another.

This will all go away soon with the updated design, but the underlying components must be reusable

PR #176, opened 2020-05-04
Chet 23 reset migration

theghostwhoforks:
Allow admin to reset migration only when state is error.

Once the migration has been reset, a user is redirected to the home page

This does not handle the case when reset migration fails. i.e. how do we handle an error on a error page. Render a flag, perhaps?

Updated screenshots 

Screenshot 2020-05-05 at 18 09 38

Screenshot 2020-05-05 at 18 09 24


bpar476:
Under what circumstances would restarting fail?

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
```diff
@@ -14,40 +14,58 @@
...
+                    This is the skeleton error page for the migration plugin{' '}
+                    <a
+                        target="_blank"
+                        rel="noreferrer noopener"
+                        href="https://confluence.atlassian.com/jirakb/how-to-use-the-data-center-migration-app-to-migrate-jira-to-an-aws-cluster-1005781495.html"
+                    >
+                        {I18n.getText('atlassian.migration.datacenter.common.learn_more')}
+                    </a>
+                </p>
+                <Button isDisabled={!isInErrorState} onClick={resetMigration} appearance="primary">
```
bpar476, line 65: I think instead of having a disabled button until we determine if we're in an error state we have a spinner while fetching the state and if the state is not "error" redirect them to home or something. I think it would be confusing if a user accidentally went to the error page and saw that there was an error but they can't follow the call to action.

theghostwhoforks:
> Under what circumstances would restarting fail?

API integration failure, AO issues etc.. This is why most error pages are static so that even when the underlying API fails, the statically rendered error page will still be rendered

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
theghostwhoforks: Hmmm - I like this - let me get this up and running.

The only thing I would change is to add a link to route to the relevant stage instead of rendering - that way the user knows what is happening instead being automatically redirected

theghostwhoforks: This is what it looks like now
Screenshot 2020-05-05 at 11 33 28

Would adding an auto redirect be more preferable? cc: @bpar476 @nanux

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
```diff
@@ -14,40 +14,101 @@
...
+                    >
+                        {learnMoreText.toLowerCase()}
+                    </a>{' '}
+                    about why this may have occurred
+                    {additionalErrorContext} <br />
+                </p>
+                <Button onClick={resetMigrationFunc} appearance="primary">
+                    {I18n.getText('atlassian.migration.datacenter.error.reset.button')}
+                </Button>
+            </>
```
dylanrathbone, line 66: my react knowledge is very basic, but it might be worth considering turning this into a re-usable component that accepts props making it dynamic.

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
```diff
@@ -14,40 +14,101 @@
...
+                This is an error page that is shown only when a migration fails. You currently have
+                a migration in progress.
+            </p>
+            <p>Click the button below to view the latest status.</p>
+            <Link to={getPathForStage(currentStage)}>
+                <Button appearance="primary">
+                    {I18n.getText('atlassian.migration.datacenter.error.current.stage')}
+                </Button>
+            </Link>
+        </>
```
dylanrathbone, line 82: same as above comment

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
bpar476: I'm not too fussed as this is kind of just a band-aid so we don't confuse the user. For MVP2 we should have a wrapper around each page that redirects to the current step if they navigate there out of band.

bpar476:
> > Under what circumstances would restarting fail?
> 
> API integration failure, AO issues etc.. This is why most error pages are static so that even when the underlying API fails, the statically rendered error page will still be rendered

I suppose if it fails a "try again in a moment" flag is sufficient but I'd defer to @ashking2 here

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
theghostwhoforks: Yup

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
theghostwhoforks: The element is invoked by a component. The entire element is re-rendered

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
```diff
@@ -14,40 +14,116 @@
...
-                    target="_blank"
-                    rel="noreferrer noopener"
-                    href="https://confluence.atlassian.com/jirakb/how-to-use-the-data-center-migration-app-to-migrate-jira-to-an-aws-cluster-1005781495.html"
-                >
-                    {I18n.getText('atlassian.migration.datacenter.common.learn_more')}
-                </a>
-            </p>
-        </MigrationErrorContainer>
+        <>
+            {redirectToNewMigration && <Redirect to={homePath} push />}
```
bpar476, line 116: I would probably split up these conditionals as we can avoid some accidental bad state. For example if `redirectToNewMigration` and `!isLoading` are both true, it looks like both the `Redirect` and the `ErrorContainer` will be rendered but really the redirect will take the user away before the error container is rendered.

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
```diff
@@ -14,40 +14,116 @@
...
+                <Button onClick={resetMigrationFunc} appearance="primary" style={buttonStyle}>
+                    {I18n.getText('atlassian.migration.datacenter.error.reset.button')}
+                </Button>
+            </>
+        );
+    }
+
+    return (
+        <>
+            <h2>Migration In Progress</h2>
```
bpar476, line 76: Are we going to add `I18n` for this string?

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
```diff
@@ -14,40 +14,116 @@
...
+    additionalErrorContext,
+}: ResetMigrationProps): ReactElement => {
+    const buttonStyle = {
+        marginTop: '20px',
+    };
+
+    if (currentStage === MigrationStage.ERROR) {
+        return (
+            <>
+                <h2>Error</h2>
```
bpar476, line 53: I18n again here? I think we have a string for "Error" already

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
```diff
@@ -14,40 +14,116 @@
...
-        </MigrationErrorContainer>
+        <>
+            {redirectToNewMigration && <Redirect to={homePath} push />}
+            {isLoading && <Spinner />}
+            {!isLoading && (
+                <MigrationErrorContainer>
+                    <StageAwareMigrationSection
+                        currentStage={currentStage}
+                        resetMigrationFunc={resetMigration}
+                        additionalErrorContext=""
```
bpar476, line 123: Is this where the error from the API will go?

theghostwhoforks: Yeah - I just created a placeholder for this for now

frontend/src/main/dc-migration-assistant-fe/components/MigrationError.tsx
theghostwhoforks: I've encapsulated this behaviour within the container now - if isLoading is set, then we just render a spinner. If not, then we render the error message

bpar476: We still have the chance that the error message and the `Redirect` can be rendered at the same time which is typically avoided.

theghostwhoforks: This will go away soon and I didn't want to spend more effort on this - it's a bit coupled for sure, but given it will stay for a short while, this should be ok.

bpar476: The usual way to do a redirect is:
```
if (doRedirect) {
   return <Redirect />
}

return <MyComponent />
```

theghostwhoforks: 👍 

We'll need to change this pattern in a couple of other places then - https://github.com/atlassian-labs/dc-migration-assistant/blob/master/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx#L270 (I found this when I was looking at how redirect was implemented)

PR #179, opened 2020-05-05
no-issue: Fix disappearing soy sidebar

mk-atlassian:
If the admin decorator does not have a valid section the sidebar part of the template will crash and render nothing.

Why can't you use `advanced_menu_section/advanced_section` (or the longer version on DAC) for the section? IDK TBH, couldn't find out what the section name is using web-fragment finder since that seems to cause 500s in newer versions of Jira S. I tried a couple of different options I ripped from the Jira source code (soy, jsp, and plugin.xml files). Maybe this is a bug to be reported to Jira?

Why won't the active tab stay highlighted? It's finicky sometimes you need to reload the page, sometimes it's because there's an error.


theghostwhoforks:
https://developer.atlassian.com/server/jira/platform/project-settings-ui-locations/#ensuring-a-web-item-is-displayed-by-default-in-a-project-configuration

TIL - web-fragment-finder is a 3rd party tool

bpar476:
<3 It works!

bpar476:
image

The menu item has moved from "Advanced" to it's own little happy spot. It's not ideal but I think the SPA fix is worth it. We can revisit the location of the plugin in the future.

bpar476:
> image
> 
> The menu item has moved from "Advanced" to it's own little happy spot. It's not ideal but I think the SPA fix is worth it. We can revisit the location of the plugin in the future.

@ashking2 is this acceptable for MVP 1? Without this change we completely lose the admin menu if the user clicks "Back" or "Forward" or refreshes the page.

jira-plugin/src/main/resources/atlassian-plugin.xml
```diff
@@ -48,7 +48,7 @@
         <link linkId="dc-migration-assistant">/plugins/servlet/dc-migration-assistant</link>
     </web-item>
 
-    <web-item key="jira-menu-item" name="Jira DC Migration Plugin" section="advanced_menu_section/advanced_section"
+    <web-item key="migrationAssistantItem" name="Jira DC Migration Plugin" section="admin_system_menu/advanced_menu_section"
```
theghostwhoforks, line 51: Question - Shouldn't this be `admin_system_menu/advanced_menu_section/advanced_section`?

Source: https://developer.atlassian.com/server/jira/platform/administration-ui-locations/

theghostwhoforks: oh is this what you mean by the version on DAC - yeah this is a bug that we need to raise

mk-atlassian: Yeah soz

PR #182, opened 2020-05-06
CHET-322: Exit with 0 with non-critical errors from pg_restore

nanux:
* Return either exit code from `pg_restore` if it wasn't `0` or print information with the errors from migration and return `0` as exit code. `pg_dump` returns non-zero code only with fatal errors - e.g. the directory is missing.
* Formatting of the error messages (when we expect people to read the log)
* Added ability to provide migrationStackId via system property
* Uploaded `migration-helper.yml` to `trebuchet-public-resources` S3 bucket and made it public

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
@@ -101,7 +101,8 @@ public class AWSMigrationHelperDeploymentService extends CloudformationDeploymen
 
     @Override
     protected void handleSuccessfulDeployment() {
-        Optional<Stack> maybeStack = cfnApi.getStack(migrationService.getCurrentContext().getHelperStackDeploymentId());
+        String stackId = System.getProperty("com.atlassian.migration.migrationStack.id", migrationService.getCurrentContext().getHelperStackDeploymentId());
+        Optional<Stack> maybeStack = cfnApi.getStack(stackId);
         if (!maybeStack.isPresent()) {
             throw new InfrastructureDeploymentError("stack was not found by DescribeStack even though it succeeded");
         }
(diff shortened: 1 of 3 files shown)


templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -147,28 +147,38 @@ Resources:
...
                     else
-                            echo "Errors:"
+                            echo "SQL migration errors:"
                             echo "$ERROR_OUTPUT"
-                            echo "ERROR: Database migration wasn't successful"
-                            exit 1
+                            echo "WARNING: Database migration finished with some errors, please review the log"
                     fi
+
+                    exit 0
```
nanux, line 181: This is the main change - returns 0 even if there are errors in the restore phase.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
```diff
@@ -101,7 +101,8 @@ private void resetStackOutputs() {
 
     @Override
     protected void handleSuccessfulDeployment() {
-        Optional<Stack> maybeStack = cfnApi.getStack(migrationService.getCurrentContext().getHelperStackDeploymentId());
+        String stackId = System.getProperty("com.atlassian.migration.migrationStack.id", migrationService.getCurrentContext().getHelperStackDeploymentId());
```
nanux, line 104: We can provide `com.atlassian.migration.migrationStack.id` sys property to override the stack ID.

theghostwhoforks: Perhaps we could call it out as an override parameter?

`com.atlassian.migration.migrationStack.id_override` or something like that?

bpar476:
Should there be a follow-up task to surface if there are partial errors during the restore to the user and direct them to the S3 object?

nanux:
> Should there be a follow-up task to surface if there are partial errors during the restore to the user and direct them to the S3 object?

🤦 sorry I forgot to mention I am working on https://aws-partner.atlassian.net/browse/CHET-355 I have implementation working but I don't like the design when I hook it up to frontend. I will open draft PR.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
nanux: I think the `com.atlassian.migration.migrationStack.id` is consistent with other overrides in this file.

PR #184, opened 2020-05-06
CHET-354: error page - Fetch error reason from server

bpar476:
## Summary

* Fix bug where cloudformation status would be failed but state machine wouldn't have transitioned to error just yet.
* Implement loading of error from migration summary in error page
* Restructure error page so unnecessary API calls are not made
* Move validation page to be alongside other app componentes


core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/CloudformationDeploymentService.java
```diff
@@ -77,37 +78,48 @@ protected void deployCloudformationStack(String templateUrl, String stackName, M
 
     protected InfrastructureDeploymentStatus getDeploymentStatus(String stackName) {
         requireNonNull(stackName);
-        return cfnApi.getStatus(stackName);
+        InfrastructureDeploymentStatus status = cfnApi.getStatus(stackName);
```
theghostwhoforks, line 81: This will only happen when the next run of the scheduler is yet to happen, but the API call to getStatus happens before the ticker right?

bpar476: Correct :)

bpar476:
> LGTM - I'm completely sold on having the watcher state in the service, but good to merge for now!

Yeah was just a quick fix. I'm sure there's a better way.

PR #185, opened 2020-05-07
CHET-360: Force lower-case only stack name for now as uppercase interferes with S3 bucket generation.

tarka:
I did look at lower-casing the stack name in cloudformation, but it involves deploying lambda functions and a macro into a separate stack; this is the simplest solution for the moment.


frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx
```diff
@@ -64,7 +64,9 @@ const ButtonRow = styled.div`
 
 const StackNameField = (): ReactElement => {
     const fieldNameValidator = (stackName: string): string => {
-        const regExpMatch = stackName.match('^[a-zA-Z][a-zA-Z0-9-.]{1,126}$');
+        // NOTE: This gets converted to an S3 bucket name (with '-migration'
+        // appended), so must also conform with bucket naming rules.
+        const regExpMatch = stackName.match('^[a-z][a-z0-9-.]{1,53}$');
```
theghostwhoforks, line 69: Nice - 53+10 = 63, the upper limit for bucket name constraints :)

https://docs.aws.amazon.com/AmazonS3/latest/dev/BucketRestrictions.html

PR #187, opened 2020-05-07
CHET-358: make keypair field required

bpar476:
If value is not provided validation text appears
image

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/quickstartToAtlaskit.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/quickstartToAtlaskit.tsx
@@ -264,10 +264,11 @@ const createKeyPairNameField: FormElementGenerator = (defaultFieldProps, param) 
         paramProperties: { Description, ConstraintDescription },
     } = param;
 
-    const validateFun = (val: string): string | void => {
+    const validateFun = (val: string): string => {
         if (val.length === 0) {
             return 'TOO_SHORT';
         }
+        return undefined;
     };
 
     return (
--- a/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/quickstartToAtlaskit.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/quickstartToAtlaskit.tsx
@@ -261,16 +261,22 @@ const createSelectFromQuickstartParam: FormElementGenerator = (defaultFieldProps
 
 const createKeyPairNameField: FormElementGenerator = (defaultFieldProps, param) => {
     const {
-        paramProperties: { Description },
+        paramProperties: { Description, ConstraintDescription },
     } = param;
 
+    const validateFun = (val: string): string | void => {
+        if (val.length === 0) {
+            return 'TOO_SHORT';
+        }
+    };
+
     return (
-        <Field isRequired {...defaultFieldProps}>
+        <Field validate={validateFun} {...defaultFieldProps}>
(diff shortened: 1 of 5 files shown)


dylanrathbone:
Minor but I think the wording for the bubble, `Please fill out this field` doesnt sound right. Maybe something like `Please provide a value for this field` would read better....

bpar476:
> Minor but I think the wording for the bubble, `Please fill out this field` doesnt sound right. Maybe something like `Please provide a value for this field` would read better....

I'm not sure how to override the text 😅

dylanrathbone:
> > Minor but I think the wording for the bubble, `Please fill out this field` doesnt sound right. Maybe something like `Please provide a value for this field` would read better....
> 
> I'm not sure how to override the text 😅

Ahh :D In that case someone must have thought about the wording longer than me

bpar476:
> > > Minor but I think the wording for the bubble, `Please fill out this field` doesnt sound right. Maybe something like `Please provide a value for this field` would read better....
> > 
> > 
> > I'm not sure how to override the text 😅
> 
> Ahh :D In that case someone must have thought about the wording longer than me

I'm sure there is a way, but I can't find it 😡

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/quickstartToAtlaskit.tsx
```diff
@@ -247,6 +247,24 @@ const createSelectFromQuickstartParam: FormElementGenerator = (defaultFieldProps
     );
 };
 
+const createKeyPairNameField: FormElementGenerator = (defaultFieldProps, param) => {
+    const {
+        paramProperties: { Description },
+    } = param;
+
+    return (
+        <Field isRequired {...defaultFieldProps}>
```
nanux, line 256: `isRequired` is the main change? From curiosity is there a way how to convert this to a prop that would then render it into the field?

nanux:
Updated the branch so it runs the builds overnight.

hannasazonava:
@bpar476 I have a couple of questions, pls ignore those that are design nitpicking, if it's too early for design review. 

- why this font differs from the ADG font?

- for the fields' validation we have a pattern to show a message https://atlaskit.atlassian.com/packages/design-system/form/example/submission-validation
which looks similar to this implementation, but I think in our case we should be aiming at complex validation https://atlaskit.atlassian.com/packages/design-system/form/example/complex-validation
Can we support complex validation with this form? 

- We have a step after AWS account connection where we ask for ASI - https://zpl.io/VxZdEP3
Do we need to ask ASI twice?

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/quickstartToAtlaskit.tsx
bpar476: Not quite sure what you mean 🤔

bpar476:
Hey @hannasazonava ,

> why this font differs from the ADG font?

The font is only different because it is running outside of Jira, the font will be the normal font you're used to seeing in the plugin :)

> for the fields' validation we have a pattern to show a message https://atlaskit.atlassian.com/packages/design-system/form/example/submission-validation
which looks similar to this implementation, but I think in our case we should be aiming at complex validation https://atlaskit.atlassian.com/packages/design-system/form/example/complex-validation
Can we support complex validation with this form?

I can set up the "complex" validation appearance so it's got the red text :+1:

> We have a step after AWS account connection where we ask for ASI - https://zpl.io/VxZdEP3
Do we need to ask ASI twice?

This is in the deployment step so it would be after they've already decided if they need an ASI :)

PR #188, opened 2020-05-07
CHET-366: wire up 'DONE' state of db transfer to complete in transferpage

bpar476:
Now once the db migration is complete we show the correct UI.
image


frontend/src/main/dc-migration-assistant-fe/api/db.ts
```diff
@@ -22,12 +22,12 @@ export const dbStatusReportEndpoint = `${dbAPIBase}/report`;
 export const dbStartEndpoint = `${dbAPIBase}/start`;
 
 export enum DBMigrationStatus {
-    NOT_STARTED,
-    EXPORTING,
-    UPLOADING,
-    IMPORTING,
-    DONE,
-    FAILED,
+    NOT_STARTED = 'NOT_STARTED',
```
nanux, line 25: Just for my benefit, why we need to map these to strings, so we can use `===`?

frontend/src/main/resources/i18n/dc-migration-assistant.properties
```diff
@@ -97,6 +97,7 @@ atlassian.migration.datacenter.db.status.exporting=Exporting
 atlassian.migration.datacenter.db.status.uploading=Uploading
 atlassian.migration.datacenter.db.status.done=Complete
 atlassian.migration.datacenter.db.status.unknown=Unknown Status
+atlassian.migration.datacenter.db.completeMessage=Database has been successfully migrated
```
nanux, line 100: Almost feels it deserves 🎉 emoji 😄

frontend/src/main/dc-migration-assistant-fe/api/db.ts
bpar476: Correct 👍

PR #189, opened 2020-05-07
CHET-355: Add SSM url to the output logs

nanux:
* Create `/migration/db/logs `GET` REST endpoint to fetch the end result when we have finished the DB restore. Example responses:
```
{
    "consoleUrl": "https://console.aws.amazon.com/s3/buckets/atl-migrationbucket-abrokes-migration-helper/trebuchet-ssm-document-logs/6cc3f383-a916-4fdd-b23b-6b614e642944/i-08f9fa6be33226139/awsrunShellScript/restoreDatabaseBackupToRDS/",
    "errorMessage": ""
}
```
* Display `SectionMessage` when there was an error in the DB migration and provide link to reach the AWS S3 Console with the `stdout` and `stderr` files.
* We also get first 8000 characters of the error but to keep the designs simple we've decided to not show it in the interface.

image

### DIFF
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationEndpoint.kt
@@ -115,7 +115,7 @@ class DatabaseMigrationEndpoint(
     @Path("/logs")
     fun getCommandOutputs(): Response {
         return try {
-            Response.ok(ssmPsqlDatabaseRestoreService.fetchCommandLogs()).build()
+            Response.ok(ssmPsqlDatabaseRestoreService.fetchCommandResult()).build()
         } catch (e: SsmPsqlDatabaseRestoreService.SsmCommandNotInitialisedException) {
             return Response.status(Response.Status.CONFLICT).entity(mapOf("error" to "SSM command wasn't executed"))
                 .build()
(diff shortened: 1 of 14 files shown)


api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationEndpoint.kt
```diff
@@ -107,4 +109,16 @@ class DatabaseMigrationEndpoint(
                 .build()
         }
     }
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    @Path("/logs")
```
nanux, line 115: Now I think this should be part of the GET`/migration/db` endpoint (part of `DatabaseMigrationStatus`)

frontend/src/main/dc-migration-assistant-fe/components/db/DatabaseMigration.tsx
```diff
@@ -67,6 +71,7 @@ const props: MigrationTransferProps = {
     inProgressStages: dbMigrationInProgressStages,
     getProgress: getProgressFromStatus,
     nextRoute: validationPath,
+    getLogs: fetchDBMigrationLogs,
```
nanux, line 74: Should we introduce `getDetails` Promise prop that can be provided by any migration page (db/fs/provisioning) to display additional details after the finish?

bpar476: I think that's probably a bit overkill for now. If it becomes necessary we can do such a thing.

bpar476:
Looks good, however, does it need some content that will inform the user they should check this for any critical error? If the screen doesn't show an error message then I probably wouldn't bother looking at the logs.

PR #194, opened 2020-05-11
CHET-209: Make reset API delete migrations irrespective of stage

theghostwhoforks:
Remove all migrations w/o considering the current stage of the migration

### DIFF
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
@@ -94,14 +94,8 @@ class MigrationEndpoint(private val migrationService: MigrationService) {
     @Path("/reset")
     @Produces(MediaType.APPLICATION_JSON)
     fun resetMigration(): Response {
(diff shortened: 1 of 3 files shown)


api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
```diff
@@ -94,14 +94,8 @@ class MigrationEndpoint(private val migrationService: MigrationService) {
...
     fun resetMigration(): Response {
-        val currentStage = migrationService.currentStage
-        if (currentStage == MigrationStage.ERROR) {
-            migrationService.deleteMigrations()
-            return Response.ok().build()
-        }
-        return Response.status(Response.Status.CONFLICT)
-                .entity(mapOf("reason" to "Cannot reset migration when current stage is $currentStage"))
-                .build()
+        migrationService.deleteMigrations()
```
theghostwhoforks, line 97: At a later time, we need a controller advice here to handle exceptions that AO could raise and handle it by responding with the appropriate error code.

Currently, if there's an AO error, a XML response text is rendered with a 500 error message

PR #196, opened 2020-05-11
no-issue: remove develop endpoint for reset migrations as the regular…

theghostwhoforks:
… endpoint will work just as well

The /reset migration endpoint does the same thing - removing the duplicate and updating the usages

### DIFF
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
@@ -98,7 +98,6 @@ class MigrationEndpoint(private val migrationService: MigrationService) {
         return Response.ok().build()
     }
 
-
     private fun migrationContextResponseEntity(): Map<String, String> {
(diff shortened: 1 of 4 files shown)


bpar476:
> This isn't the same thing. The regular reset is only allowed in an error state, so doesn't work for testing, where you want to reset the migration state unconditionally.
> One alternative would be to allow non-error resets when the dev profile is enabled, but that would mean that the rest endpoint semantics would be contextual, which isn't ideal. Having an explicit testing endpoint is clearer and doesn't complicate the main code path.

bpar476:
oops

bpar476:
> This isn't the same thing. The regular reset is only allowed in an error state, so doesn't work for testing, where you want to reset the migration state unconditionally.
> One alternative would be to allow non-error resets when the dev profile is enabled, but that would mean that the rest endpoint semantics would be contextual, which isn't ideal. Having an explicit testing endpoint is clearer and doesn't complicate the main code path.

Recently changed so that we can "cancel" a migration. #194

PR #197, opened 2020-05-11
CHET-378: Prerequisite checks

tarka:
Kind of ugly, but the basic functionality is there.

### DIFF
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
@@ -23,6 +23,7 @@ import com.atlassian.migration.datacenter.spi.exceptions.MigrationAlreadyExistsE
 import javax.ws.rs.*
 import javax.ws.rs.core.MediaType
 import javax.ws.rs.core.Response
+import com.fasterxml.jackson.module.kotlin.jacksonObjectMapper
 
 /**
  * REST API Endpoint for managing in-product DC migrations.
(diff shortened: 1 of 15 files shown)


api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
```diff
@@ -90,6 +91,17 @@ class MigrationEndpoint(private val migrationService: MigrationService) {
         }
     }
 
+    @Produces(MediaType.APPLICATION_JSON)
+    @Consumes(MediaType.APPLICATION_JSON)
+    @GET
+    @Path("/ready")
+    fun getMigrationReadyStatus(): Response {
+        val status = migrationService.readyStatus;
+        return Response
```
theghostwhoforks, line 100: do we need to use jackson mapper explicitly here? maps are serialized into a JSON string and with a few annotations the status object could be serialized as well?

spi/src/main/java/com/atlassian/migration/datacenter/spi/MigrationService.java
```diff
@@ -73,6 +73,13 @@
      */
     void transition(MigrationStage to) throws InvalidMigrationStageError;
 
+    /**
+     * Check whether migration prerequisites met.
+     *
+     * @return whether migration prerequisites met
+     */
+    MigrationReadyStatus getReadyStatus();
```
theghostwhoforks, line 81: MigrationPrerequisites sounds better I think?

api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
tarka: Tried annotating the class, but it didn't like it for some reason. I decided to do the simplest thing rather than disappear down the serialisation/annotation rabbit-hole.

theghostwhoforks: Yeah - it's been pretty finicky

PR #198, opened 2020-05-11
CHET-394: listen to issue created event

bpar476:
## Summary

* Create interface for capturing attachments
* Register an issue event listener
* get **_all_** attachments on the issue
* Capture them with a dummy attachment capturer

### Blockers

the `filename` on the issue is the actual filename of the attachment, not the path to the attachment on the shared home.

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/fs/capture/JiraIssueAttachmentListener.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/fs/capture/JiraIssueAttachmentListener.java
@@ -48,6 +48,7 @@ public class JiraIssueAttachmentListener implements InitializingBean, Disposable
 
     @EventListener
     public void onIssueEvent(IssueEvent issueEvent) {
+        logger.trace("received jira event with type {}", issueEvent.getEventTypeId());
         if (issueEvent.getEventTypeId().equals(EventType.ISSUE_CREATED_ID)) {
             logger.trace("got issue created event");
             issueEvent.getIssue().getAttachments().forEach(
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/fs/capture/JiraIssueAttachmentListener.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/fs/capture/JiraIssueAttachmentListener.java
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2020 Atlassian
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package com.atlassian.migration.datacenter.core.fs.capture;
+
+
+import com.atlassian.event.api.EventListener;
+import com.atlassian.event.api.EventPublisher;
+import com.atlassian.jira.event.issue.IssueEvent;
+
+public class JiraIssueAttachmentListener {
+
+    AttachmentCapturer attachmentCapturer;
+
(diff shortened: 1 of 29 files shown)


core/src/main/java/com/atlassian/migration/datacenter/core/fs/capture/JiraIssueAttachmentListener.java
```diff
@@ -0,0 +1,65 @@
...
+
+    public JiraIssueAttachmentListener(EventPublisher eventPublisher, AttachmentCapturer attachmentCapturer) {
+        this.eventPublisher = eventPublisher;
+        this.attachmentCapturer = attachmentCapturer;
+    }
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
+        logger.info("registering with event publisher");
+        eventPublisher.register(this);
```
theghostwhoforks, line 46: Why do we need to register w/ the event publisher here?

theghostwhoforks:
While this is OK for now, the core module should not include the `jira-api` dependency

This will make it difficult to extend this to other products.

I'll carve this out into the plugin, where this should belong.

core/src/main/java/com/atlassian/migration/datacenter/core/fs/capture/JiraIssueAttachmentListener.java
```diff
@@ -0,0 +1,68 @@
...
+    @EventListener
+    public void onIssueEvent(IssueEvent issueEvent) {
+        logger.trace("received jira event with type {}", issueEvent.getEventTypeId());
+        if (issueEvent.getEventTypeId().equals(EventType.ISSUE_CREATED_ID)) {
+            logger.trace("got issue created event");
+            issueEvent
+                    .getIssue()
+                    .getAttachments()
+                    .stream()
+                    .map(x -> Paths.get(x.getFilename()))
```
nanux, line 58: We might get the full path from [AttachmentStore.getAttachmentFile](https://docs.atlassian.com/software/jira/docs/api/7.0.5/com/atlassian/jira/issue/attachment/store/DefaultFileSystemAttachmentStore.html#getAttachmentFile-com.atlassian.jira.issue.attachment.AttachmentKey-) and pass the Id from the attachment. I am not sure if the Key is different than Id because I don't see `Attachment.getKey()`

Also there is: https://docs.atlassian.com/software/jira/docs/api/7.0.5/com/atlassian/jira/util/AttachmentUtils.html

jira-plugin/src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
```diff
@@ -257,4 +261,14 @@ public QuickstartDeploymentService quickstartDeploymentService(CfnApi cfnApi, Mi
     public AWSMigrationHelperDeploymentService awsMigrationHelperDeploymentService(CfnApi cfnApi, MigrationService migrationService, Supplier<AutoScalingClient> autoScalingClientFactory) {
         return new AWSMigrationHelperDeploymentService(cfnApi, autoScalingClientFactory, migrationService);
     }
+
+    @Bean
+    public AttachmentCaptor attachmentCapturer() {
```
nanux: Non-native speaker here but does it make sense to call it `attachmentCaptor` to keep it consistent?

theghostwhoforks: Just pushed this in 👍

core/src/main/java/com/atlassian/migration/datacenter/core/fs/capture/JiraIssueAttachmentListener.java
theghostwhoforks: Yeah - this is slightly confusing as the attachment store/utils are all deprecated in favour of attachment manager
```
/**
 * Static utilities for working with the attachment files and their directories.
 *
 * @deprecated Use {@link com.atlassian.jira.issue.AttachmentManager} only. Since v6.1
 */
```

There's a separate story that I'm working on to build out something that can get attachment path or content that we can then upload to S3.

bpar476:
> While this is OK for now, the core module should not include the `jira-api` dependency
> 
> This will make it difficult to extend this to other products.
> 
> I'll carve this out into the plugin, where this should belong.

We already include the JiraHome API. I don't think we should add additional indirection to the plugin until it's necessary. It just makes the code base more difficult to grok.

PR #200, opened 2020-05-12
CHET-402: Update the Home design

nanux:
* Formatted the landing page closer to our design specification
* Added lozenges for the requirements
* Added spinner (instead of `...`) when we are fetching the status for each requirement
* Added same formatting for the Auth page
* Updated the title to `Jira Data Center Migration App`

The landing page
image

Spinners while loading
image

Auth page formatting
image

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/Home.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/Home.tsx
@@ -16,10 +16,12 @@
 
 import React, { ReactElement, useEffect, useState, FunctionComponent } from 'react';
 import Button from '@atlaskit/button';
-import InlineMessage from '@atlaskit/inline-message';
 import styled from 'styled-components';
 import { Redirect } from 'react-router-dom';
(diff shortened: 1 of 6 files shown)


frontend/src/main/dc-migration-assistant-fe/components/Home.tsx
```diff
@@ -84,7 +89,7 @@ const MigrationActionSection: FunctionComponent<ActionSectionProps> = ({ startBu
             <p>
                 <ReadyStatus />
             </p>
-            <ButtonContainer>
+            <MigrationsContainer>
                 <Button
```
theghostwhoforks, line 93: Shouldn't this button be disabled when the prerequisites are not met?

nanux: I was thinking the same and had implementation locally but then I've backed off in case we might have problems when detecting the prerequisites and effectively block people from migration. I am happy to bring it in (in the next PR) if we want that - I don't think we've just decided 100% it should be there.

PR #203, opened 2020-05-13
CHET-397: Integrate attachment listener with S3 FS migration

bpar476:
### Summary

* Implement and test starting the Attachment listener
* Break out bulk S3 copy from chungus `S3FileSystemMigration` class 
* Hide functionality behind spring profile


core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -115,29 +93,20 @@ public Boolean scheduleMigration() throws InvalidMigrationStageError {
...
     @Override
     public void startMigration() throws InvalidMigrationStageError {
-        logger.trace("Beginning migration. Uploading shared home dir {} to S3 bucket {}", getSharedHomeDir(),
-                getS3Bucket());
         if (isRunning()) {
             logger.warn("Filesystem migration is currently in progress, aborting new execution.");
             return;
         }
 
-        s3AsyncClient = this.s3AsyncClientSupplier.get();
```
bpar476, line 125: Adding the attachement listener to this class was just another responsibility for this already large class so I moved the initial bulk copy logic to a separate class which should maket this one a bit more maintainable.

core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -115,29 +97,25 @@ public Boolean scheduleMigration() throws InvalidMigrationStageError {
...
-        report = new DefaultFileSystemMigrationReport();
-
         migrationService.transition(MigrationStage.FS_MIGRATION_COPY_WAIT);
-        report.setStatus(UPLOADING);
-
-        Crawler homeCrawler = new DirectoryStreamCrawler(report);
 
-        S3UploadConfig s3UploadConfig = new S3UploadConfig(getS3Bucket(), s3AsyncClient, getSharedHomeDir());
-        Uploader s3Uploader = new S3Uploader(s3UploadConfig, report);
+        for (String profile : environment.getActiveProfiles()) {
```
bpar476, line 107: This is our "feature flag" mechanism for now

core/src/main/java/com/atlassian/migration/datacenter/core/fs/copy/S3BulkCopy.java
```diff
@@ -0,0 +1,105 @@
...
+import org.slf4j.LoggerFactory;
+import software.amazon.awssdk.services.s3.S3AsyncClient;
+
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.util.function.Supplier;
+
+import static com.atlassian.migration.datacenter.spi.fs.reporting.FilesystemMigrationStatus.UPLOADING;
+
+public class S3BulkCopy {
```
theghostwhoforks, line 39: This is good. 

I assume we will need to modify this to get the incremental copy to work with the deferred sync, correct?

bpar476: I think the final sync would be completed by a different class as there is no "directory crawling" required in that instance.

theghostwhoforks: Yup - I just wanted to get an idea of how you are thinking abut this 👍

PR #204, opened 2020-05-13
CHET-399: render captured files in offline warning page

bpar476:
## Summary

* Implement component for showing list of files which were captured
* Render component in warning page

image

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/api/fs.ts
+++ b/frontend/src/main/dc-migration-assistant-fe/api/fs.ts
@@ -46,35 +46,33 @@ type FileSystemMigrationStartResponse = {
 
 export const fs = {
     getFsMigrationStatus: async (): Promise<FileSystemMigrationStatusResponse> => {
-        return callAppRest('GET', RestApiPathConstants.fsStatusRestPath).then(result =>
-            result.json()
-        );
+        const result = await callAppRest('GET', RestApiPathConstants.fsStatusRestPath);
+        return result.json();
     },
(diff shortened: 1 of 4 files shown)


bpar476:
Note that this UI is not end-product, just an interim solution to help demonstrate the file listener.

frontend/src/main/dc-migration-assistant-fe/api/fs.ts
```diff
@@ -45,14 +45,14 @@ type FileSystemMigrationStartResponse = {
 };
 
 export const fs = {
-    getFsMigrationStatus: (): Promise<FileSystemMigrationStatusResponse> => {
+    getFsMigrationStatus: async (): Promise<FileSystemMigrationStatusResponse> => {
```
nanux, line 48: Why we have sometimes `async` for our `Promises` and sometimes we don't? (e.g. https://github.com/atlassian-labs/dc-migration-assistant/blob/09d47b42d002d79a5694f3ebae1d359f42c7cccc/frontend/src/main/dc-migration-assistant-fe/components/db/DatabaseMigration.tsx#L62)

Is that a good practice to add `async` to all Promises? I thought `async` is useful when you want to run the code with `await` and avoid the necessary chaining with for example `promise.then().finally()`.

bpar476: The typescript server was giving me a warning that the function "could be async". So I suppose it's enouraged?

PR #207, opened 2020-05-14
Chet 395 issue modified event

theghostwhoforks:
Depends on #208 to be merged in first.

Once done, this PR adds the ability to listen to multiple jira issue events and fixes the broken AttachmentStore dependency on `Promise` class from `com.atlassian.concurrent.util` library


core/src/main/java/com/atlassian/migration/datacenter/core/fs/captor/DefaultAttachmentCaptor.java
```diff
@@ -0,0 +1,45 @@
...
+    private final AttachmentStore attachmentStore;
+
+    public DefaultAttachmentCaptor(AttachmentStore attachmentStore) {
+        this.attachmentStore = attachmentStore;
+    }
+
+    private static final Logger logger = LoggerFactory.getLogger(DefaultAttachmentCaptor.class);
+
+    @Override
+    public void captureAttachment(Attachment attachment) {
```
theghostwhoforks, line 37: This refactoring from AttachmentPathCaptor to AttachmentCaptor was made as part of this PR.


This bloats the PR a bit, but I was testing the AttachmentStore and did this small fix anyway

core/src/main/java/com/atlassian/migration/datacenter/core/fs/listener/JiraIssueAttachmentListener.java
```diff
@@ -14,35 +14,33 @@
...
 
-import java.io.File;
-import java.util.Collection;
+import java.util.Arrays;
+import java.util.List;
 
 public class JiraIssueAttachmentListener implements InitializingBean, DisposableBean {
 
     private static final Logger logger = LoggerFactory.getLogger(JiraIssueAttachmentListener.class);
+    private static final List<Long> ISSUE_EVENT_TYPES_TO_LISTEN = Arrays.asList(EventType.ISSUE_CREATED_ID, EventType.ISSUE_UPDATED_ID);
```
theghostwhoforks, line 36: These 2 events seemed like what we need to listen to.

See all jira issue events here - 
Screenshot 2020-05-13 at 14 29 26

theghostwhoforks: Do we need to listen to ISSUE_MOVED_EVENT perhaps 🤔 ?

core/pom.xml
```diff
@@ -74,6 +74,18 @@
             <groupId>com.atlassian.jira</groupId>
             <artifactId>jira-api</artifactId>
         </dependency>
+        <!-- TODO: Verify if scope is valid in 7.13 -->
```
theghostwhoforks, line 77: This will not be a part of this PR once #208  gets merged in

pom.xml
```diff
@@ -230,6 +230,16 @@
                 <version>7.13.12</version>
                 <scope>provided</scope>
             </dependency>
+            <dependency>
```
theghostwhoforks, line 233: This will not be a part of this PR once #208 gets merged in

core/src/main/java/com/atlassian/migration/datacenter/core/fs/captor/AttachmentCaptor.java
```diff
@@ -16,9 +16,9 @@
 
 package com.atlassian.migration.datacenter.core.fs.captor;
 
-import java.nio.file.Path;
+import com.atlassian.jira.issue.attachment.Attachment;
 
 @FunctionalInterface
-public interface AttachmentPathCaptor {
-    void captureAttachmentPath(Path attachmentPath);
+public interface AttachmentCaptor {
```
bpar476, line 22: Why change the interface to capture `Attachments` and not their paths?

core/src/main/java/com/atlassian/migration/datacenter/core/fs/listener/JiraIssueAttachmentListener.java
bpar476: I believe the attachment path stays the same even when the issue is moved

bpar476: We should double-check this as a bug task

bpar476:
#209 Is my PR which will have conflicts, for reference.

theghostwhoforks:
> If the motivation is to reduce the number of things the listener is responsible for then I suggest we have something inbetween the `AttachmentPathCaptor` and the `JiraIssueListener` which deals with splitting the attachment and its thumbnails.

Yeah - that was the main intention and it's good to have an attachment captor that deals with the internals of dealing with how an attachment and thumbnail paths are processed.

I've spoken with @bpar476 offline and we will proceed with retaining the attachment path captor approach and may follow with a different PR with the above changes laterr.

theghostwhoforks:
ready to be merged now

core/src/test/java/com/atlassian/migration/datacenter/core/fs/listener/JiraIssueAttachmentListenerTest.java
```diff
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
-package com.atlassian.migration.datacenter.core.fs.captor;
+package com.atlassian.migration.datacenter.core.fs.listener;
```
theghostwhoforks, line 17: The benefit of a decoupled listener is is visible when we look at the setup in this test. The more cumbersome a test is to write, it's a sign that the class is a bit complex. 

An attachment captor should just capture an attachment. What the captor does with that attachment is just an internal responsibility. The side effect is that the listener and the captor have a single responsibility

core/pom.xml
```diff
@@ -76,6 +76,15 @@
             <groupId>com.atlassian.jira</groupId>
             <artifactId>jira-api</artifactId>
         </dependency>
+        <dependency>
```
bpar476, line 79: Are these not a duplicate of the dependencies below?

pom.xml
```diff
@@ -243,6 +243,16 @@
                 <artifactId>atlassian-util-concurrent</artifactId>
                 <version>3.0.0</version>
             </dependency>
+            <dependency>
+                <groupId>com.atlassian.util.concurrent</groupId>
```
bpar476, line 247: This looks duplicated too

core/src/main/java/com/atlassian/migration/datacenter/core/fs/captor/AttachmentPathCaptor.java
```diff
@@ -16,9 +16,9 @@
 
 package com.atlassian.migration.datacenter.core.fs.captor;
 
+
 import java.nio.file.Path;
 
-@FunctionalInterface
```
bpar476, line 21: Why remove the functional interface?

theghostwhoforks: done

core/pom.xml
theghostwhoforks: fixed 🤦

PR #208, opened 2020-05-14
Chet 408 fix supplier imports

theghostwhoforks:
Use java util supplier, as used by `LazyReference` class from `io.atlassian.concurrent` libraries

### DIFF
--- a/core/pom.xml
+++ b/core/pom.xml
@@ -70,10 +70,22 @@
             <groupId>com.atlassian.scheduler</groupId>
             <artifactId>atlassian-scheduler-api</artifactId>
         </dependency>
+
+        <!-- Jira dependencies will have to be moved to jira-plugin when we start supporting multiple products -->
         <dependency>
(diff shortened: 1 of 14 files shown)


theghostwhoforks:
I've verified that the OSGI integration works as expected with this chagne

bpar476:
Does the disabled test passs?

nanux:
> I've verified that the OSGI integration works as expected with this chagne

Was there any benefit by using the Atlassian package?

theghostwhoforks:
> Does the disabled test passs?

it does - this is done in the other PR #207

theghostwhoforks:
> > I've verified that the OSGI integration works as expected with this chagne
> 
> Was there any benefit by using the Atlassian package?

Nothing as far as I could tell. I think the atlassian package had the function definition before it landed in java 8.

io.atlassian.util library defers to using the java definition and not the homegrown one

PR #209, opened 2020-05-14
CHET-396: persist listened files

bpar476:
### Summary

* Implement DTO class for capturing paths "listened to" during the FS copy
* Associate these paths with the current migration
* Implement and test, storing and fetching these paths

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/fs/captor/DefaultAttachmentPathCaptor.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/fs/captor/DefaultAttachmentPathCaptor.java
@@ -16,6 +16,7 @@
 
 package com.atlassian.migration.datacenter.core.fs.captor;
 
+import com.atlassian.activeobjects.external.ActiveObjects;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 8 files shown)


core/src/main/java/com/atlassian/migration/datacenter/core/fs/captor/DefaultAttachmentPathCaptor.java
```diff
@@ -16,17 +16,58 @@
...
+    }
 
     @Override
     public void captureAttachmentPath(Path attachmentPath) {
         logger.debug("captured attachment for final sync: {}", attachmentPath.toString());
+
+        FileSyncRecord record = ao.create(FileSyncRecord.class);
+
+        record.setFilePath(attachmentPath.toString());
+        // FIXME: Should we cache the migration identifier so we don't hit the DB twice every time we capture a path?
```
theghostwhoforks, line 50: May not be required as the ao cache should serve this up. caching would mean saving state somewhere, which is not necessary right now, IMO

core/src/main/java/com/atlassian/migration/datacenter/core/fs/captor/DefaultAttachmentPathCaptor.java
```diff
@@ -16,17 +16,58 @@
...
+        record.setMigration(migrationService.getCurrentMigration());
+
+        record.save();
+    }
+
+    @Override
+    public Set<FileSyncRecord> getCapturedAttachments() {
```
theghostwhoforks, line 57: I don't think this belongs to an `AttachmentCaptor` as this is more a service to retrieve saved attachments. This should probably belong to a different interface, but this is OK to be merged in as is. 

May require a look into later

bpar476: This is a different interface, `AttachmentSyncManager` but it composes the AttachmentCaptor. I probably could have updated the name though, will do in a follow up PR.

PR #210, opened 2020-05-14
Chet 398 final sync files endpoint

bpar476:
### Summary
* Implement API for getting files discovered during final sync
* Wire up frontend to ⬆️ API

#### Note

Will update with upstream once #203 is merged


frontend/src/main/dc-migration-assistant-fe/api/fs.ts
```diff
@@ -72,7 +77,9 @@ export const fs = {
         return Promise.reject(JSON.stringify(result));
     },
 
-    getCapturedFiles: (): Promise<Array<string>> => {
-        return Promise.resolve(['/home/benpar/test', '/home/benpar/fake']);
+    getCapturedFiles: async (): Promise<Array<string>> => {
```
theghostwhoforks, line 80: This is not used anywhere right now, correct?

bpar476: It's used in the `WarningStage` page

PR #212, opened 2020-05-15
Transfer page and fs listener bugfixes

bpar476:
* Fix some null dereferences in transfer page
* Drop thumbnailable filter from issue listener - it was causing NPE's and missing the thumbnails

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/fs/listener/JiraIssueAttachmentListener.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/fs/listener/JiraIssueAttachmentListener.java
@@ -67,7 +67,7 @@ public class JiraIssueAttachmentListener implements DisposableBean {
 
             attachments
                     .stream()
-                    // FIXME: I create NPE's...
+                    // FIXME: I lie about thumbnailable
                     //.filter(Attachment::isThumbnailable)
                     .map(attachmentStore::getThumbnailFile)
                     .filter(Objects::nonNull)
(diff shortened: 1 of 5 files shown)


bpar476:
Created this Bug to track the `thumbNailable` API issue https://aws-partner.atlassian.net/browse/CHET-414

core/src/main/java/com/atlassian/migration/datacenter/core/fs/listener/JiraIssueAttachmentListener.java
```diff
@@ -66,8 +67,10 @@ public void onIssueEvent(IssueEvent issueEvent) {
 
             attachments
                     .stream()
-                    .filter(Attachment::isThumbnailable)
+                    // FIXME: I lie about thumbnailable
+                    //.filter(Attachment::isThumbnailable)
                     .map(attachmentStore::getThumbnailFile)
+                    .filter(Objects::nonNull)
                     .map(File::toPath)
```
theghostwhoforks, line 74: Do we need to add a filter(Files::exists) here in the chain? i.e. the case when DB says there's a thumbnail, but the FS does not contain the file?

bpar476: That's a good idea. I'll create a bug for it :)

bpar476: https://aws-partner.atlassian.net/browse/CHET-415

PR #213, opened 2020-05-18
Chet 408 jira 713

theghostwhoforks:
WIP - pom and amps and spring fixes


core/pom.xml
```diff
@@ -140,13 +149,85 @@
         <dependency>
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-xml</artifactId>
-            <version>${jackson.version}</version>
         </dependency>
+
         <dependency>
-            <groupId>com.fasterxml.jackson.module</groupId>
```
bpar476, line 146: Did jackson end up being unused in `core`?

jira-plugin/pom.xml
```diff
@@ -104,13 +124,21 @@
                             *
                         </Import-Package>
                     </instructions>
+                    <products>
```
bpar476, line 127: What does this do?

bpar476:
It looks like a few dependencies are duplicated across the root pom and the module poms. Is this intentional?

core/pom.xml
theghostwhoforks: As far as I could tell, no

jira-plugin/pom.xml
theghostwhoforks: Amps supports multiple products. Having this gave me the option to specify a http.port, which was a bit cleaner and required for testing (I was running Jira on docker and jira via amps to test multiple versions)

theghostwhoforks:
> It looks like a few dependencies are duplicated across the root pom and the module poms. Is this intentional?

There shouldn't be any duplicates - can you please point me to instances where you see this and I'll sort this out?

 I've just defined all the dependencies used across the project in the `<dependenciesManagement>` section (i.e. Group, Artifact & Version) and specified the usages in the leaf POMs with the scope added (no version).

I got rid of a few dependencies that were specified twice during this process.

bpar476:
> > It looks like a few dependencies are duplicated across the root pom and the module poms. Is this intentional?
> 
> There shouldn't be any duplicates - can you please point me to instances where you see this and I'll sort this out?
> 
> I've just defined all the dependencies used across the project in the `<dependenciesManagement>` section (i.e. Group, Artifact & Version) and specified the usages in the leaf POMs with the scope added (no version).
> 
> I got rid of a few dependencies that were specified twice during this process.

`javax.annotation-api`, `localstack-utils`, `activeobjects-test`, `junit-jupiter-api` are duplicated between root and core. There might be some more, that's just from looking for a few minutes.

theghostwhoforks:
> `javax.annotation-api`

These aren't duplicates in that sense. `root` specifies the versions and the `leaf` modules specifies the scope. This follows the convention set by other atlassian-plugins (& multi-module projects in general)

The advantage here is that child/leaf modules specify just what is required by them instead of inheriting dependencies, specified in the parent, that may not be required by them. `localstack` is one such example. `spi` and `api` don't need it, but `core` needs the library.

If we feel like we are repeating test dependencies, we can extract those into a separate test module and import the pom later on.

bpar476:
> > `javax.annotation-api`
> 
> These aren't duplicates in that sense. `root` specifies the versions and the `leaf` modules specifies the scope. This follows the convention set by other atlassian-plugins (& multi-module projects in general)
> 
> The advantage here is that child/leaf modules specify just what is required by them instead of inheriting dependencies, specified in the parent, that may not be required by them. `localstack` is one such example. `spi` and `api` don't need it, but `core` needs the library.
> 
> If we feel like we are repeating test dependencies, we can extract those into a separate test module and import the pom later on.

TIL

PR #214, opened 2020-05-19
INNO: Kotlin conversion

tarka:


pom.xml
```diff
@@ -409,26 +411,36 @@
                     <jvmTarget>1.8</jvmTarget>
                 </configuration>
             </plugin>
+
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.5.1</version>
                 <executions>
+                    <!-- Replacing default-compile as it is treated specially by maven -->
```
theghostwhoforks, line 420: Should this change in the root pom plugin definition too?

tarka: This is the root pom for the project. Am I missing something?

theghostwhoforks: ah nope - I misread

PR #215, opened 2020-05-19
INNO: More Kotlin

tarka:


spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/infrastructure/ProvisioningConfig.kt
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.atlassian.migration.datacenter.spi.infrastructure
+
+class ProvisioningConfig(val templateUrl: String, val stackName: String, params: Map<String, Any>) {
```
theghostwhoforks, line 18: I understand that you are translating java code to kotlin code, but I don't think we need constructor param to be `Map<String,Any>`.

We convert any array option type values to a comma separated list, which is then consumed as is by the CFn SDK. See: https://github.com/atlassian-labs/dc-migration-assistant/blob/master/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx#L250


So, I think we can convert the consturctor to  -

`class ProvisioningConfig(val templateUrl: String, val stackName: String, params: Map<String, String>)`

spi/src/test/kotlin/com/atlassian/migration/datacenter/spi/infrastructure/ProvisioningConfigTest.kt
```diff
@@ -0,0 +1,27 @@
...
+import org.junit.jupiter.api.Assertions
+import org.junit.jupiter.api.Test
+import java.util.*
+
+class ProvisioningConfigTest {
+    @Test
+    fun shouldGetListParametersAsCommaSeparatedValues() {
+        val config = ProvisioningConfig("https://template.url", "stackName", object : HashMap<String, Any>() {
+            init {
+                put("azs", object : ArrayList<String?>() {
```
theghostwhoforks, line 12: To the point above, the UI sends this across as 

`"azs": "us-east-1a, us-east-2a"`

bpar476:
Do you reckon we could keep the java tests until the conversion of the code is complete? I'm just a bit worried that we'll break some existing functionality and not notice if we're also rewriting the tests?

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationService.kt
```diff
@@ -13,87 +13,87 @@
...
      * Creates a new migration in the initial stage. Using this method will create just one migration object in the database
-     * <b>or</b> find the existing migration object and return it.
+     * **or** find the existing migration object and return it.
      *
-     * @throws {@link MigrationAlreadyExistsException} when a migration object already exists.
+     * @throws [MigrationAlreadyExistsException] when a migration object already exists.
      */
-    Migration createMigration() throws MigrationAlreadyExistsException;
+    @Throws(MigrationAlreadyExistsException::class)
+    fun createMigration(): Migration
```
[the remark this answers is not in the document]
tarka: I'm not sure. I'm not sure what is idiomatic Kotlin yet, and this interface is mostly an automatic conversion. I'm going to do a second pass through some of these once the bulk of the conversion is complete to make them more consistent.

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/exceptions/InvalidMigrationStageError.kt
```diff
@@ -0,0 +1,36 @@
...
+package com.atlassian.migration.datacenter.spi.exceptions
+
+import com.atlassian.migration.datacenter.spi.MigrationStage
+import org.apache.commons.lang3.StringUtils
+
+class InvalidMigrationStageError(message: String) : Exception(message) {
+    companion object {
+        fun errorWithMessage(expected: MigrationStage, actual: MigrationStage, prefix: String): InvalidMigrationStageError {
+            var message = String.format("Attempt to migrate from `%s` to `%s`", expected, actual)
+            if (StringUtils.isNotBlank(prefix)) {
```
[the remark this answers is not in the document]
tarka: Good point.

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/infrastructure/ProvisioningConfig.kt
tarka: I don't think Map<String,String> work currently. The value from JSON could potentially be a boolean, integer, string, or list. In particular the list needs to be decoded server-side. I agree we could convert everything to string client-side, but that erases types, so is less flexible down the line.

tarka:
> Do you reckon we could keep the java tests until the conversion of the code is complete? I'm just a bit worried that we'll break some existing functionality and not notice if we're also rewriting the tests?

I run the tests as Java first, then do the conversion.

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/infrastructure/ProvisioningConfig.kt
theghostwhoforks: Sounds good!

PR #216, opened 2020-05-20
INNO: Kotlin conversion

tarka:
This is the rest of the SPI module.

I've also performed some manual testing of the full stack to ensure that the JSON serialisation still works; some related tweaks included.


spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/fs/reporting/FailedFileMigration.kt
```diff
@@ -0,0 +1,30 @@
...
+package com.atlassian.migration.datacenter.spi.fs.reporting
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect
+import java.nio.file.Path
+
+@JsonAutoDetect
+class FailedFileMigration(val filePath: Path, val reason: String){
+
+    override fun equals(obj: Any?): Boolean {
+        if (obj is FailedFileMigration) {
```
theghostwhoforks, line 25: Me being cheeky 😁 

`return obj is FailedFileMigration && filePath == obj.filePath && reason == obj.reason`

core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -77,7 +77,7 @@ public FileSystemMigrationReport getReport() {
     }
 
     @Override
-    public Boolean scheduleMigration() throws InvalidMigrationStageError {
+    public boolean scheduleMigration() throws InvalidMigrationStageError {
```
theghostwhoforks, line 80: Was this change necessary?

tarka: Temporary; it gets tricky if you have a Kotlin interface that returns a non-Nullable, but the Java implementation doesn't know the difference. This is a simple workaround.

PR #218, opened 2020-05-22
INNO: Start of smaller Kotlin conversion PRs

tarka:


core/src/main/kotlin/com/atlassian/migration/datacenter/TemplateServlet.kt
```diff
@@ -0,0 +1,65 @@
...
+            } else {
+                response.sendRedirect(loginUriProvider.getLoginUri(URI.create(request.requestURL.toString())).toASCIIString())
+            }
+        } catch (exception: IOException) {
+            logger.error("Unable to render template", exception)
+            response.sendError(HttpServletResponse.SC_INTERNAL_SERVER_ERROR)
+        }
+    }
+
+    @Throws(IOException::class)
```
bpar476, line 52: Does this need `@Throws` if it's private? My understanding is the annotation is just for java that consumes the method

core/src/test/kotlin/com/atlassian/migration/datacenter/TemplateServletTest.kt
```diff
@@ -0,0 +1,100 @@
...
+import java.net.URI
+import javax.servlet.http.HttpServletRequest
+import javax.servlet.http.HttpServletResponse
+import javax.ws.rs.core.MediaType
+
+@ExtendWith(MockitoExtension::class)
+class TemplateServletTest {
+    private var sut: TemplateServlet? = null
+
+    @Mock
```
bpar476, line 39: The API module uses `MockK` as the mocking library. It might be worthwhile to stay consistent, would get rid of the "\`when\`"` too

core/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
```diff
@@ -0,0 +1 @@
+mock-maker-inline
```
bpar476, line 1: What's this 🤔

bpar476:
I think perhaps we should table the kotlin conversion for now until we can get the GA release out.

core/src/test/resources/mockito-extensions/org.mockito.plugins.MockMaker
tarka: Enables mocking of `final` classes. This simplifies porting of mocked tests to Kotlin.

tarka:
> I think perhaps we should table the kotlin conversion for now until we can get the GA release out.

Yeah, I'm pulling this in favour of incremental porting. I achieved the main thing I wanted (enabling mixed-language Maven modules so we can do more fine-grained porting as needed).

PR #223, opened 2020-05-27
CHET-422: Add backend for ASI existence check

tarka:


core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/AtlassianInfrastructureService.kt
```diff
@@ -0,0 +1,25 @@
...
+import com.atlassian.migration.datacenter.core.aws.CfnApi
+import software.amazon.awssdk.services.cloudformation.model.Stack
+
+/**
+ * Encapsulates some higher-level operations relating to Atlassian-defined VPC stacks (ASIs).
+ */
+class AtlassianInfrastructureService(val cfnApi: CfnApi) {
+
+    fun findASIs() : List<Stack> {
+        val stacks = cfnApi.listStacksFull();
```
theghostwhoforks, line 12: should we just check if the default export is there first and then return that? If the default export (`ATL_`) is not there, then we run the heuristic?

tarka: Well, they might have just added a new ASI called "ATL2-" for this migration (e.g. while testing); I think if the user has created a compatible ASI we should show it to them. However if "ATL-" exists it should definitely be at the top. But that's something that should be done at the UI/UX layer.

theghostwhoforks: The FE layer cannot speak to CFn directly - it has to go via the backend APIs. 

In which case, we would need to add that check here, correct? What am I missing?

tarka: This might just be my personal preference, but the way I normally slice these things is:

The backend will return all available candidate ASIs via the API endpoint (which I'm writing now). So the frontend would receive a list of ASIs, and the decision on how to display those ASIs can be made there. So if we decide that if "ATL-" exists we ignore all other ASIs we can just do that there; it becomes a UX/UI issue rather than something that is baked into the backend.

Or possibly we're talking about completely different things? :)

PR #226, opened 2020-05-27
CHET-323: Scroll to top of page on page transitions

dylanrathbone:
When transitioning from step `Step 2 of 6: Deploy on AWS` by selecting the `Deploy` button the subsequent page is scrolled to the bottom. This fix ensures the subsequent scrolls to the top on load.

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/aws/AwsRoutes.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/aws/AwsRoutes.tsx
@@ -22,23 +22,6 @@ import { QuickStartDeploy } from './quickstart/QuickStartDeploy';
 import { AuthenticateAWS, CredSubmitFun, QueryRegionFun } from './auth/AuthenticateAWS';
 import { callAppRest, RestApiPathConstants } from '../../utils/api';
 import { ProvisioningStatusPage } from './quickstart/ProvisioningStatusPage';
-import { useEffect } from "react";
-import { useLocation } from "react-router-dom";
-
-/*
- * Pages that are navigated to from long content pages 
- * stay scrolled down. This is a component that will 
- * scroll the window up.
- */
-const ScrollToTop = (): null => {
-    const { pathname } = useLocation();
-
-    useEffect(() => {
-        window.scrollTo(0, 0);
-    }, [pathname]);
-
-    return null;
-};
 
 const getRegions: QueryRegionFun = () => {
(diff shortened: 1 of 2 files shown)


frontend/src/main/dc-migration-assistant-fe/components/aws/AwsRoutes.tsx
```diff
@@ -22,6 +22,23 @@ import { QuickStartDeploy } from './quickstart/QuickStartDeploy';
...
 import { ProvisioningStatusPage } from './quickstart/ProvisioningStatusPage';
+import { useEffect } from "react";
+import { useLocation } from "react-router-dom";
+
+/*
+ * Pages that are navigated to from long content pages 
+ * stay scrolled down. This is a component that will 
+ * scroll the window up.
+ */
+const ScrollToTop = (): null => {
```
bpar476, line 33: Whats the reasoning for having this as a component instead of just putting the effect inside of the `ProvisioningStatusPage`?

dylanrathbone: I think it was an initial misunderstanding of `useEffect`. I see now that it always runs after the component rendered and applies its effect. I've made the relevant changes.

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
```diff
@@ -73,6 +75,18 @@ const inProgressStages = [
...
 
 export const ProvisioningStatusPage: FunctionComponent = () => {
+    
+  /*
+   * Pages that are navigated to from long content pages 
+   * stay scrolled down. 
+   * 
+   * Scroll the window up after render.
+   */
+    const { pathname } = useLocation();
```
bpar476, line 85: Is it necessary to use the pathname? This only needs to run when the `ProvisioningStatusPage` is first rendered. Once the path changes, the `ProvisioningStatusPage` won't be mounted anymore so this hook can't run 🤔

dylanrathbone: My initial thoughts on this were that it was required as the target page to which the effect should be applied. However this is not the case; `If present, effect will only activate if the values in the list change`. As you suggest the parameter is redundant in this context.

PR #227, opened 2020-05-27
Chet 408 fix jira 8.2

theghostwhoforks:
Fixes failing plugin in jira 8.2. Should fix in jira 7.13 as well 🤞


core/pom.xml
```diff
@@ -140,13 +149,90 @@
         <dependency>
             <groupId>com.fasterxml.jackson.dataformat</groupId>
             <artifactId>jackson-dataformat-xml</artifactId>
-            <version>${jackson.version}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>jakarta.xml.bind</groupId>
```
bpar476, line 154: What does this library do?

bpar476:
Maybe we can repurpose the `func-tests` module to do a quick and dirty integration test

core/pom.xml
theghostwhoforks: it pulls in xml bind library from the jakarta that is a multi version library.

It brings in java 9 specific features - in most cases this is OK. However, older versions of atlassian-rest parses the jar, and reads the java 9 class, causing the rest module to fail.

It's not a direct dependency; it's a nested dependency of jackson. Sheer madness.

theghostwhoforks:
> Maybe we can repurpose the `func-tests` module to do a quick and dirty integration test

Vica and I are planning on doing exactly this. 👍

PR #229, opened 2020-05-28
CHET-425: Add endpoint for ASI candidates.

tarka:


api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/AWSGlobalInfrastructureEndpoint.kt
```diff
@@ -38,4 +40,18 @@ class AWSGlobalInfrastructureEndpoint(private val globalInfrastructure: GlobalIn
             .ok(regions)
             .build()
     }
+
+    @GET
+    @Path("/asi")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    fun getAvailableASIs(): Response {
```
bpar476, line 48: Are we able to make this return the ASI prefix? That would make integrating with the UI a lot easier

tarka: Good idea. Let me take a look.

tarka: I've added ExportPrefix as a validity check too.

PR #230, opened 2020-05-28
CHET-428: Helper for uploading all captured paths to S3

bpar476:
# Summary

* New class which uses attachment sync manager to get listened files
* Use S3Uploader to upload them all to S3

## Follow up

* Create instance during Final sync phase (DB migration)
* Run upload

## Notes

* I wanted to make this a service and export it as a bean injecting the Uploader is difficult as it depends on the report and `S3Config`. I tried picking apart the uploader but it's a bit of a rabbit hole so I'm just going to construct it at migration time.
* A good future refactor would be to decouple the S3 config, S3 uploading, and S3 reporting from the `S3Uploader` implementation

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/S3FinalFileSync.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/S3FinalFileSync.kt
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2020 Atlassian
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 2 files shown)


core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/S3FinalFileSync.kt
```diff
@@ -0,0 +1,36 @@
...
+ */
+
+package com.atlassian.migration.datacenter.core.fs
+
+import com.atlassian.migration.datacenter.core.fs.captor.AttachmentSyncManager
+import com.atlassian.migration.datacenter.core.util.UploadQueue
+import java.nio.file.Path
+import java.nio.file.Paths
+
+class S3FinalFileSync(private val attachmentSyncManager: AttachmentSyncManager, private val uploader: Uploader) {
```
theghostwhoforks, line 24: like all the kotlin here in core!

theghostwhoforks:
Agree to pushing the uploader refactoring. It's a bit intertwined ATM

PR #231, opened 2020-05-28
CHET-409: run file final sync during offline period

bpar476:
## Summary

* Implement scheduled job for running final file sync
* Start scheduled job in DB API endpoint

## Note

* This is not very well integrated with the state machine, however I think that can come as a follow-up once the final sync is completed end-to-end
* I'm thinking of changing the name and path of the DB endpoint so it encompasses both the DB migration and the fs final sync. Maybe `FinalSyncEndpoint` or something like that. Any thoughts?

### DIFF
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationEndpoint.kt
@@ -17,6 +17,7 @@ package com.atlassian.migration.datacenter.api.db
 
 import com.atlassian.migration.datacenter.core.aws.db.DatabaseMigrationService
 import com.atlassian.migration.datacenter.core.aws.db.restore.SsmPsqlDatabaseRestoreService
+import com.atlassian.migration.datacenter.core.fs.captor.S3FinalSyncService
 import com.atlassian.migration.datacenter.spi.MigrationService
 import com.atlassian.migration.datacenter.spi.exceptions.InvalidMigrationStageError
 import com.fasterxml.jackson.annotation.JsonAutoDetect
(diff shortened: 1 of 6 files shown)


api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationEndpoint.kt
```diff
@@ -56,40 +58,45 @@ class DatabaseMigrationEndpoint(
...
         if (migrationService.currentStage.isDBPhase) {
             return Response
-                .status(Response.Status.CONFLICT)
-                .entity(mapOf("status" to migrationService.currentStage))
-                .build()
+                    .status(Response.Status.CONFLICT)
+                    .entity(mapOf("status" to migrationService.currentStage))
+                    .build()
         }
         val started = databaseMigrationService.scheduleMigration()
```
theghostwhoforks, line 65: we can check if migration has started before the next step - that way we don't have to schedule the finalsync and cancel it instead of a composite condition later

api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationEndpoint.kt
```diff
@@ -56,40 +58,45 @@ class DatabaseMigrationEndpoint(
...
+        val fsSyncStarted = finalSyncService.scheduleSync()
```
theghostwhoforks, line 66: we can check if migration has started before the next step - that way we don't have to schedule the finalsync and cancel it instead of a composite condition later

api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationEndpoint.kt
```diff
@@ -36,16 +37,17 @@ import javax.ws.rs.core.Response
 
 @Path("/migration/db")
 class DatabaseMigrationEndpoint(
```
theghostwhoforks, line 39: /nit should we change the name of this endpoint to FinalSyncEndpoint?

bpar476: > I'm thinking of changing the name and path of the DB endpoint so it encompasses both the DB migration and the fs final sync. Maybe FinalSyncEndpoint or something like that. Any thoughts?

I thought maybe too. Let's revisit when completing the final sync story

api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationEndpoint.kt
bpar476: Will fix in follow up

PR #236, opened 2020-05-29
CHET-185: Lock down migration bucket

dylanrathbone:
Tests using a `make createstack` show that the migration buckets are now created with the correct public access settings:

image

### DIFF
--- a/templates/migration-stack/Makefile
+++ b/templates/migration-stack/Makefile
@@ -35,4 +35,6 @@ deletestack:
 	aws cloudformation delete-stack --stack-name $(STACK_NAME)
 updatestack:
 	aws cloudformation update-stack --stack-name $(STACK_NAME) --template-body file://migration-helper.yml --parameters $(PARAMS) --capabilities CAPABILITY_IAM
+uploadtemplate: rendertemplate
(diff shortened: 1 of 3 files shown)


templates/migration-stack/Makefile
```diff
@@ -35,4 +35,6 @@ deletestack:
 	aws cloudformation delete-stack --stack-name $(STACK_NAME)
 updatestack:
 	aws cloudformation update-stack --stack-name $(STACK_NAME) --template-body file://migration-helper.yml --parameters $(PARAMS) --capabilities CAPABILITY_IAM
+uploadtemplate: rendertemplate
+	aws s3 cp --acl=public-read migration-helper.yml s3://trebuchet-public-resources/migration-helper.yml
```
theghostwhoforks, line 39: This is good! Also note - we don't need to allow CORS for the migration template as it is used by the java SDK

theghostwhoforks:
Is this tested and confirmed @dylanrathbone ?

dylanrathbone:
> Is this tested and confirmed to do what is required @dylanrathbone ?

Testing a migration now to make sure we get the desired effect on the bucket i.e. 

image

PR #240, opened 2020-05-29
Hotfix: missing beans and bad asi endpoint in ui

bpar476:


frontend/src/main/dc-migration-assistant-fe/api/provisioning.ts
```diff
@@ -27,7 +27,7 @@ export enum ProvisioningStatus {
 
 enum RestApiPathConstants {
     GetDeploymentStatusPath = 'aws/stack/status',
-    GetASIInfoPath = `/aws/global-infrastructure/asi`,
+    GetASIInfoPath = `aws/global-infrastructure/asi`,
```
theghostwhoforks, line 30: 🤣

PR #245, opened 2020-06-01
Chet 443 store physical resource

theghostwhoforks:
Store queue resources and add a SQS API interface

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/SqsApi.kt
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/SqsApi.kt
@@ -1,0 +1,7 @@
+package com.atlassian.migration.datacenter.core.aws
+
+import java.util.*
+
+interface SqsApi {
+    fun getQueueLength(queueName: String) : Optional<Int>
+}
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
@@ -31,12 +31,14 @@ import software.amazon.awssdk.services.autoscaling.model.AutoScalingGroup;
 import software.amazon.awssdk.services.autoscaling.model.DescribeAutoScalingGroupsRequest;
 import software.amazon.awssdk.services.autoscaling.model.DescribeAutoScalingGroupsResponse;
 import software.amazon.awssdk.services.autoscaling.model.Instance;
+import software.amazon.awssdk.services.cloudformation.model.Output;
 import software.amazon.awssdk.services.cloudformation.model.Stack;
+import software.amazon.awssdk.services.cloudformation.model.StackResource;
 
-import java.util.HashMap;
 import java.util.Map;
 import java.util.Optional;
 import java.util.function.Supplier;
+import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
 /**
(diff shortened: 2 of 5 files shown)


spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/infrastructure/MigrationInfrastructureDeploymentService.kt
```diff
@@ -30,4 +30,7 @@ interface MigrationInfrastructureDeploymentService : DeploymentService {
      */
     @Throws(InvalidMigrationStageError::class)
     fun deployMigrationInfrastructure(params: Map<String, String>)
+
+    fun getDeadLetterQueueResource() : String
```
bpar476, line 34: Do we need these methods in the interface? This is injecting the implementation into the interface.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
```diff
@@ -195,4 +213,25 @@ private void ensureStackOutputsAreSet() {
...
+        return String.format("%s-migration", migrationService.getCurrentContext().getApplicationDeploymentId());
+    }
+
+    protected void storeMigrationStackDetailsInContext(String migrationStackDeploymentId) {
+        final MigrationContext context = migrationService.getCurrentContext();
+        context.setHelperStackDeploymentId(migrationStackDeploymentId);
+        context.save();
+    }
+
+    //TODO: CHET443-Store in context and no longer needed to do this?
```
bpar476, line 227: Good suggestion :) Do we need a TODO comment if there is a ticket for it?

core/src/main/java/com/atlassian/migration/datacenter/core/aws/SqsApi.kt
```diff
@@ -0,0 +1,7 @@
+package com.atlassian.migration.datacenter.core.aws
+
+import java.util.*
+
+interface SqsApi {
```
bpar476, line 5: What's the reason for having an interface for this class?

theghostwhoforks: Will convert to a class after this is merged - I wanted to get this in first and then convert to help parallelize the tasks in this story

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
theghostwhoforks: I used the same ticket number :)

I'll remove it in the subsequent PR

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/infrastructure/MigrationInfrastructureDeploymentService.kt
theghostwhoforks: I thought we are injecting the interface in the bean configuration and that this was necessary? Looks like the actual class is being used.

Let me remove this right now; we'll have a refactoring to use the interface type later on

bpar476: We use it as the actual class because we don't want to tie the interface to the AWS migration stack. However, it doesn't make much difference at the moment.

PR #246, opened 2020-06-01
CHET-362: Inform users on approximate deployment times

dylanrathbone:
Update the wording on the deployment stage to give user an indication on the approximate time it will take for the deployment to complete. Create hyperlink to deploying stack.

- Deployment without an ASI takes 50 mins
- Deployment with an ASI takes 20 mins

**Deployment with an ASI**
`We're deploying Jira Data Center on AWS now. This should take around 50 minutes.`
image

**Deployment without an ASI**
`We're deploying Jira Data Center on AWS now. This should take around 30 minutes.`
image


frontend/src/main/resources/i18n/dc-migration-assistant.properties
```diff
@@ -75,7 +75,7 @@ atlassian.migration.datacenter.provision.aws.form.defaultError=must match
 atlassian.migration.datacenter.provision.aws.form.wrongNumberAZError=You must supply exactly 2 Availability Zones
 atlassian.migration.datacenter.provision.aws.form.stackName.label=Stack Name
 atlassian.migration.datacenter.provision.aws.form.stackName.validationMessage=Stack name should only include lower-case letters, numbers, periods, and hyphens.
-atlassian.migration.datacenter.provision.aws.wait.description=We're deploying each stack of your AWS infrastructure now
+atlassian.migration.datacenter.provision.aws.wait.description=We're deploying each stack of your AWS infrastructure now. This can take up to 20 minutes, use the CloudFormation console to monitor the deployment in real-time.
```
theghostwhoforks, line 78: Should we remain consistent with quickstarts and use the same terminology (about 30 mins)

https://aws.amazon.com/quickstart/architecture/jira/

bpar476: It aso depends on if there is an ASI or not

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPage.tsx
```diff
@@ -202,6 +202,15 @@ export const MigrationTransferPage: FunctionComponent<MigrationTransferProps> =
             <TransferContentContainer>
                 <h1>{heading}</h1>
                 <p>{description}</p>
+                <a
```
bpar476, line 205: I don't think we can put this here as it will be rendered in the file copy and final sync stages too but it's only relevant for the quickstart part. What's the content of the link?

bpar476: We could change the `description` to accept the type `ReactNode | String` which would let you pass the description text, and the `<a>` as a prop

dylanrathbone: Damn. Good catch, let me update with a potential fix

PR #251, opened 2020-06-02
CHET-450: final sync status includes file queue status

bpar476:
# Summary

* Expand final sync status endpoint to include the status of the SQS processing
* Implement test that serialisation works as expected
* Update UI to use new API


api/pom.xml
```diff
@@ -82,5 +82,10 @@
             <artifactId>kotlin-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
bpar476, line 85: Turns out this is required to mock the `DatabaseMigrationService` or any class that uses `com.atlassian.scheduler`

api/src/test/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpointTest.kt
```diff
@@ -0,0 +1,104 @@
...
+    @MockK
+    lateinit var migrationService: MigrationService
+    @MockK
+    lateinit var ssmPsqlDatabaseRestoreService: SsmPsqlDatabaseRestoreService
+    @MockK
+    lateinit var s3FinalSyncService: S3FinalSyncService
+    @InjectMockKs
+    lateinit var sut: FinalSyncEndpoint
+
+    class DurationDeserializer(t: Class<Duration>?) : StdDeserializer<Duration>(t) {
```
bpar476, line 66: For some reason, the Duration couldn't be deserialized by Jackson so I wrote a naive deserializer so that we could test the serialization works as expected

api/src/main/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpoint.kt
```diff
@@ -44,7 +45,7 @@ class FinalSyncEndpoint(
         private val ssmPsqlDatabaseRestoreService: SsmPsqlDatabaseRestoreService,
         private val finalSyncService: S3FinalSyncService
 ) {
-    private val mapper: ObjectMapper = ObjectMapper()
+    private val mapper: ObjectMapper = ObjectMapper().registerKotlinModule()
```
theghostwhoforks, line 48: 👍
We should collect all usages of object mapper and either register the mapper with spring or make it a bean that we can inject in

api/src/test/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpointTest.kt
```diff
@@ -0,0 +1,104 @@
...
+import java.time.Duration
+import java.util.Optional
+import kotlin.test.assertEquals
+
+
+@ExtendWith(MockKExtension::class)
+internal class FinalSyncEndpointTest {
+
+    companion object {
+        val mapper = ObjectMapper()
```
theghostwhoforks, line 50: 👍
We should collect all usages of object mapper and either register the mapper with spring or make it a bean that we can inject in

PR #253, opened 2020-06-03
Chet 443 sqs api

theghostwhoforks:
Store Stack outputs and resources in migration context. Amend tests in migration service to verify this change in behaviour

Current response from  API -

Screenshot 2020-06-03 at 12 33 45

The S3 file service will fetch uploaded file count from the report object as part of a subsequent PR


theghostwhoforks:
The `-2` in the screenshot above is due to the test number of uploaded files being equal to 0. This will be set right in a subsequent PR

core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/SqsApiImpl.kt
```diff
@@ -13,31 +15,46 @@ class SqsApiImpl(private val sqsClientSupplier: Supplier<SqsAsyncClient>) : SqsA
...
     }
 
+    private val sqsAsyncClient: SqsAsyncClient
+        get() {
+            val sqsAsyncClient: SqsAsyncClient? = _sqsAsyncClient ?: sqsClientSupplier.get()
+            return sqsAsyncClient!!
+        }
+
+    @Throws(AwsQueueError::class)
     override fun getQueueLength(queueUrl: String): Int? {
```
bpar476, line 28: Can this just return `Int` now instead of `Int?` it looks like all code paths return a non-null now :)

core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/captor/S3FinalSyncService.kt
```diff
@@ -49,16 +52,20 @@ class S3FinalSyncService(private val migrationRunner: MigrationRunner, private v
...
     fun getFinalSyncStatus() : FinalFileSyncStatus {
-        return FinalFileSyncStatus(0,0)
+        val currentContext = migrationService.currentContext
+        val migrationQueueUrl = currentContext.migrationQueueUrl
+        val migrationDLQueueUrl = currentContext.migrationDLQueueUrl
+
+        val itemsInQueue = sqsApi.getQueueLength(migrationQueueUrl)
+        val itemsInDlQueue = sqsApi.getQueueLength(migrationDLQueueUrl)
+
+        return FinalFileSyncStatus(0, itemsInDlQueue!! + itemsInQueue!!)
```
bpar476, line 63: With the current implementation, the items in the deadletter queue are ignored. Do we still want to report them as "enqueued"? I think it's a bit ingenuous as the "enqueued" part of the file sync status suggests that it is in the queue to be downloaded.

core/src/test/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentServiceTest.java
```diff
@@ -242,6 +245,41 @@ private void givenMigrationStackDeploymentWillCompleteSuccessfully() {
             put(STACK_RESOURCE_QUEUE_NAME, StackResource.builder().physicalResourceId(QUEUE_PHYSICAL_RESOURCE_ID).build());
             put(STACK_RESOURCE_DEAD_LETTER_QUEUE_NAME, StackResource.builder().physicalResourceId(DEAD_LETTER_QUEUE_PHYSICAL_RESOURCE_ID).build());
         }});
+
+        expectStackDetailsToBePersistedInMigrationContext();
+    }
+
+    //Migration context stub object/data-class may make this test simpler
+    private void expectStackDetailsToBePersistedInMigrationContext() {
+        doNothing().when(mockContext).setFsRestoreSsmDocument("");
```
bpar476, line 254: Can we make an assertion that these methods are called with the given arguments we stub them?

core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/SqsApiImpl.kt
```diff
@@ -13,31 +15,46 @@ class SqsApiImpl(private val sqsClientSupplier: Supplier<SqsAsyncClient>) : SqsA
 
     companion object {
         private val logger = LoggerFactory.getLogger(SqsApiImpl::class.java)
+        private var _sqsAsyncClient: SqsAsyncClient? = null
     }
 
+    private val sqsAsyncClient: SqsAsyncClient
```
bpar476, line 21: I think maybe we shouldn't cache the client. If the user re-authenticates or changes region between the first access and the second access, the client will be stale. This has caused issues for us during development in the past.

core/src/test/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentServiceTest.java
theghostwhoforks: Because these aren't lenient mocks, if the mockContext gets a call on `setFsRestoreSsmDocument` say, it would fail.

Having a verification would mean that we make the mock lenient.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/SqsApiImpl.kt
theghostwhoforks: Sure thing!

core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/SqsApiImpl.kt
theghostwhoforks: Done

core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/captor/S3FinalSyncService.kt
theghostwhoforks: I'll remove it for now - if items exist in the DL queue, they user may want to know that there are fails that are there in the S3 bucket that have not been copied over to the FS - a topic for discussion next week

PR #254, opened 2020-06-03
CHET-432: Prevent chrome from remebering AWS creds

dylanrathbone:
Setting `autocomplete="off"` on fields has two effects:

1. It tells the browser not to save data inputted by the user for later autocompletion on similar forms, though heuristics for complying vary by browser.

2. It stops the browser from caching form data in the session history. When form data is cached in session history, the information filled in by the user is shown in the case where the user has submitted the form and clicked the Back button to go back to the original form page.

[Source](https://developer.mozilla.org/en-US/docs/Web/Security/Securing_your_site/Turning_off_form_autocompletion)

*NOTE:*
Something to bare in mind is that in doing this (`autocomplete="off"`) we may break password manager integrations. However the counter argument is that using trebuchet is a once off task which will unlikely need password manager assistance.


bpar476:
Should we also add `type="password"` so we get the ******* when they enter their input?

dylanrathbone:
> Should we also add `type="password"` so we get the ******* when they enter their input?

True, these are sensitive details after all. Shoulder 🏄 be gone...

image

PR #255, opened 2020-06-03
Chet 415 file attachment

theghostwhoforks:
Refactored attachment listener to capture attachments, not just paths.

Removes the twin stream use case in attachment listener


core/src/main/java/com/atlassian/migration/datacenter/core/fs/captor/DefaultAttachmentSyncManager.java
```diff
@@ -69,4 +63,34 @@ public void captureAttachmentPath(Path attachmentPath) {
...
+            Path attachmentPath = attachmentFile.toPath();
+            logger.debug("Recording file path - {} ", attachmentPath);
+            this.captureAttachmentPath(attachmentPath);
+        }
+    }
+
+    private void captureAttachmentPath(Path attachmentPath) {
+        logger.debug("captured attachment for final sync: {}", attachmentPath.toString());
+
+        FileSyncRecord record = ao.create(FileSyncRecord.class);
```
dylanrathbone, line 89: Minor, but down here its hard to know what `ao` is. Can we rename to `activeObjects`?

theghostwhoforks:
Hey @bpar476 @dylanrathbone 

I'd made a few changes to this yesterday - I ensured that the api does not have a dependency on `jira-api`by pulling attachmentsync and attachmentcaptor to 2 separate implementations.

Can you quickly take a look and I'll merge this in soon

core/src/main/java/com/atlassian/migration/datacenter/core/fs/jira/captor/DefaultAttachmentCaptor.java
```diff
@@ -0,0 +1,73 @@
...
+import com.atlassian.jira.issue.attachment.AttachmentStore;
+import com.atlassian.migration.datacenter.dto.FileSyncRecord;
+import com.atlassian.migration.datacenter.spi.MigrationService;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+import java.nio.file.Path;
+
+public class DefaultAttachmentCaptor implements AttachmentCaptor {
```
theghostwhoforks, line 30: Because we have this in a separate interface, we don't need to mock the same class in API - this ensures that we don't pull jira-api into the `api` module,even as a test dependency

PR #260, opened 2020-06-04
CHET-445: Create no-op implementation of a queue watcher to transitio…

theghostwhoforks:
…n to final sync await stage after the database migration is complete

This PR attempts to maintain compatibility with the current state machine transition by -

1. Transitioning to FS_SYNC_AWAIT after the DB migration is complete
2. Transitioning  to Validate from FS Sync after the queue has drained *

* This is a no-op in this PR. An implementation similar to how we wait for CFn stack deployment to complete will be built in a subsequent PR

Merging this will help iterate incrementally towards the final sync using SQS


theghostwhoforks:
@bpar476  @tarka  Feel free to merge this in post approval

api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationProgress.kt
```diff
@@ -5,7 +5,7 @@ import com.atlassian.migration.datacenter.spi.MigrationStage.*
 import java.time.Duration
 
 
-enum class DBMigrationStatus {
+enum class FinalSyncMigrationStatus {
```
bpar476, line 8: It looks like the enum has been renamed but not the parent class.

Also, the final sync endpoint only uses this status to report the progress of the DB migration. Do we really need to rename this enum?

theghostwhoforks:
> Currently I think the behaviour will error if the queue drains before the database restore completes as we expect to transition to `VALIDATE` from `FS_SYNC_WAIT`. I think we need to either:
>     * wait for the queue to drain after the DB migration has completed (will be instant if the queue is already empty otherwise, block until the queue is empty)
> 
>     * Do the transition to `VALIDATE` somewhere else which can check that the DB migration and FS sync have completed

My thinking is this -

* Database dump transitions to FS_SYNC_WAIT after the restore completes successfully
* The final sync job uploads the captured files and waits for the state to transition to FS_SYNC_WAIT
* Once the state is in FS_SYNC_WAIT, it will poll the queue. If it is already empty, it'll make the transition to VALIDATE. Else it will keep waiting until the queue is empty

This way, even if the queue drains before the database restore completes, the scheduled job will wait until the state transitions to FS_SYNC_WAIT, checks if the queue is empty, 
awaits the queue to be drained and then transitions to validate. This should ensure that even if the queue drains before the file system and ensures that we complete the transition in step. 

Is there something that I am not seeing here?

theghostwhoforks:
here's a draft POC - https://github.com/atlassian-labs/dc-migration-assistant/compare/CHET-445-transition-to-validation...CHET-445-transition-to-validation-spike-1?expand=1

bpar476:
> here's a draft POC - https://github.com/atlassian-labs/dc-migration-assistant/compare/CHET-445-transition-to-validation...CHET-445-transition-to-validation-spike-1?expand=1



> here's a draft POC - https://github.com/atlassian-labs/dc-migration-assistant/compare/CHET-445-transition-to-validation...CHET-445-transition-to-validation-spike-1?expand=1

Looks like this will work :) However there a lot of moving pieces. Might be good to have a test that the transitions occur in the correct order

api/src/main/kotlin/com/atlassian/migration/datacenter/api/db/DatabaseMigrationProgress.kt
theghostwhoforks: Done

PR #261, opened 2020-06-04
CHET-452: cleanup database password secret

bpar476:
## Summary

* Use composition to orchestrate deletion or resources
* Implement deletion of secrest manager secret
* Update migration infrastructure interface to handle mixed state of some resources deleted, some not deleted

image


core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/AWSMigrationInfrastructureCleanupService.kt
```diff
@@ -30,11 +30,26 @@ class AWSMigrationInfrastructureCleanupService : MigrationInfrastructureCleanupS
     override fun getMigrationInfrastructureCleanupStatus(): InfrastructureCleanupStatus {
         logger.info("querying state of AWS resource cleanup")
 
-        return InfrastructureCleanupStatus.CLEANUP_COMPLETE
+        val tasks = cleanupTaskFactory.getCleanupTasks()
+        if (tasks.any { it.getMigrationInfrastructureCleanupStatus() == InfrastructureCleanupStatus.CLEANUP_IN_PROGRESS }) {
```
bpar476, line 34: This is how we can handle mixed states of deletion.

First, check if any delete is in progress. If so, delete is in pogress.
Then, check if any resources are not started deletion (i.e. are still around and need to be deleted)
Then check fi all of them have been deleted successfully (i.e. nothing left to delete)
Then, check if any errors occured. This should happen rarely, only if our attempt to delete doesn't work and retrying won't help

jira-plugin/src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantBeanConfiguration.java
```diff
@@ -318,7 +322,19 @@ public S3FinalSyncService s3FinalSyncService(MigrationRunner migrationRunner, S3
...
-    public MigrationInfrastructureCleanupService awsMigrationInfrastructureCleanupService() {
-        return new AWSMigrationInfrastructureCleanupService();
+    public DatabaseSecretCleanupService databaseSecretCleanupService(Supplier<SecretsManagerClient> secretsManagerClientSupplier, TargetDbCredentialsStorageService targetDbCredentialsStorageService
+    ) {
+        return new DatabaseSecretCleanupService(secretsManagerClientSupplier, targetDbCredentialsStorageService);
+    }
+
+    @Bean
+    public AWSCleanupTaskFactory cleanupTaskFactory(DatabaseSecretCleanupService databaseSecretCleanupService) {
+        return new AWSCleanupTaskFactory(databaseSecretCleanupService);
```
theghostwhoforks, line 332: Does this contain a task of just one list for now?

PR #264, opened 2020-06-05
CHET-298: Source Jira templates from GitHub & Sync migration template to bucket

dylanrathbone:
# Change 1
Reference Jira templates from official GitHub AWS Org repo:
https://github.com/aws-quickstart/quickstart-atlassian-jira

# Change 2
On push we want to re-sync all of our `.yml` files with the `S3` bucket `trebuchet-public-resources` i.e.

- `migration-helper.yml`

To facilitate pushing these assets we now have:

- A new GitHub Action called `UploadMigrationTemplate`
- A new AWS user `TrebuchetAssetS3Uploader`
- New secrets for `TrebuchetAssetS3Uploader`, see below:

image


dylanrathbone:
> Hmm, pinning the templates here is a good idea as we can be more fine grained about features shown off in trebuchet. However, they're still going to use the `aws-quickstart` version for the nested stacks. Should we put every template in here? Maybe use the quickstart as a submodule and synchronise the whole directory with S3, then force the customer to use our S3 bucket as the `QuickstartS3Bucket`?

Yes this goes back to the whole, upstream changes forced on you thing.... If we want to prevent potential upstream breaking changes then adding all of the templates to this bucket would definitely worth considering. Is enforcing usage of only our bucket (`QuickstartS3Bucket`) an option though? Even so, it would require documentation changes I'm sure.

bpar476:
The tests are probably just failing because the quickstart URL is changed. We have assertions that expect the trebuchet bucket as the template URL

bpar476:
> Ooh just remembered, the frontend also needs an update to the quickstart template locations!

I think this is the file: `frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/QuickStartDeploy.tsx`

frontend/.env
```diff
@@ -0,0 +1,2 @@
+DEFAULT_QUICKSTART_WITH_VPC_PARAMETER_URL=https://aws-quickstart.s3.amazonaws.com/quickstart-atlassian-jira/templates/quickstart-jira-dc-with-vpc.template.yaml
+DEFAULT_QUICKSTART_STANDALONE_PARAMETER_URL=https://aws-quickstart.s3.amazonaws.com/quickstart-atlassian-jira/templates/quickstart-jira-dc.template.yaml
```
dylanrathbone, line 2: As the `.env` is frontend specific it now lives under `frontend`

frontend/config/jest/jest.config.js
```diff
@@ -22,6 +22,7 @@ const paths = {
 };
 
 module.exports = {
+    setupFiles: ['dotenv/config'],
```
dylanrathbone, line 25: Source `.env` for test purposes

frontend/package.json
```diff
@@ -109,7 +109,7 @@
         "build": "webpack --config ./config/webpack.config.js",
         "build:dev": "webpack --config ./config/webpack.config.js --mode=development",
         "dev-server": "webpack-dev-server --config ./config/webpack.config.js --mode=development --env=dev-server",
-        "test": "jest --config ./config/jest/jest.config.js --passWithNoTests",
+        "test": "jest --setupFiles dotenv/config --config ./config/jest/jest.config.js --passWithNoTests",
```
dylanrathbone, line 112: Source `.env` for test purposes

frontend/.env
```diff
@@ -0,0 +1,2 @@
+DEFAULT_QUICKSTART_WITH_VPC_PARAMETER_URL=https://raw.githubusercontent.com/aws-quickstart/quickstart-atlassian-jira/develop/templates/quickstart-jira-dc-with-vpc.template.yaml
```
theghostwhoforks, line 1: From what I remember, this needs to be prefixed by `REACT_APP_` for it to be picked up.

See `Your project can consume variables declared in your environment as if they were declared locally in your JS files. By default you will have NODE_ENV defined for you, and any other environment variables starting with REACT_APP_.` in https://create-react-app.dev/docs/adding-custom-environment-variables/

bpar476: If I recall correctly, That's only when using create-react-app as they handle the dotenv integration for you

bpar476:
Should we just close this PR and open a new one now that we're going to pull them into the repo?

frontend/.env
dylanrathbone: Dont seem to need this Varun.

theghostwhoforks: When I was testing this, I had to prefix this variable name with `REACT_APP_`, else the webpack plugin didn't replace the variables correctly.

Also, this functionality comes from react-scripts (which we include in our `package.json`)

@dylanrathbone - you can check this quickly by adding the prefix and verifying if the property is replaced & read

dylanrathbone:
> Should we just close this PR and open a new one now that we're going to pull them into the repo?

So slight change in thinking. I think we may have an easy win here, we can actually do this now by pulling the templates from GitHub:

```
DEFAULT_QUICKSTART_WITH_VPC_PARAMETER_URL=https://raw.githubusercontent.com/aws-quickstart/quickstart-atlassian-jira/develop/templates/quickstart-jira-dc-with-vpc.template.yaml
DEFAULT_QUICKSTART_STANDALONE_PARAMETER_URL=https://raw.githubusercontent.com/aws-quickstart/quickstart-atlassian-jira/develop/templates/quickstart-jira-dc.template.yaml
```

I've tested this approach and it works.

The reason for the change in approach was down to a couple of things that fell out of me looking into the static bucket option:

1. The issues we see with un-versioned templates etc. are only really related to existing stacks not new ones as in Trebuchet's case.
2. Maintaining these templates in our own bucket brings with it a lot of overhead. When making changes to the QS templates themselves we would also have to update the ones living in the Trebuchet code base (assuming we sync from the code base to the bucket via the github action we have)
3. Determining whether an update to the code base is necessary would also require additional effort.

The only real risk I can see here is if GitHub were to become unavailable

frontend/.env
dylanrathbone: I'm quite sure its doing what it should, on build of the frontend we see webpack sourcing the env variables correctly:

```
INFO] --- frontend-maven-plugin:1.9.1:yarn (compile-the-ui) @ frontend ---
[INFO] Running 'yarn build' in /Users/drathbone/Git/dc-migration-assistant/frontend/src/main/dc-migration-assistant-fe
[INFO] yarn run v1.19.1
[INFO] $ webpack --config ./config/webpack.config.js
[INFO] Loaded env vars
[INFO] {
[INFO]   'process.env.DEFAULT_QUICKSTART_WITH_VPC_PARAMETER_URL': '"https://raw.githubusercontent.com/aws-quickstart/quickstart-atlassian-jira/develop/templates/quickstart-jira-dc-with-vpc.template.yaml"',
[INFO]   'process.env.DEFAULT_QUICKSTART_STANDALONE_PARAMETER_URL': '"https://raw.githubusercontent.com/aws-quickstart/quickstart-atlassian-jira/develop/templates/quickstart-jira-dc.template.yaml"'
[INFO] }
```

The tests are also passing and the plugin is sourcing the templates from the correct locations based on these envs:

image

image

Sorry for the verbosity, just want to make sure we are all on the same page.

theghostwhoforks: oh nice! 👍

theghostwhoforks:
> The only real risk I can see here is if GitHub were to become unavailable

Yeah - I'd have been happier with if the AWS S3 bucket supports CORS, as S3 has higher uptime, but I reckon we should create a task to make the S3 bucket to source the templates configurable in the front end, merge this in and close this story out

dylanrathbone:
> > The only real risk I can see here is if GitHub were to become unavailable
> 
> Yeah, S3 has a higher uptime. I'd have been happier with if the AWS S3 bucket supports CORS, as S3 has higher uptime, but I reckon we should create a task to make the S3 bucket to source the templates configurable in the front end, merge this in and close this story out

So a task to make the location of these templates configurable from the frontend? I can create a task for that.

If you're happy with these changes can you approve and I'll merge.

PR #266, opened 2020-06-05
Chet 445 transition to validation IT

theghostwhoforks:
Writes an integration test to verify if futures are chained properly


core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/captor/SqsQueueWatcher.kt
```diff
@@ -31,11 +33,33 @@ class SqsQueueWatcher(private val sqsAPi: SqsApi,
...
     override fun awaitQueueDrain(): Boolean {
-        awaitRunnableToComplete(::checkForStateToBeInFsSyncAwait).get()
-        awaitRunnableToComplete(::checkForQueueToBeEmpty).get()
-        migrationService.transition(MigrationStage.VALIDATE)
-        return true
+        //TODO: Change from when complete to toAccept/toCombine idioms
+        val completableFuture = this.awaitRunnableToComplete(::checkForStateToBeInFsSyncAwait)
+                .whenComplete { _, _ ->
+                    run {
+                        awaitRunnableToComplete(::checkForQueueToBeEmpty).get()
```
theghostwhoforks, line 46: the `get` call here should be removed when the method changes to thenAccept.. or we use co-routines. 

Leaving it in for now as it does the job

PR #269, opened 2020-06-09
Update atlaskit packages. Update version of websocket-extensions with CVE

bpar476:
This fixes the dependabot issue

I had to regenerate yarn.lock which caused all the atlaskit resolutions to update. I had to do a workaround for a bug in the latest version of react-select.


frontend/package.json
```diff
@@ -100,6 +99,10 @@
         "ts-jest": "^25.4.0",
         "typescript": "^3.6.3"
     },
+    "resolutions": {
+        "**/websocket-extensions": "~0.1.4",
```
bpar476, line 103: This forces to use 0.1.x where x > 4 i.e. the version with the fix for the CVE that dependabot found

frontend/package.json
```diff
@@ -100,6 +99,10 @@
...
+        "**/@types/react-select": "3.0.11"
```
bpar476, line 104: When re-generating yarn.lock, our atlaskit resolutions updated. There is a bug in atlaskit select so we need to hardcode the react-select types version for now. I've informed the design system team and they're looking into it.

PR #270, opened 2020-06-09
CHET-453: empty migration bucket

bpar476:
# Summary

* Implement cleanup job for emptying the migration bucket
* Implement unit and integration test for the cleanup

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/MigrationBucketCleanupService.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/MigrationBucketCleanupService.kt
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2020 Atlassian
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/MigrationBucketCleanupService.kt
```diff
@@ -0,0 +1,95 @@
...
+        while (contents.size != 0) {
+            val lastSize = contents.size
+            contents.forEach { obj ->
+                client.deleteObject { builder -> builder.bucket(bucket).key(obj.key()) }
+            }
+            response = client.listObjectsV2 { it.bucket(bucket).continuationToken(continuation) }
+            contents = response.contents()
+            continuation = response.continuationToken()
+
+            if (lastSize == contents.size) {
```
theghostwhoforks, line 65: Can the first iteration of `listObjects` return the same number of contents as the subsequent call with the `continuationToken`? In which case, we will enter the retry conditional block?

bpar476: I don't think so. This is from AWS docs:

    NextContinuationToken is sent when isTruncated is true, which means there are more keys in the bucket that can be listed. The next list requests to Amazon S3 can be continued with this NextContinuationToken . NextContinuationToken is obfuscated and is not a real key

PR #271, opened 2020-06-09
Chet 464 final sync uploaded count

theghostwhoforks:
Ok, this PR got a little bloated and now it contains -

1. A mechanism to stop attachment listener in FINAL_SYNC_WAIT stage. This is necessary as we don't want to keep listening to file records in case the migration fails and a user does not cancel/clean up the migration

2. Refactored the object graph/construction of Database migration service. We don't need to create both the service and the callback in the database service. Instead, we create the archive/upload/restore service with the callbacks

3.  We read the file system record collection for a given migration and using this, determine the number of files that have been uploaded to S3.

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
@@ -43,7 +43,6 @@ public class DatabaseMigrationService {
     private final DatabaseArchivalService databaseArchivalService;
     private final DatabaseArchiveStageTransitionCallback stageTransitionCallback;
     private final DatabaseArtifactS3UploadService s3UploadService;
-    private final DatabaseUploadStageTransitionCallback uploadStageTransitionCallback;
     private final SsmPsqlDatabaseRestoreService restoreService;
     private final MigrationService migrationService;
     private final MigrationRunner migrationRunner;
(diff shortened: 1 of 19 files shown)


core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -42,32 +41,23 @@
...
 
     public DatabaseMigrationService(Path tempDirectory, MigrationService migrationService,
-            MigrationRunner migrationRunner, DatabaseArchivalService databaseArchivalService,
-            DatabaseArchiveStageTransitionCallback stageTransitionCallback,
-            DatabaseArtifactS3UploadService s3UploadService,
-            DatabaseUploadStageTransitionCallback uploadStageTransitionCallback,
-            SsmPsqlDatabaseRestoreService restoreService,
-            DatabaseRestoreStageTransitionCallback restoreStageTransitionCallback,
-            AWSMigrationHelperDeploymentService migrationHelperDeploymentService) {
+                                    MigrationRunner migrationRunner, DatabaseArchivalService databaseArchivalService,
```
theghostwhoforks, line 53: I made this refactoring thinking I had to make changes to the database migration service. HAlfway through, I realized that I can make the changes to stop the event listener from the FinalSync job runner ,but persisted with these changes are they are useful to have in place

PR #272, opened 2020-06-09
CHET-454: delete migration bucket

bpar476:
# Summary

* Delete the bucket after emptying the bucket
* Extend unit and integration tests to ensure this works

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/MigrationBucketCleanupService.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/MigrationBucketCleanupService.kt
@@ -22,6 +22,7 @@ import com.atlassian.migration.datacenter.spi.infrastructure.MigrationInfrastruc
 import org.slf4j.Logger
 import org.slf4j.LoggerFactory
 import software.amazon.awssdk.services.s3.S3Client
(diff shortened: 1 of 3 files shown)


core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/MigrationBucketCleanupService.kt
```diff
@@ -72,23 +73,37 @@ class MigrationBucketCleanupService(private val migrationService: MigrationServi
             }
         }
         logger.info("all objects in bucket deleted")
+
+        if (!failedToEmpty) {
+            logger.info("deleting migration bucket")
+            try {
+                client.deleteBucket { it.bucket(bucket) }
+            } catch (e: NoSuchBucketException) {
+                logger.info("bucket already doesn't exist...")
```
dylanrathbone, line 82: can we add the bucket name to the log message?

bpar476: I didn't want to include the bucket name in case it included PII

bpar476: I can add it in next time I'm in that part of the code

PR #273, opened 2020-06-10
Chet 467 helper stack outputs

theghostwhoforks:
Ensure stack outputs are set after persisting values from CFn API

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
@@ -109,10 +109,14 @@ public class AWSMigrationHelperDeploymentService extends CloudformationDeploymen
         persistStackDetails(stackOutputs, migrationStackResources);
 
         try {
+            ensureStackOutputsAreSet();
             migrationService.transition(MigrationStage.FS_MIGRATION_COPY);
         } catch (InvalidMigrationStageError invalidMigrationStageError) {
             logger.error("error transitioning to FS_MIGRATION_COPY stage after successful migration stack deployment", invalidMigrationStageError);
             migrationService.error(invalidMigrationStageError.getMessage());
+        } catch (InfrastructureDeploymentError ide){
+            logger.error("Error storing migration stack outputs in the database", ide);
+            migrationService.error(ide.getMessage());
         }
     }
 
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
```diff
@@ -191,8 +195,6 @@ private String getMigrationStackPropertyOrOverride(Supplier<String> supplier, St
         if (documentOverride != null) {
             return documentOverride;
         }
-
-        ensureStackOutputsAreSet();
```
bpar476, line 195: Removing this check potentially means that anything that tries to query the migration stack before it's available will get incorrect values. If we're okay with this I reckon we get rid of the `ensureStackOutputsAreSet` method as it's only called immediately after persisting the stack outputs so a bit redundant.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
```diff
@@ -161,28 +161,27 @@ public String getDeadLetterQueueResource() {
...
-
-            AutoScalingGroup migrationStackGroup = response.autoScalingGroups().get(0);
-            Instance migrationInstance = migrationStackGroup.instances().get(0);
-
-            return migrationInstance.instanceId();
         }
+
+        String migrationStackAsg = getMigrationStackPropertyOrOverride(() -> migrationService.getCurrentContext().getMigrationStackAsgIdentifier(), "com.atlassian.migration.asgIdentifier");
+
+        //TODO: Store this once in the migration context once.
```
bpar476, line 172: I think we don't want to store the instance in the migration context in case it's destroyed for some reason which is why we use an autoscaling group.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
bpar476: Nice :)

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
theghostwhoforks: removed!

PR #274, opened 2020-06-10
CHET-451: integrate standalone migration stack deployment

bpar476:
# Summary
* Move migration stack related kotlin code to new package within `core/aws/infrastructure`
* Clean up some duplication between the parameter discovery mechanisms
* Implement factory that can provide a parameter strategy based on deployment mode
* Store deployment mode in the deployment context
* Update quickstart deployment to use the factory to get a strategy for the current deployment mode

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/migrationStack/QuickstartWithVPCMigrationStackInputGatheringStrategy.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/migrationStack/QuickstartWithVPCMigrationStackInputGatheringStrategy.kt
@@ -17,40 +17,20 @@
 package com.atlassian.migration.datacenter.core.aws.infrastructure.migrationStack
 
 import com.atlassian.migration.datacenter.core.aws.CfnApi
+import com.atlassian.migration.datacenter.spi.infrastructure.InfrastructureDeploymentError
 import software.amazon.awssdk.services.cloudformation.model.Stack
 
-class QuickstartWithVPCMigrationStackInputGatheringStrategy(private val cfnApi: CfnApi) : MigrationStackInputGatheringStrategy {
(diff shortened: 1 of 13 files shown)


core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/migrationStack/QuickstartWithVPCMigrationStackInputGatheringStrategy.kt
```diff
@@ -0,0 +1,36 @@
...
+import com.atlassian.migration.datacenter.spi.infrastructure.InfrastructureDeploymentError
+import software.amazon.awssdk.services.cloudformation.model.Stack
+
+class QuickstartWithVPCMigrationStackInputGatheringStrategy(private val cfnApi: CfnApi, private val standaloneMigrationStackInputGatheringStrategy: QuickstartStandaloneMigrationStackInputGatheringStrategy) : MigrationStackInputGatheringStrategy {
+
+    override fun gatherMigrationStackInputsFromApplicationStack(stack: Stack): Map<String, String> {
+        val applicationResources = cfnApi.getStackResources(stack.stackName())
+
+        val jiraStackResource = applicationResources["JiraDCStack"]
+        val jiraStack = cfnApi.getStack(jiraStackResource!!.physicalResourceId())
```
bpar476, line 29: Re-using the standalone strategy means we make one additional API request, however, we reduce a lot of code duplication so I reckon it's worth it. Basically we just get the standalone application stack from the with-vpc stack and pass that stack to the standalone stack input gatherer

theghostwhoforks: This is worth the extra call to get stack resources

PR #275, opened 2020-06-11
CHET-442: Create log directory if it does not exist.

theghostwhoforks:
SSM commands were trying to write log files to a directory that was not being created. This PR 

1. Creates a log directory under `/var/log/atlassian/dc-migration-assistant`
2. Refers to the above directory to store sync, error and db restore logs


theghostwhoforks:
> 🤯 How did this ever work

It's so weird - I don't know how it was working before. But it was definitely working!

PR #276, opened 2020-06-11
Chet 442 verify migration

theghostwhoforks:
Bunch of changes as I was testing the e2e flow


core/src/main/java/com/atlassian/migration/datacenter/core/fs/captor/DefaultAttachmentSyncManager.java
```diff
@@ -49,7 +49,7 @@ public DefaultAttachmentSyncManager(ActiveObjects activeObjects, MigrationServic
 
         final FileSyncRecord[] recordsForCurrentMigration = activeObjects.find(
                 FileSyncRecord.class,
-                Query.select().where("migration_id = ?", migration.getID()));
+                Query.select().where("MIGRATION_ID = ?", migration.getID()));
```
theghostwhoforks, line 52: had to do this to keep AO happy 

@vicav - is this expected when using AO?

jira-e2e-tests/jira/Dockerfile
```diff
@@ -1,7 +1,12 @@
 FROM atlassian/jira-software
 
+RUN apt-key adv --keyserver keyserver.ubuntu.com --recv-keys B97B0AFCAA1A47F044F244A07FCC7D46ACCC4CF8
```
theghostwhoforks, line 3: @tarka - Without this (i.e. postgres client libs), we can't complete a migration (i.e. go past the db export phase)

templates/migration-stack/migration-helper.yml
```diff
@@ -120,31 +120,31 @@ Resources:
   StackNameParameter:
     Type: AWS::SSM::Parameter
     Properties:
-      Name: /atlassian/config/migration-helper/cloud.aws.stack.name
+      Name: !Sub '/atlassian/config/migration-helper/${AWS::StackName}/cloud.aws.stack.name'
```
theghostwhoforks, line 123: These shouldn't be params. Instead the file processor should read these values from the CFn API 🤷‍♂️

bpar476: We can't include the stack name in this parameter because it's used by spring cloud Parameter Store configuration to get the stack name 😅

bpar476: I'm sure there is some way to do this better but I think it would be a decent chunk of work as the filesystem processor app is highly dependent on these parameter store params

theghostwhoforks: Done. Retained the older mechanism.

Note - this will limit us to 1 migration stack in a region :(

PR #277, opened 2020-06-11
CHET-89: Plugin analytics

tarka:


core/src/main/kotlin/com/atlassian/migration/datacenter/analytics/events/MigrationCompleteEvent.kt
```diff
@@ -0,0 +1,9 @@
+package com.atlassian.migration.datacenter.analytics.events
+
+import com.atlassian.analytics.api.annotations.EventName
+
+@EventName("atl.dc.migration.complete")
+data class MigrationCompleteEvent (
+        val pluginVersion: String,
+        val runTime: Long  // Seconds
```
bpar476, line 8: should we maybe name the field `runTimeSeconds` so we don't need the comment and we know the unit when reading the analytic event in redash or something?

core/src/main/kotlin/com/atlassian/migration/datacenter/analytics/events/MigrationFailedEvent.kt
```diff
@@ -0,0 +1,11 @@
+package com.atlassian.migration.datacenter.analytics.events
+
+import com.atlassian.analytics.api.annotations.EventName
+
+@EventName("atl.dc.migration.failed")
+data class MigrationFailedEvent(
+        val pluginVersion: String,
+        val failedStage: String,
+        val error: String,
+        val runTime: Long  // Seconds
```
bpar476, line 10: Ditto for previous `runTime` field

jira-plugin/src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantOsgiImportConfiguration.java
```diff
@@ -60,6 +61,11 @@ public JiraHome getJiraHome() {
         return importOsgiService(JiraHome.class);
     }
```
bpar476, line 63: Do we need an `EventPublisher` bean here too?

jira-plugin/src/main/resources/whitelist/analytics_whitelist.json
```diff
@@ -0,0 +1,12 @@
+{
+  "atl.dc.migration.started":[
+    "pluginVersion"
+  ],
+  "atl.dc.migration.complete": [
+    "pluginVersion"
+  ],
+  "atl.dc.migration.failed": [
+    "pluginVersion",
+    "failedStage"
```
bpar476, line 10: Do we also need to whitelist the `error` property?

jira-plugin/src/main/java/com/atlassian/migration/datacenter/configuration/MigrationAssistantOsgiImportConfiguration.java
tarka: Apparently not. It works without it.

jira-plugin/src/main/resources/whitelist/analytics_whitelist.json
tarka: Good point. I should add a note about updating the whitelist to the event objects.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
```diff
@@ -130,10 +136,19 @@ public MigrationReadyStatus getReadyStatus()
...
+
+        MigrationStage failStage = context.getMigration().getStage();
+        Long now = System.currentTimeMillis() / 1000L;
+
+        setCurrentStage(migration, ERROR);
         context.setErrorMessage(message);
+        context.setEndEpoch(now);
         context.save();
+
+        eventPublisher.publish(new MigrationFailedEvent(applicationConfiguration.getPluginVersion(),
```
theghostwhoforks, line 149: The start and end time can stay in the migration object and not the migration context object?

core/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
```diff
@@ -48,21 +51,24 @@
...
         this.jiraHome = jiraHome;
+        this.eventPublisher = eventPublisher;
     }
 
     @Override
     public Migration createMigration() throws MigrationAlreadyExistsException
     {
         Migration migration = findFirstOrCreateMigration();
         if (migration.getStage().equals(NOT_STARTED)) {
+            eventPublisher.publish(new MigrationCreatedEvent(applicationConfiguration.getPluginVersion()));
```
theghostwhoforks, line 71: Can this go into the findFirstOrCreateMethod instead? i.e. inside the create flow

jira-e2e-tests/helpers/run-local
```diff
@@ -0,0 +1,12 @@
+#!/bin/bash -x
+
+jsrc=jira-plugin/target/jira-plugin-1.0.0.jar
+jtarget=jira-e2e-tests/jira/jira-plugin-1.0.0.jar
+
+if [[ $jsrc -nt $jtarget ]]; then
```
theghostwhoforks, line 6: Thanks! I ran into the same issue yesterday and spent some time trying to figure out why the latest code went missing!

tarka:
> Is there a story to include analytics from the state transitions? i.e. include the event publisher and publish a migration transition event?

Sure, I was thinking about that too. I mainly wanted to get the baseline in, but I can easily add those too.

theghostwhoforks:
> > Is there a story to include analytics from the state transitions? i.e. include the event publisher and publish a migration transition event?
> 
> Sure, I was thinking about that too. I mainly wanted to get the baseline in, but I can easily add those too.

Sounds good. Let's get this merged in then. I'll approve the PR now

core/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
tarka: If I add all the transitions as analytics that probably makes more sense. Will do.

bpar476:
> Is there a story to include analytics from the state transitions? i.e. include the event publisher and publish a migration transition event?

What's the intention of this event? What customer behaviour do we want to try and read from this? I'm a fan of more data but I think it's important to know why we're capturing it so that it doesn't add noise.

tarka:
> > Is there a story to include analytics from the state transitions? i.e. include the event publisher and publish a migration transition event?
> 
> What's the intention of this event? What customer behaviour do we want to try and read from this? I'm a fan of more data but I think it's important to know why we're capturing it so that it doesn't add noise.

It should allow us to see migrations that were abandoned.

PR #281, opened 2020-06-12
CHET-89: Fix analytics whitelist, and be more aggressive with container cleanup

tarka:
I hate JSON

### DIFF
--- a/jira-plugin/src/main/resources/whitelist/analytics_whitelist.json
+++ b/jira-plugin/src/main/resources/whitelist/analytics_whitelist.json
@@ -15,11 +15,11 @@
   "atl.dc.migration.transition": [
     "pluginVersion",
     "fromStage",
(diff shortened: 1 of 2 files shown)


jira-plugin/src/main/resources/whitelist/analytics_whitelist.json
```diff
@@ -15,11 +15,11 @@
   "atl.dc.migration.transition": [
     "pluginVersion",
     "fromStage",
-    "toStage",
+    "toStage"
   ],
   "atl.dc.migration.transition.failed": [
     "pluginVersion",
     "fromStage",
-    "toStage",
```
vicav, line 23: Since fromStage & toStage are strings we have to enumerate their expected values in the whitelist here, otherwise they will not come through.

PR #283, opened 2020-06-12
CHET-442: Add final_sync_wait to list of database stages to await in …

theghostwhoforks:
…final_sync component

This is required to signal end to the process in migration transfer component. As we auto-transition from FINAL_SYNC_WAIT to VALIDATE, the UI. needs to be aware that `validate` is a stage that the state can reach, when the filesync status is being polled from the UI.

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/final-sync/FinalSync.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/final-sync/FinalSync.tsx
@@ -72,8 +72,8 @@ const fsSyncStatusToProgress = (status: FinalSyncStatus): Progress => {
     if (hasProgressedToNextStage) {
         builder.setCompleteness(1);
     } else {
-        // Uploaded file count can be zero when no attachments are uploaded
-        builder.setCompleteness(uploaded === 0 ? 1 : downloaded / uploaded);
+        // If there are no files to upload (i.e. uploaded = 0), the state will be transitioned to `Validate` and conditional branch will not be evaluated due to the `hasProgressedToNextStage` check above.
+        builder.setCompleteness(uploaded === 0 ? 0 : downloaded / uploaded);
     }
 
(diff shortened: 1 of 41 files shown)


frontend/src/main/dc-migration-assistant-fe/components/final-sync/FinalSync.tsx
```diff
@@ -62,12 +64,17 @@ const dbStatusToProgress = (status: DatabaseMigrationStatusResult): Progress =>
...
     const builder = new ProgressBuilder();
     builder.setPhase(I18n.getText('atlassian.migration.datacenter.sync.fs.phase'));
     // FIXME: This time will be wrong when one of the components of final sync completes
     builder.setElapsedSeconds(db.elapsedTime.seconds);
-    builder.setCompleteness(uploaded === 0 ? 0 : downloaded / uploaded);
+    if (hasProgressedToNextStage) {
+        builder.setCompleteness(1);
+    } else {
+        // Uploaded file count can be zero when no attachments are uploaded
+        builder.setCompleteness(uploaded === 0 ? 1 : downloaded / uploaded);
```
bpar476, line 76: Wouldn't this mark the sync as complete when no attachments have been uploaded but are currently being uploaded?

theghostwhoforks: That's correct. From a timing perspective, this page should be rendered after starting a migration.

However, the point you raise is valid and I think the previous check will catch scenario when no files exist to be uploaded (i.e. uploaded will be equal to downloaded, a valid terminal case). I'll change this back to report zero completion when uploaded  count is 0

theghostwhoforks:
Here are a few screenshots of a successful migration

Screenshot 2020-06-15 at 15 27 29
Screenshot 2020-06-15 at 15 27 42

PR #284, opened 2020-06-12
Chet 439 add rest tests to pipeline

vicav:
- Trigger Rest tests from smoketest pipeline
- Introduce Jira_Version parameter in preperation of running more jira versions
- Change get /migration to return 200 instead of 404 even if migration hasn't started


.github/workflows/mvn_test.yml
```diff
@@ -9,6 +9,7 @@ jobs:
   run_tests:
     name: Java${{ matrix.java }} ${{ matrix.profile }} test
     runs-on: ubuntu-latest
+    timeout-minutes: 25
```
vicav, line 12: Seems like a good idea to give job an upper limit

.github/workflows/mvn_test.yml
```diff
@@ -61,7 +63,14 @@ jobs:
...
           ./postgres/inject-license
-          docker-compose up --force-recreate --build --exit-code-from cypress --abort-on-container-exit
+          JIRA_VERSION=latest docker-compose up --force-recreate --build -d
+
+      - name: Rest api tests with production image built by previous step
+        timeout-minutes: 15
+        if: matrix.profile == 'smoketest'
+        run: |
+          cd jira-e2e-tests
+          until docker-compose ps --services --filter "status=stopped" | grep cypress ; do sleep 5; done; mvn -f ../pom.xml verify -pl func-tests -Prest; docker-compose down
```
vicav, line 73: waiting until cypress container has been stopped. Unfortunately waiting for jira/status to return 200 proved too early to run the rest tests
$(curl -s -o /dev/null -w ''%{http_code}'' http://localhost:2990/jira/status)" != "200"; is more what I was going for, but alas, I'll keep digging to see if I can get another rest endpoint for reference

func-tests/src/test/kotlin/com/atlassian/migration/api/MigrationEndpointRestTest.kt
```diff
@@ -35,11 +51,36 @@ class MigrationEndpointRestTest : BaseRestTest() {
...
-                )
+                equalTo(Response.Status.OK.statusCode)
             )
         }
     }
+
+    @Test
+    fun `Migration can be started`() {
+        Given {
+            spec(requestSpec).auth().preemptive().basic("admin", "admin").contentType(ContentType.JSON)
```
theghostwhoforks, line 62: The credentials come from the requestspec and may not be necessary to add again

jira-plugin/pom.xml
```diff
@@ -129,6 +129,11 @@
                             <id>jira</id>
                             <version>${amps.jira.version}</version>
                             <productDataVersion>${amps.jira.version}</productDataVersion>
+                            <jvmArgs>
```
vicav, line 132: some useful flags, in particular  -Datlassian.darkfeature.jira.onboarding.feature.disabled=true will skip setup

.github/workflows/mvn_test.yml
theghostwhoforks, line 73: Alternately, we could run both cypress and rest tests in parallel in 2 different containers - this would reduce the total time taken to run the full suite, correct?

jira-e2e-tests/jira/waitport
```diff
@@ -4,7 +4,7 @@ set -e
 
 host=$1
 port=$2
-secs=600
+secs=1200
```
theghostwhoforks, line 7: Curious to understand why this was necessary? The upgrade task as we use an database dump compliant with 7.13?

.github/workflows/mvn_test.yml
theghostwhoforks: maybe we could check if there's an UPM endpoint that tells us when all plugins have been loaded 🤞

vicav: Sure - I'd like to do that as a separate task if that is ok though?

func-tests/src/test/kotlin/com/atlassian/migration/api/MigrationEndpointRestTest.kt
vicav: I actually made this change due to auth error that pointed me to need of PreemptiveBasicAuthScheme

jira-e2e-tests/jira/waitport
vicav: Yes, this is to give some more room for jira startup

.github/workflows/mvn_test.yml
vicav: Yes, once I figure out how to start the rest tests once jira is ready not just up I'm happy to make this change. Currently rest tests are super fast so they don't actually add much time on. The biggest cost is building/starting jira.

func-tests/src/test/kotlin/com/atlassian/migration/api/MigrationEndpointRestTest.kt
vicav: I'll push a change to change the request spec to use that as default, that might be nicer

theghostwhoforks:
I'll merge this in @vicav 

Hopefully, I won't leave a broken master, but it's worthwhile having lesser PRs on Monday to go through

PR #285, opened 2020-06-14
CHET-298: Build and upload migration template for pushes to master

dylanrathbone:
Skipping for branches other than `master`

image

### DIFF
--- a/.github/workflows/migration_template_uploader.yml
+++ b/.github/workflows/migration_template_uploader.yml
@@ -3,14 +3,15 @@ name: UploadMigrationTemplate
 on: [push]
 
 env:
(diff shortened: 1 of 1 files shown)


.github/workflows/migration_template_uploader.yml
```diff
@@ -10,6 +10,7 @@ env:
   
 jobs:
   upload:
+    if: github.ref == 'refs/heads/master'
```
theghostwhoforks, line 13: The branch to compare to can be parameterized, like so -
https://github.com/atlassian-labs/dc-migration-assistant/blob/master/.github/workflows/mvn_test.yml#L89

PR #286, opened 2020-06-15
CHET-89: Add OS/FS/DB analytics and add some enumerations to whitelist.

tarka:


core/src/main/java/com/atlassian/migration/datacenter/core/util/AnalyticsHelper.java
```diff
@@ -0,0 +1,32 @@
+package com.atlassian.migration.datacenter.core.util;
+
+import org.apache.commons.lang3.SystemUtils;
+
+public class AnalyticsHelper
```
vicav, line 5: /nit: Not sure if AnalyticsHelper wrapper around OsType enum is needed. I'd expect simply the OsType enum with the method getSystemOsType()

core/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
```diff
@@ -117,12 +119,12 @@ public synchronized void transition(MigrationStage to) throws InvalidMigrationSt
 
         if (!currentStage.isValidTransition(to)) {
             eventPublisher.publish(new MigrationTransitionFailedEvent(applicationConfiguration.getPluginVersion(),
-                                                                      currentStage.toString(), to.toString()));
+                                                                      currentStage, to));
```
theghostwhoforks, line 122: I'm wondering if we should push this into the `transition` method within the statemachine? That way the publish logic will be controlled as a state transitions without the services calling the transition having to explicitly invoke the publish method?

or worst case, we just call transition with an optional/nullable `eventPublisher` callback

core/src/main/java/com/atlassian/migration/datacenter/core/util/AnalyticsHelper.java
tarka: Good point, done.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/AWSMigrationService.java
tarka: Unfortunately the AWSMigrationService is the closest thing we have to a state machine currently.

PR #288, opened 2020-06-15
CHET-502: Construct stub infrastructure status object when migration …

theghostwhoforks:
…state is inbetween stages i.e. provision application and migration stages

The main issue is that the API returns a response that looks like this - 

`{"status":"PREPARING_MIGRATION_INFRASTRUCTURE_DEPLOYMENT"}`

while the FE deserializes into an object that looks like this -

```
type StackStatusResponse = {
    status: {
        state: 'CREATE_IN_PROGRESS' | 'CREATE_COMPLETE' | 'CREATE_FAILED' | 'PREPARING_MIGRATION_INFRASTRUCTURE_DEPLOYMENT';
        reason: string;
    };
    phase?: 'app_infra' | 'migration_infra';
};
```

The FE expects status to be a `dict` and not a `string`. We catch errors and redirect a user to the error page, which we reach in this case due to not being able to deserialize the response


api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/ProvisioningStatusResponse.kt
```diff
@@ -0,0 +1,23 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.atlassian.migration.datacenter.api.aws
+
+import com.fasterxml.jackson.annotation.JsonProperty
+
+class ProvisioningStatusResponse(@get:JsonProperty("state") val state: String){
```
theghostwhoforks, line 20: @bpar476 - this is the Status response holder class that we spoke about offline. Feel free to use/rewrite/discard this implementation to accommodate the status API changes.

Note - Status is a string only because `PREPARING_MIGRATION_INFRASTRUCTURE_DEPLOYMENT` does not exist in the `InfrastructureDeploymentStatus` enum

theghostwhoforks:
In process of verifying. Will merge soon after getting a successful verification

PR #290, opened 2020-06-16
CHET-495: remove reason from infrastructure status

bpar476:
## Summary

* Delete InfrastructureDeploymentStatus structure
* Simply use enum to communicate stack deployments status

This is a refactor until the stack deployment failure error retrieval is implemented fully.


bpar476:
> I think we need to change the class `ProvisioningStatusResponse` that contains a nested state structure within `status` (i.e. reflect this change)
> 
> See this test that has to change - `com.atlassian.migration.datacenter.api.aws.CloudFormationEndpointTest#shouldReturnIntermediatePhaseWhileBetweenDeployments`

I've just deleted that structure as it's not really necessary anymore.

theghostwhoforks:
> > I think we need to change the class `ProvisioningStatusResponse` that contains a nested state structure within `status` (i.e. reflect this change)
> > See this test that has to change - `com.atlassian.migration.datacenter.api.aws.CloudFormationEndpointTest#shouldReturnIntermediatePhaseWhileBetweenDeployments`
> 
> I've just deleted that structure as it's not really necessary anymore.

Yup - Looks good now

PR #292, opened 2020-06-16
CHET-163: Add filesystem filter.

tarka:


core/src/main/java/com/atlassian/migration/datacenter/core/fs/DirectoryStreamCrawler.java
```diff
@@ -28,10 +28,14 @@
 import java.nio.file.Files;
 import java.nio.file.NoSuchFileException;
 import java.nio.file.Path;
+import java.util.regex.Pattern;
 
 public class DirectoryStreamCrawler implements Crawler {
     private static final Logger logger = LoggerFactory.getLogger(DirectoryStreamCrawler.class);
 
+    private static final String ignorePattern = "^(analytics-logs|caches|import|plugins/.bundled_plugins|plugins/.osgi-plugins)";
```
theghostwhoforks, line 36: What about dbconfig.xml?

And do we need to copy `analytics-logs`?

core/src/main/java/com/atlassian/migration/datacenter/core/fs/DirectoryStreamCrawler.java
```diff
@@ -61,12 +65,19 @@ public void crawlDirectory(Path start, UploadQueue<Path> queue) throws IOExcepti
         }
     }
 
-    private void listDirectories(UploadQueue<Path> queue, DirectoryStream<Path> paths) {
+    private void listDirectories(Path base, UploadQueue<Path> queue, DirectoryStream<Path> paths) {
+        int off = base.getNameCount();
         paths.forEach(p -> {
+            // NOTE: This should be possible with the directoryStream regex glob,
+            // but in practice doesn't play well with nested directory patterns.
+            String subpath = p.subpath(off, p.getNameCount()).toString();
```
theghostwhoforks, line 73: So, for `plugins/.osgi_-plugins`, won't the subpath be just `.osgi-plugins`?

tarka: No, as we save the length of the base-path at the start and alway use that. There's a test for that case.

core/src/main/java/com/atlassian/migration/datacenter/core/fs/DirectoryStreamCrawler.java
tarka: Yeah, we should probably remove dbconfig. I'm not sure about analytics, it's that just a log rather than state?

tarka: I've removed analytics-logs just in case.

PR #293, opened 2020-06-17
CHET-495: detailed stack errors

bpar476:
# Summary

* Guess the root cause of the stack deployment failure
* We get the earliest error for the stack
* If the error is for a nested stack, we look for the earliest error in that stack until we hit an error that is not for a nested stack


core/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
```diff
@@ -119,6 +125,58 @@ public InfrastructureDeploymentState getStatus(String stackName) {
...
+                            CREATE_FAILED.equals(stackEvent.resourceStatus()))
+                    .collect(Collectors.toList());
+
+            if (failedEvents.isEmpty()) {
+                return Optional.empty();
+            }
+
+            StackEvent earliestEvent = getEarliestEventFromStackEvents(failedEvents);
+
+            if (isEmbeddedStackError(earliestEvent)) {
```
theghostwhoforks, line 152: So, If a root stack (containing 3 nested stacks) were to fail and the 3rd stack had the failure reason is present in the 2nd stack and not the first one, then won't this method an empty optional?

theghostwhoforks: Note - I ran the tests and verified that the case that I have in mind won't happen as the failing stack physical ID is used to perform failed event lookup.

Not an issue. Approved

PR #294, opened 2020-06-17
CHET-511: Remove the FS size check due to time on large installs.

tarka:


frontend/src/main/dc-migration-assistant-fe/components/Home.tsx
```diff
@@ -131,9 +131,6 @@ export const ReadyStatus: FunctionComponent = () => {
                 <li>
                     is running on <strong>Linux</strong> &nbsp; {readyString(ready?.osCompatible)}
                 </li>
-                <li>
-                    has a home directory under 400GB &nbsp; {readyString(ready?.fsSizeCompatible)}
```
bpar476, line 135: I think maybe we should display this prerequisite anyway but say that we can't verify this for them 🤔 What do you think @ddomingorh @ashking2

tarka: I've created https://aws-partner.atlassian.net/browse/CHET-513 for @ddomingorh or @ashking2  to write new copy for this.

PR #296, opened 2020-06-17
Chet 494 retry provisioning

theghostwhoforks:


api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.kt
```diff
@@ -102,5 +102,21 @@ class CloudFormationEndpoint(private val deploymentService: ApplicationDeploymen
...
+    @POST
+    @Path("/reset")
+    @Consumes(MediaType.APPLICATION_JSON)
+    @Produces(MediaType.APPLICATION_JSON)
+    fun resetProvisioningStage(): Response {
+        return when (val currentStage = migrationService.currentStage) {
+            MigrationStage.ERROR, //Only until we change transitioning to errors in provisioning steps
+            MigrationStage.PROVISIONING_ERROR -> {
+                migrationService.transition(MigrationStage.PROVISION_APPLICATION)
+                Response.accepted().build()
```
bpar476, line 114: Should this be `ok` instead of `accepted`? There isn't any asynchronous work running after this request returns.

api/src/test/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpointTest.kt
```diff
@@ -57,6 +57,24 @@ class MigrationEndpointTest {
         assertThat(response.entity.toString(), Matchers.containsString(MigrationStage.AUTHENTICATION.toString()))
     }
 
+    @Test
+    fun shouldBeErrorWhenStageIsASpecificErrorStage() {
+        every { migrationService.currentStage } returns MigrationStage.FINAL_SYNC_ERROR
+
+        val response = sut.getMigrationStatus()
+
+        assertThat(response.entity.toString(), Matchers.containsString(MigrationStage.ERROR.toString()))
```
bpar476, line 66: Why do we want to obfuscate the specific error stage in the API?

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
```diff
@@ -19,56 +19,83 @@ package com.atlassian.migration.datacenter.spi
  * Represents all possible states of an on-premise to cloud migration.
  */
 enum class MigrationStage {
-    NOT_STARTED,
+    NOT_STARTED(),
     AUTHENTICATION(NOT_STARTED),
+
     PROVISION_APPLICATION(AUTHENTICATION),
```
bpar476, line 25: Does this need to be valid from `PROVISIONING_ERROR` as well?

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
```diff
@@ -19,56 +19,83 @@ package com.atlassian.migration.datacenter.spi
...
+        }
     }
 
     fun isAfter(stage: MigrationStage): Boolean {
         if (this == stage || this == NOT_STARTED) return false
         if (this == FINISHED || this == ERROR) return true
-        this.validFrom!!.let { it: MigrationStage ->
-            return if (it == stage) true else it.isAfter(stage)
-        }
+        return this.validFromStages.map {
```
bpar476, line 82: can this not just be `validFromStages.contains(stage)`

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
```diff
@@ -19,56 +19,83 @@ package com.atlassian.migration.datacenter.spi
...
+        }.contains(true)
     }
 
     // Hacky, but OK for now.
     val isDBPhase: Boolean
         get() =// Hacky, but OK for now.
             this.toString().startsWith("DB_")
 
+    fun isErrorStage(): Boolean {
+        return listOf(ERROR, FS_MIGRATION_ERROR, PROVISIONING_ERROR, FINAL_SYNC_ERROR).contains(this)
```
bpar476, line 96: This is a very minor thing but would this be better as
```
return when(this) {
     ERROR, FS_MIGRATION_ERROR, PROVISIONING_ERROR, FINAL_SYNC_ERROR -> true
     else -> false
}
```
This way we don't need to construct a list object or iterate over it

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
```diff
@@ -19,56 +19,83 @@ package com.atlassian.migration.datacenter.spi
...
+    constructor(vararg validFromStages: MigrationStage) {
+        exception = null;
+        this.validFromStages = validFromStages.asList()
     }
 
+
     fun isValidTransition(to: MigrationStage): Boolean {
-        return to.validFrom?.equals(this) ?: true
+        when (this) {
+            PROVISIONING_ERROR -> return to == PROVISION_APPLICATION
```
bpar476, line 66: If we add `PROVISIONING_ERROR` to the `validFromStages` of `PROVISION_APPLICATION` we don't need to have this edge case.

api/src/test/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpointTest.kt
theghostwhoforks: This was done just to maintain compatibility with the currrent UI.

When we play the UI story, I'll relook at the stage specific errors and see if they need to be handled

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
theghostwhoforks: That causes a circular dependency when declaring both `PROVISIONING_ERROR(PROVISION_APPLICATION)` and `PROVISIONING_APPLICATION(PROVISIONING_ERROR)`

We need to pick either or :(

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
theghostwhoforks: In this case, yes. But the recursion was easier to model in terms of nesting in each element of the `validFromStages` list

  A
 |-|
B  C
----|
      D

Here, with recursion D.isAfter(A) will evaluate to true as we check C's ancestors.

With `validFromStages.contains(stage)`, we would stop at `C` and not traverse to `A`.

Does this make sense of have I misunderstood something?

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
theghostwhoforks: see below

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
bpar476: Retrying is inherently a cycle and if you were to draw the state machine as a graph there would indeed be a cycle between PROVISION_APPLICATION to PROVISIONING_ERROR.

Where exactly does the circular dependency cause an error. Is it just the `isAfter` functionality?

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
bpar476: Oh I understand. Contains would not find transitive "afters"

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
bpar476: I don't think it would be infinite recursion, I think it would just say that PROVISION_APPLICATION and PROVISIONING_ERROR are both after each other. Which is... kind of true 🙃

bpar476: Oh wait, it would cause an actual syntax error. I wonder if we could do this with method extensions after all states are declared 🤔

theghostwhoforks: :) yeah. The forward reference error with enums makes this annoying.

We can take a stab at the method extensions later after this PR is merged?

bpar476: Sure, sounds good.

PR #298, opened 2020-06-17
CHET-512: add detailed error to stack status endpoint

bpar476:
# Summary
* Refactor cloudformation endpoint to be less spaghetti
* Add fetching of "root cause" error of cloudformation deployment to endpoint
* Add link to cloudformation console in status response
* Begin integration with frontend

## Examples
Request output:
image
Corresponding Console:
image


api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.kt
```diff
@@ -68,39 +82,67 @@ class CloudFormationEndpoint(private val deploymentService: ApplicationDeploymen
...
+                MigrationStage.PROVISION_MIGRATION_STACK_WAIT -> helperDeploymentService.deploymentStatus
+                else -> error("phase conditional doesn't match when cases")
+            }
+            val entity = mutableMapOf(
+                    "status" to deploymentStatus,
+                    "phase" to phase
+            )
+
+            if (deploymentStatus == InfrastructureDeploymentState.CREATE_FAILED) {
+                val deploymentId = when (currentMigrationStage) {
```
theghostwhoforks, line 120: When the status in CFn is `InfrastructureDeploymentState.CREATE_FAILED`, won't the statemachine have transitioned the status (i.e. `currentMigrationStage`) to `Error`. In which case, we will always enter the else flow here. Correct?

bpar476: Not necessarily. The poll rate of the UI is faster than the poll rate of the backend so when this endpoint calls `DeploymentService#getDeploymentStatus()`, it may be before the backend has polled and realised that the deployment has errored. In that case, it will transition the state machine to error and return `CREATE_FAILED`. i.e. the transition may occur after the request is accepted for processing.

See `com.atlassian.migration.datacenter.core.aws.infrastructure.CloudformationDeploymentService.getDeploymentStatus()`

theghostwhoforks: Yeah - this addresses the original comment. 👍 

I think the issue with that method is that a GET call changes the state of the system, breaching the idempotence requirement of RESTful GET methods. I wonder if we should just mirror the CFn API status in the database, and refer to just a single source while looking up the status. This will make the `CloudforrmationDeploymentService` stateless by removing the `deploymentWatcher` scheduledFuture reference from the service. 

Anyway, this needs to be tackled separately.

api/src/test/kotlin/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpointTest.kt
```diff
@@ -83,8 +103,8 @@ internal class CloudFormationEndpointTest {
         val errorMessage = "migration status is FUBAR"
```
theghostwhoforks, line 103: I approve of this error message

theghostwhoforks:
Will merge this in. Will make a related PR easier to merge

api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.kt
bpar476: Hmm, that's a good point re: changing system in GET.

PR #299, opened 2020-06-18
CHET-254/219/296 Render errors when deploying from quickstart form and fix form clearing after submit.

bpar476:
# Summary
* Throw exceptions from deployment all the way to caller
* Use exceptions to get meaningful error message when deployment fails
* Render deployment error in flag in UI when quickstart form is rejected

image

Bonus: I've fixed the issue where the select inputs in the quickstart for were being cleared. This is because we were modifying the form data before submitting it as it was only copied by _reference_


core/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
```diff
@@ -203,13 +204,16 @@ private StackEvent getEarliestEventFromStackEvents(List<StackEvent> failedEvents
 
             if (!response.sdkHttpResponse().isSuccessful()) {
                 logger.error("create stack {} http response failed with reason: {}", stackName, response.sdkHttpResponse().statusText());
-                return Optional.empty();
+                throw new InfrastructureDeploymentError(response.sdkHttpResponse().statusText().isPresent() ? response.sdkHttpResponse().statusText().get() : "Stack creation failed for unknown reason");
```
bpar476, line 207: We use a runtime exception so that the `DeploymentService` interface is not dirtied with exceptions

bpar476:
@ddomingorh What do you think of the copy? The heading is defined by us but the content is the raw error

core/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
theghostwhoforks: Maybe it's worthwhile making in a checked exception adding it to the contract? It's not unreasonable to see `InfrastructureDeploymentError` in the contract.

Also, in the docs, we see @Throws InfrastructureDeploymentError, but the error is not declared in the method definition, which could be confusing

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
```diff
@@ -93,6 +94,16 @@ public void deployApplication(@NotNull String deploymentId, @NotNull Map<String,
...
 
+    /**
+     * Commences the deployment of the AWS Quick Start. It will transition the state machine upon completion of the
+     * deployment. If the deployment finishes successfully we transition to the next stage, otherwise we transition
+     * to an error. The migration will also transition to an error if the deployment takes longer than an hour.
+     *
+     * @param deploymentId the stack name
+     * @param params       the parameters for the cloudformation template. The key should be the parameter name and the value
+     *                     should be the parameter value.
+     * @throws InfrastructureDeploymentError when quickstart fails to deploy
```
theghostwhoforks, line 105: This is what I mean by the doc comment above. It's defined in the doc, but not in the method

core/src/main/java/com/atlassian/migration/datacenter/core/aws/CfnApi.java
bpar476: I'll do this as a follow-up

PR #302, opened 2020-06-18
CHET-496: render new error message and url in stack status

bpar476:
# Summary
* Modify provisioning api util to return more error data
* Modify quickstart status page to include additional error data
* Stop "handling" a failed deployment when the UI finds out before the backend


jira-plugin/pom.xml
```diff
@@ -133,6 +133,7 @@
                                 -XX:+UseG1GC -Datlassian.darkfeature.jira.onboarding.feature.disabled=true
                                 -Datlassian.mail.senddisabled=false -Djira.startup.warnings.disable=true
                                 -Dtroubleshooting.dev.mode=true -Xmx1536m -Xms512m -XX:ReservedCodeCacheSize=256m -XX:MaxMetaspaceSize=512m
+                                -Dspring.profiles.active=${amps.spring.profiles}
```
theghostwhoforks, line 136: Thanks for this. I ran into this yesterday and had the same addition in my local after debugging :)

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/CloudformationDeploymentService.java
```diff
@@ -79,14 +79,6 @@ protected InfrastructureDeploymentState getDeploymentStatus(String stackName) {
         requireNonNull(stackName);
         InfrastructureDeploymentState status = cfnApi.getStatus(stackName);
 
-        if (isFailedToCreateDeploymentState(status)) {
```
bpar476, line 82: This was causing issues because every request to get the status of the deployment would hit the DB and as @theghostwhoforks pointed out, violates the GET contract.

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
```diff
@@ -58,7 +81,7 @@ const getDeploymentProgress: ProgressCallback = async () => {
             builder.setPhase(
                 I18n.getText('atlassian.migration.datacenter.provision.aws.status.error')
             );
-            builder.setError(err.message);
+            builder.setError(JSON.stringify(err));
```
bpar476, line 84: We handle the error like this in case the error does not have a `message` property

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPage.tsx
```diff
@@ -30,7 +30,7 @@ import { migrationErrorPath } from '../../utils/RoutePaths';
 import { CommandDetails as CommandResult } from '../../api/final-sync';
 import { MigrationErrorSection } from './MigrationErrorSection';
 
-const POLL_INTERVAL_MILLIS = 3000;
+const POLL_INTERVAL_MILLIS = 8000;
```
bpar476, line 33: I increased this because the requests started overlapping

spi/src/main/kotlin/com/atlassian/migration/datacenter/dto/MigrationContext.kt
```diff
@@ -17,13 +17,18 @@ package com.atlassian.migration.datacenter.dto
...
 
 interface MigrationContext : Entity {
     var migration: Migration
     var applicationDeploymentId: String
     var helperStackDeploymentId: String
     var serviceUrl: String
-    var errorMessage: String
+
+    // We use methods for these properties because the annotation cannot target a var
+    @StringLength(value = 450)
```
bpar476, line 29: The AWS error messages can be longer than the default 255 characters. 450 is the maximum length we can apply that isn't unbounded.

nanux: What type of column will be created in different databases? Not really important just I believe postgres supports more than 450 in varchar.

api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
```diff
@@ -113,7 +113,7 @@ class MigrationEndpoint(private val migrationService: MigrationService) {
 
         return mapOf(
             "instanceUrl" to currentContext.serviceUrl,
-            "error" to currentContext.errorMessage
+            "error" to currentContext.getErrorMessage()
```
nanux, line 116: Just from curiosity - Kotlin by default access the properites via `getProperty()`. This didn't work because it is Interface?

spi/src/main/kotlin/com/atlassian/migration/datacenter/dto/MigrationContext.kt
bpar476: 450 is the maximum supported by AO other than unlimited.

api/src/main/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpoint.kt
bpar476: I think it doesn't work because it's explicitly declared as a method in a `.kt` interface. It needs to be a method because otherwise the annotation won't work.

spi/src/main/kotlin/com/atlassian/migration/datacenter/dto/MigrationContext.kt
theghostwhoforks: I just saw this in my postgres database (9.6) and thought I'd share it here -

Screenshot 2020-06-22 at 15 10 12

The type of column generated for error message is `character varying(255)`, i.e. 255 character upper limit. It seems that the `StringLength` annotation is not taking effect.

Here's an example - 

The phrase `Thequickbrownfoxjumpsoverthelazydog` has 35 characters. 7 repetitions of the phrase can be stored, while 8 cannot. See -

Screenshot 2020-06-22 at 15 16 37

PR #304, opened 2020-06-19
Chet 520 configure again

theghostwhoforks:
Create a retry checkbox and allow users to retry a migration phase when the stage is retryable

Screenshots - 
Checkbox to retry 

Screenshot 2020-06-22 at 10 00 08

Allowed to retry 

Screenshot 2020-06-22 at 10 00 14

On retry click, redirect to phase start 

Screenshot 2020-06-22 at 10 00 20

On Jira, the retry screen looks like (The error message is empty, set up for simulate a failure)

Screenshot 2020-06-22 at 14 44 02


frontend/src/main/dc-migration-assistant-fe/api/provisioning.ts
```diff
@@ -114,7 +114,7 @@ export const provisioning = {
                     }
                 }
                 if (resp.error) {
-                    // hanle error
+                    // handle error
```
bpar476, line 117: Nice

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
```diff
@@ -77,6 +78,10 @@ type DeploymentMode = {
     deploymentMode: string;
 };
 
+const retryProvisioningCallback: RetryCallback = () => {
```
bpar476, line 81: Lately, I've been trying to abstract the REST stuff from anything interested in it and instead just build it into an object that encompasses all the methods for that resource, see the `provisioning` symbol in this file. Is there a reason you prefer calling `callAppRest` directly?

theghostwhoforks:
the `retryText` string here is used as a conditional to retry or redirect. This is only temporary and when all stages are retryable, this will be removed

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPage.tsx
```diff
@@ -123,13 +131,18 @@ export const MigrationTransferPage: FunctionComponent<MigrationTransferProps> =
...
+    onRetryRoute,
 }) => {
     const [progressList, setProgressList] = useState<Array<Progress>>([]);
     const [loading, setLoading] = useState<boolean>(true);
     const [progressFetchingError, setProgressFetchingError] = useState<string>();
     const [started, setStarted] = useState<boolean>(false);
     const [finished, setFinished] = useState<boolean>(false);
+    const [failed, setFailed] = useState<boolean>(false);
     const [commandResult, setCommandResult] = useState<CommandResult>();
+    const errorStages = [MigrationStage.ERROR];
```
bpar476, line 145: Hmm, I've put `ERROR` as an in progress stage for the ProvisioningStatus page so that if they refresh the page we still see the inline error message instead of being redirected to the error page 🤔 What's the purpose of checking for current stage being an error stage?

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPage.tsx
```diff
@@ -269,6 +291,10 @@ export const MigrationTransferPage: FunctionComponent<MigrationTransferProps> =
                             onRefresh={updateProgress}
                             started={started}
                             loading={loading}
+                            failed={failed}
```
bpar476, line 294: Hmm, this component is getting very bloated. I think we should break it down so we have different components for when it's finished or unstarted or failed and render conditionally in this component. However, that's probbaly out of scope for this story. Where can we track this refactor, is a ticket best?

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
theghostwhoforks: Yup - moving to provisioning now

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPage.tsx
theghostwhoforks: I wanted a place to setFailed to true. This scenario happens when someone lands on the status page after firing a deployment, closing the tab, reopens the tab.

The update progress call (before your PR adds it as a progress stage), will not be fired and we needed to set the failed flag to true, when we get an error stage response.

I'll rebase with master and refactor this

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPage.tsx
theghostwhoforks: It is - it got a bit tricky to handle different cases also (i.e. in-progress v/s error v/s retry ..)

Composing this will be helpful - perhaps as a subtask when we play the last story in this series?

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
theghostwhoforks: Done

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPage.tsx
theghostwhoforks: setFailed is not set in the update callback, which is now called even when the stage is `error`

When this is set and the stage is retryable, the reset button is now rendred

PR #308, opened 2020-06-22
CHET-523: Drop existing DB objects and re-create if they exist

dylanrathbone:
Changes made:

-  Updated `docker-compose` db name to align it with what Jira actually creates under the hood. 
- Update `pg_restore` function to use `-c --if-exists` flag. This will re-create those objects within the `db_dump` only if they exist within the target db created by Jira.

This changes have been tested manually and after a re-index work. Before merging I will perform an actual migration pointing to a test environment


bpar476:
to be clear, this means any `CREATE TABLE X` will be replaced with `DROP TABLE X IF EXISTS; CREATE TABLE X`?

theghostwhoforks:
> to be clear, this means any `CREATE TABLE X` will be replaced with `DROP TABLE X IF EXISTS; CREATE TABLE X`?

My understanding of this is `-c` or `--clean` along with `--if-exists` drops database objects in the destination, if they already exist. So what you are saying should be happening. 

@dylanrathbone to confirm after running this as part of a migration

theghostwhoforks:
My recommendation is to merrge this ASAP so that we can test e2e migrations, if required.

Temporarily, we can manually restart the instance to see a working Jira instance. Let's make the restart automation happen in a separate PR

dylanrathbone:
> > to be clear, this means any `CREATE TABLE X` will be replaced with `DROP TABLE X IF EXISTS; CREATE TABLE X`?
> 
> My understanding of this is `-c` or `--clean` along with `--if-exists` drops database objects in the destination, if they already exist. So what you are saying should be happening.
> 
> @dylanrathbone to confirm after running this as part of a migration
> 
> source: https://www.postgresql.org/docs/9.4/app-pgrestore.html

Correct

PR #311, opened 2020-06-22
CHET-528: Add retry transition for final sync

bpar476:
## Summary

* Refactor migration stage so that retry transitions are not special cases
* Implement cycle detection in is after
* Add transition from FINAL_SYNC_ERROR to DB_MIGRATION_EXPORT and FINAL_SYNC_WAIT

### DIFF
--- a/spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
+++ b/spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
@@ -19,47 +19,78 @@ package com.atlassian.migration.datacenter.spi
  * Represents all possible states of an on-premise to cloud migration.
  */
 enum class MigrationStage {
-    NOT_STARTED(),
-    AUTHENTICATION(NOT_STARTED),
+    NOT_STARTED {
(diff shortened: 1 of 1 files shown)


spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
```diff
@@ -19,47 +19,79 @@ package com.atlassian.migration.datacenter.spi
...
+    FS_MIGRATION_ERROR {
+        override val validAncestorStages = setOf(FS_MIGRATION_COPY, FS_MIGRATION_COPY_WAIT)
+    },
+    OFFLINE_WARNING {
+        override val validAncestorStages = setOf(FS_MIGRATION_COPY_WAIT)
+    },
+
+    DB_MIGRATION_EXPORT {
+        override val validAncestorStages
+            get() = setOf(OFFLINE_WARNING, FINAL_SYNC_ERROR)
```
theghostwhoforks, line 58: Is there a reason for a property `getter` here and not in the other stages?

bpar476: The getter allows the forward reference

PR #312, opened 2020-06-22
CHET-503: Allow SSM API execution without upper boundary

nanux:
This prevents SSMApi consumer to fail after 100 seconds even when the command is still running.


nanux:
Looking into the build failures - running it locally.

PR #313, opened 2020-06-23
CHET-526: independent final sync processes from API

bpar476:
## Summary

* New endpoints: `/migration/final-sync/retry/{db,fs}`
* Only does retry if state is `FINAL_SYNC_ERROR`
* Each will move the state machine to the starting state for the respective process then start the process.

### DIFF
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpoint.kt
@@ -93,6 +93,7 @@ class FinalSyncEndpoint(
     fun retryFsSync(): Response {
         return when (migrationService.currentStage) {
             MigrationStage.FINAL_SYNC_ERROR -> {
+                migrationService.transition(MigrationStage.FINAL_SYNC_WAIT)
                 Response.status(if (finalSyncService.scheduleSync()) Response.Status.ACCEPTED else Response.Status.CONFLICT)
             }
             else -> Response.status(Response.Status.BAD_REQUEST)
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpoint.kt
@@ -89,7 +89,6 @@ class FinalSyncEndpoint(
     }
 
     @PUT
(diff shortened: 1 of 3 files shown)


api/src/main/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpoint.kt
```diff
@@ -39,6 +39,8 @@ import javax.ws.rs.Produces
 import javax.ws.rs.core.MediaType
 import javax.ws.rs.core.Response
 
+typealias MigrationOperation = () -> Boolean
```
theghostwhoforks, line 42: Why is this preferable over `Supplier<Boolean>`?

api/src/main/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpoint.kt
```diff
@@ -88,6 +90,18 @@ class FinalSyncEndpoint(
                 .build()
     }
 
+    @PUT
+    @Path("/retry/fs")
+    fun retryFsSync(): Response {
+        return retryMigrationOperation(finalSyncService::scheduleSync, MigrationStage.FINAL_SYNC_WAIT)
```
theghostwhoforks, line 96: Should we attempt to remove an existing job, if it exists before retrying?

theghostwhoforks:
Is the idea that the FE will invoke both retry operations on retry action? Or will it detect the failed step and trigger just the step that has failed?

api/src/main/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpoint.kt
bpar476: Probably a good idea

api/src/main/kotlin/com/atlassian/migration/datacenter/api/FinalSyncEndpoint.kt
bpar476: This is the kotlin-y way to do it. `Supplier` is just the java workaround for not having functions as first class objects

bpar476:
> Is the idea that the FE will invoke both retry operations on retry action? Or will it detect the failed step and trigger just the step that has failed?

The idea is that each process can be retried from the frontend completely independently. Each section of the migration can have it's own retry button. I will reach out to Ashlee and chat about a design.

PR #315, opened 2020-06-23
CHET-523: Restart Jira instance node after DB restore

dylanrathbone:
Logs confirm restart is occurring

```
2020-06-24 07:23:19,914+0000 Caesium-1-1 INFO ServiceRunner     [c.a.m.d.c.aws.infrastructure.RemoteInstanceCommandRunnerService] Restarting Jira instance [i-0c3198e785a3866dc] now...
2020-06-24 07:23:19,916+0000 Caesium-1-1 DEBUG ServiceRunner     [c.a.m.d.c.aws.ssm.SSMApi] running document AWS-RunShellScript against instance i-0c3198e785a3866dc with parameters [commands=[sudo systemctl restart jira]]
```

Restarted node node comes up successfully. Once a re-index is performed:
- We see all the issues as expected
- DC license carried over as expected


core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/SsmPsqlDatabaseRestoreService.java
```diff
@@ -78,7 +81,13 @@ public void restoreDatabase()
 
         SuccessfulSSMCommandConsumer consumer = new EnsureSuccessfulSSMCommandConsumer(ssm, commandId,
                 migrationInstanceId);
-
+        
+        try {
+            applicationRestartService.restartJiraService();
+        } catch (JiraRestartFailure jiraRestartFailure) {
+            throw new JiraRestartFailure("Error when trying to restart Jira service", jiraRestartFailure);
```
theghostwhoforks, line 88: should we fail the migration or give the user an option to restart manually? Retry a migration because of a failed restart command seems excessive

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/ApplicationRestartService.java
```diff
@@ -0,0 +1,57 @@
...
+import software.amazon.awssdk.services.ec2.Ec2Client;
+import software.amazon.awssdk.services.ec2.model.DescribeInstancesResponse;
+import software.amazon.awssdk.services.ec2.model.Filter;
+import software.amazon.awssdk.services.ec2.model.Instance;
+import software.amazon.awssdk.services.ec2.model.Reservation;
+
+import java.util.Collections;
+import java.util.function.Supplier;
+
+public class ApplicationRestartService {
```
theghostwhoforks, line 18: We should call this `RemoteInstanceCommandRunnerService` or something like that to make it clear that this runs restart on a remote instance

core/src/test/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationServiceIT.java
```diff
@@ -98,6 +99,8 @@
     private MigrationRunner migrationRunner;
     @Mock
     AWSMigrationHelperDeploymentService migrationHelperDeploymentService;
+    @Mock
```
theghostwhoforks, line 102: We could make the test more specific to expect the specific command to mock?

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/ApplicationRestartService.java
```diff
@@ -0,0 +1,57 @@
...
+            for(Reservation reservation : response.reservations()) {
+                for(Instance instance : reservation.instances()) {
+                    try {
+                        ssm.runSSMDocument("AWS-RunShellScript", instance.instanceId(),
+                                ImmutableMap.of("commands", Collections.singletonList("sudo systemctl restart jira")));
+                    } catch (S3SyncFileSystemDownloader.CannotLaunchCommandException e) {
+                        throw new JiraRestartFailure("unable to restart Jira instance using AWS-RunShellScript SSM playbook");
+                    }
+                }
+            }
```
dylanrathbone, line 54: This will be cleaned up

core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/SsmPsqlDatabaseRestoreService.java
dylanrathbone: Yes, I think we just log the failure. Code updated.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/RemoteInstanceCommandRunnerService.java
```diff
@@ -0,0 +1,70 @@
...
+    public RemoteInstanceCommandRunnerService(SSMApi ssm, MigrationService migrationService, Supplier<Ec2Client> ec2ClientSupplier) {
+        this.ssm = ssm;
+        this.migrationService = migrationService;
+        this.ec2ClientSupplier = ec2ClientSupplier;
+    }
+    
+    public void restartJiraService() {
+        String stackName = migrationService.getCurrentContext().getApplicationDeploymentId();
+        Ec2Client ec2Client = ec2ClientSupplier.get();
+        DescribeInstancesResponse ec2InstanceMetaData = describeInstances(stackName, ec2Client);
```
theghostwhoforks, line 37: Do we have to guard for more exceptions that these 2 methods would raise? I notice that we do not have a try catch when we invoke this method; doing this could throw an unhandled exception and stop the db migration progress.

dylanrathbone: Was on the todo list :) I've added defensive code to address this.

PR #323, opened 2020-06-25
CHET-530: Fix startup and prerequisite check errors

tarka:


core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/DefaultDatabaseExtractorFactory.kt
```diff
@@ -0,0 +1,30 @@
...
+ * limitations under the License.
+ */
+package com.atlassian.migration.datacenter.core.db
+
+import com.atlassian.migration.datacenter.core.application.ApplicationConfiguration
+import com.atlassian.migration.datacenter.core.application.DatabaseConfiguration.DBType
+import com.atlassian.migration.datacenter.spi.exceptions.DatabaseMigrationFailure
+
+class DefaultDatabaseExtractorFactory(val config: ApplicationConfiguration) : DatabaseExtractorFactory {
+    override val extractor: DatabaseExtractor by lazy {
```
theghostwhoforks, line 23: This is nice!

core/src/main/java/com/atlassian/migration/datacenter/core/application/DatabaseConfigurationXmlElement.java
```diff
@@ -60,44 +63,38 @@ private void validateRequiredValues(String... values) throws ConfigurationReadEx
...
-            throw new ConfigurationReadException("Database configuration file has invalid values");
+            throw new ConfigurationReadException("Database configuration file has invalid or missing values");
         }
     }
 
-    public boolean isDataSourcePresent() {
-        return dbConfigXmlElement != null;
-    }
-}
+    public Optional<URI> getURI() {
```
theghostwhoforks, line 70: Does this need to be public?

core/src/main/java/com/atlassian/migration/datacenter/core/application/DatabaseConfigurationXmlElement.java
```diff
@@ -60,44 +63,38 @@ private void validateRequiredValues(String... values) throws ConfigurationReadEx
...
     }
 
-    public String getUrl() {
-        return url;
-    }
-
-    public String getUserName() {
-        return userName;
-    }
+    public Optional<DatabaseConfiguration.DBType> getDBType() {
```
theghostwhoforks, line 85: same comment here - does this need to be public?

core/src/main/java/com/atlassian/migration/datacenter/core/application/DatabaseConfigurationXmlElement.java
tarka: Not necessarily, but there is no harm in making them so, and it opens them up to use later.

PR #324, opened 2020-06-25
CHET-149: Unschedule jobs on plugin disable

theghostwhoforks:
Quick implementation of plugin disable event listener. 

Unfortunately, this does not work as expected as the scheduled job keys are based on a constant with the current migration ID appended to it.

The current migration ID is fetched from the database using `ao`. However, at the time of disable, the osgi bean is destroyed, leaving the ao reference that we inject into our services in an `IllegalState`

See: https://github.com/atlassian-labs/dc-migration-assistant/blob/master/core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/captor/S3FinalSyncService.kt#L69

The error that we see when we try to access ao in the disable event flow looks something like - 
```
java.lang.IllegalStateException: Invalid BundleContext.
        at org.apache.felix.framework.BundleContextImpl.checkValidity(BundleContextImpl.java:511)
        at org.apache.felix.framework.BundleContextImpl.getService(BundleContextImpl.java:451)
```


theghostwhoforks:
@vicav - I just confirmed that making the scheduled services `DisposableBean`s and implementing the destroy method to unschedule the job works. i.e. the `ao` bean is still present

PR #326, opened 2020-06-25
CHET-532/1: Render errors and retry message for each process in final sync

bpar476:
## Summary

* Refactor MigrationTransfer* so that the MigrationTransferProgress element renders the error message and the retry button
* Add API module for fs and db retry endpoints
* Wire up retry calls to final sync screen

## NB
There is a bug when re-scheduling the FS sync as the job hangs until it is complete, even if the deadletter queue has some elements in it. This wouldn't be a big deal but aborting the job doesn't seem to work.


core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
```diff
@@ -74,7 +74,13 @@ public FileSystemMigrationErrorReport performMigration()
         migrationService.transition(MigrationStage.DB_MIGRATION_EXPORT);
         startTime.set(Optional.of(LocalDateTime.now()));
 
-        Path pathToDatabaseFile = databaseArchivalService.archiveDatabase(tempDirectory);
+        Path pathToDatabaseFile;
```
bpar476, line 77: This is a fix I made along the way where errors during the backup e.g. not a postgres DB were not causing the migration to fail.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/captor/S3FinalSyncService.kt
```diff
@@ -50,8 +50,6 @@ class S3FinalSyncService(private val migrationRunner: MigrationRunner,
         migrationRunner.abortJobIfPresesnt(getScheduledJobId())
 
         logger.warn("Aborting running final file sync")
-
-        migrationService.error("Aborted final file sync")
```
bpar476, line 54: These had to be removed because you cannot retry the operation if aborting it sets the state to error.

bpar476: There should not be any issues with this as there is no way in the UI to abort a job unless cancelling your migration or retrying something

frontend/src/main/dc-migration-assistant-fe/components/shared/Progress.ts
```diff
@@ -45,6 +64,7 @@ export type Progress = {
     errorMessage?: ReactNode;
     completeMessage?: CompleteMessage;
     failed?: boolean;
+    retryProps: RetryProperties;
```
bpar476, line 67: This is probably not the best place for the retry properties. However, refactoring the migration transfer pages so that the retry properties are fed in directly to the component would have been a significant amount of work to get working with a multi-process migration.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/captor/S3FinalSyncService.kt
theghostwhoforks: If we don't set it to error, then the state would be in one of the Final sync states?

Should we not set it to `FINAL_SYNC_ERROR` here? I as as wont the invocation to `retryMigrationOperation(finalSyncService::scheduleSync, MigrationStage.FINAL_SYNC_WAIT)` fail if the state is not in `FINAL_SYNC_ERROR`?

bpar476: You're right. I think I never noticed this because I only ever aborted the job when the state was already error. I'll fix this now.

bpar476: Done, and confirmed that this works

PR #328, opened 2020-06-25
CHET-538: SQL to restore to pre-migration state

theghostwhoforks:
draft solution for - 

1. Cluster nodes being unreachable
2. clustered job deletion on destination instance
3. Plugin disable on destination


templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -258,6 +258,16 @@ Resources:
...
                 - "amazon-linux-extras install -y postgresql${PSQLClientVersion}"
                 - { PSQLClientVersion: !Ref PSQLVersion }
+        create_db_post_restore_script:
+          files:
+            /opt/atlassian/dc-migration-assistant/db_post_restore_script.sql:
+              content: !Sub
+                - |
+                  DELETE FROM clusternode;
+                  -- Or set this in the deployment form
+                   INSERT INTO pluginstate (pluginkey, pluginenabled) VALUES ('com.atlassian.migration.datacenter.jira-plugin','false');
```
dylanrathbone, line 268: Can you explain what is happening here?

theghostwhoforks, line 268: The plugin state stores the of a plugin, defined by a `pluginkey`. Setting this to false, reports the plugin as disabled on the UI. This is what we want to do to disable the plugin on the destination

PR #330, opened 2020-06-26
CHET-538: Refactor template src

dylanrathbone:


templates/migration-stack/pkg/migration-helper.yml.src
```diff
@@ -261,12 +261,13 @@ Resources:
         create_db_post_restore_script:
           files:
             /opt/atlassian/dc-migration-assistant/db_post_restore_script.sql:
-              content: !Sub
-                - |
+              content:
```
nanux, line 264: We don't need `!Sub` because we don't do variable replacement, correct?

dylanrathbone: correct

PR #333, opened 2020-06-26
CHET-535: Prefill application version and disable the field

nanux:
* added `GET` `/application/properties` endpoint - currently returning just the application version (already in Postman)
* set the version as a default for the `JiraVersion` field
* make the field `disabled` so customers will always deploy the same version as they are currently running

### DIFF
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/application/ApplicationPropertiesEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/application/ApplicationPropertiesEndpoint.kt
@@ -17,9 +17,8 @@
 package com.atlassian.migration.datacenter.api.application
 
 import com.atlassian.migration.datacenter.core.application.ApplicationConfiguration
-import com.fasterxml.jackson.annotation.JsonAutoDetect
-import com.fasterxml.jackson.annotation.PropertyAccessor
 import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.module.kotlin.registerKotlinModule
 import org.slf4j.LoggerFactory
 import javax.ws.rs.GET
 import javax.ws.rs.Path
@@ -29,7 +28,7 @@ import javax.ws.rs.core.Response
 
 @Path("/application")
 class ApplicationPropertiesEndpoint(private val configuration: ApplicationConfiguration) {
-    private val mapper = ObjectMapper().setVisibility(PropertyAccessor.ALL, JsonAutoDetect.Visibility.ANY)
+    private val mapper = ObjectMapper().registerKotlinModule()
 
     companion object {
         val log = LoggerFactory.getLogger(ApplicationPropertiesEndpoint::class.java)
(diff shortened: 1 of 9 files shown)


api/src/main/kotlin/com/atlassian/migration/datacenter/api/application/ApplicationPropertiesEndpoint.kt
```diff
@@ -0,0 +1,48 @@
...
+        val log = LoggerFactory.getLogger(ApplicationPropertiesEndpoint::class.java)
+    }
+
+    @GET
+    @Produces(MediaType.APPLICATION_JSON)
+    fun getApplicationProperties(): Response {
+        val props = ApplicationProperties(configuration.applicationVersion)
+        log.trace("Application properties: {}", props)
+
+        return Response.ok(mapper.writeValueAsString(props)).build()
```
theghostwhoforks, line 44: do we need to use the object mapper because the data class does not have a default constructor?

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/quickstartToAtlaskit.tsx
```diff
@@ -166,6 +166,13 @@ const createStringInputFromQuickstartParam: FormElementGenerator = (defaultField
         defaultValue: (Default as string) || '',
     };
 
+    if (param.paramKey === 'JiraVersion') {
+        overrideInputProps = {
+            isDisabled: true,
```
theghostwhoforks, line 171: Will this be disabled even when, say the `appversion` returned from the API is empty?

This should not happen, but since we `setAppVersion('')` in the `QuickstartDeployForm`, we may need to be toggle the disable/enable state based on the value

Thoughts?

api/src/main/kotlin/com/atlassian/migration/datacenter/api/application/ApplicationPropertiesEndpoint.kt
```diff
@@ -0,0 +1,48 @@
...
+import com.fasterxml.jackson.annotation.PropertyAccessor
+import com.fasterxml.jackson.databind.ObjectMapper
+import org.slf4j.LoggerFactory
+import javax.ws.rs.GET
+import javax.ws.rs.Path
+import javax.ws.rs.Produces
+import javax.ws.rs.core.MediaType
+import javax.ws.rs.core.Response
+
+@Path("/application")
```
dylanrathbone, line 30: Shouldnt, by convention, RESTful endpoints be pluralised? maybe change to something like `applicationproperties`

theghostwhoforks: Yes, they should. That's a separate task to make all our endpoints more restful in nature

i.e. from `/migration/<>` `/migrations/latest/<>` etc..

nanux: I've changed the endpoint to `/application/properties`, I was also pondering `/application-properties` but this way in future we can expand `/application` capabilities.

api/src/main/kotlin/com/atlassian/migration/datacenter/api/application/ApplicationPropertiesEndpoint.kt
nanux: I've tried several ways but I wasn't able to get `javax.ws.rs` responses serialize correctly `data class` and I don't see any example of it in our codebase. I have simplified the code a bit anyway.

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/quickstartToAtlaskit.tsx
nanux: I was thinking about it as well but as the default is always filled in case we don't have value coming from the endpoint, it shouldn't happen.

PR #334, opened 2020-06-28
CHET-546: Check for and delete pg_dump archive

dylanrathbone:
Prior to creating `pg_dump` archive, check for and delete it if it exists.

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresExtractor.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresExtractor.kt
@@ -104,7 +104,7 @@ class PostgresExtractor(private val applicationConfiguration: ApplicationConfigu
         }
 
     @Throws(DatabaseMigrationFailure::class)
-    override fun startDatabaseDump(target: Path?): Process? {
+    override fun startDatabaseDump(target: Path): Process {
         return startDatabaseDump(target, false)
     }
 
@@ -122,8 +122,8 @@ class PostgresExtractor(private val applicationConfiguration: ApplicationConfigu
      * @throws DatabaseMigrationFailure on failure.
      */
     @Throws(DatabaseMigrationFailure::class)
-    override fun startDatabaseDump(target: Path?, parallel: Boolean?): Process? {
-        val numJobs = if (parallel!!) 4 else 1 // Common-case for now, could be tunable or num-CPUs.
+    override fun startDatabaseDump(target: Path, parallel: Boolean): Process {
+        val numJobs = if (parallel) 4 else 1 // Common-case for now, could be tunable or num-CPUs.
 
         val pgdump = pgdumpPath ?: throw DatabaseMigrationFailure("Failed to find appropriate pg_dump executable.")
         val config = applicationConfiguration.databaseConfiguration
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresExtractor.kt
```diff
@@ -168,5 +173,14 @@ class PostgresExtractor(private val applicationConfiguration: ApplicationConfigu
             throw DatabaseMigrationFailure("pg_dump process exited with non-zero status: $exit")
         }
     }
-
+    
+    @Throws(IOException::class)
+    fun deleteDatabaseDump(to: Path?) {
```
tarka, line 178: This probably doesn't need to be nullable.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresExtractor.kt
```diff
@@ -168,5 +173,14 @@ class PostgresExtractor(private val applicationConfiguration: ApplicationConfigu
...
+        try {
+            FileUtils.deleteDirectory(File(to.toString()))
```
theghostwhoforks, line 180: You can convert from path to file using `path.ToFile()`

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresExtractor.kt
```diff
@@ -142,6 +143,10 @@ class PostgresExtractor(private val applicationConfiguration: ApplicationConfigu
         builder.environment()["PGPASSWORD"] = config.password
 
         return try {
+            if(Files.exists(target!!))  {
```
tarka, line 146: Unrelated to this PR but we can probably remove the nullables from the methods in this class.

theghostwhoforks, line 146: Why is target a `Path?` instead of just `Path`

The caller makes sure that the Path always has a value, correct?

dylanrathbone: Good question, I'm not sure why the parameter `target` in `startDatabaseDump` has been marked as allowing nulls (`?`). The client calling it should always provide it. Let me cleanup.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresExtractor.kt
dylanrathbone: nice!

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresExtractor.kt
tarka: My omission, I'm on it.

dylanrathbone: > My omission, I'm on it.

Pulled in your changes Steve 👍

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresExtractor.kt
dylanrathbone: Updated now that target path is always provided

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresExtractor.kt
dylanrathbone: Updated now that target path is always provided

PR #337, opened 2020-06-29
Chet-549 Unschedule scheduled jobs created by a migration on cancellation/reset of a migration

theghostwhoforks:
This PR creates a 

1. A new cancellable service interface
2. The ability to un-schedule a job on migration delete
3. An event listener mechanism to listen to migration events and handle cancellation of jobs


spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/CancellableMigrationService.kt
```diff
@@ -0,0 +1,16 @@
...
+ */
+interface CancellableMigrationService {
+    /**
+     * Unschedule a scheduled job for a migration
+     *
+     * @param migrationId an identifier for a migration that spawned the scheduled job
+     *
+     * @return true if the job was unscheduled, false when the job to unschedule does not exist
+     */
+    fun unscheduleMigration(migrationId: Int): Boolean
```
nanux, line 15: Do we need to always pass the ID? We are always using current migration ID but I can see its use in future when you can have multiple migrations.

Do we want to have `unscheduleMigrations()` to cancel all migrations - even if we drop the context for example?

core/src/main/java/com/atlassian/migration/datacenter/core/aws/CancellableMigrationServiceHandler.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Override
+    public void afterPropertiesSet() throws Exception {
+        this.eventPublisher.register(this);
+    }
+
+    @EventListener
+    public void onMigrationResetEvent(MigrationResetEvent event){
+        int migrationId = event.getMigrationId();
+        this.cancellableServices.forEach(x -> x.unscheduleMigration(migrationId));
```
nanux, line 52: Are we going to need something similar when we are disabling the plugin?

nanux:
What is the best way how to test this?

core/src/main/java/com/atlassian/migration/datacenter/core/aws/CancellableMigrationServiceHandler.java
theghostwhoforks: yeah - This PR https://github.com/atlassian-labs/dc-migration-assistant/pull/324 tackles that in a similar manner

theghostwhoforks:
> What is the best way how to test this?

I started a migration and cancelled it. I set the migration context to point to an existing stack, started migration and cancelled it. I had a debug point and verified that it was called and the job was unscheduled (check in the DB)

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/CancellableMigrationService.kt
theghostwhoforks: As our scheduled jobs are created with the migration ID identifier, unscheduleAllMigrations would require us to get a list of all migrations.

Ideally, we should have just one at any time, so perhaps we may not really need this functionality?

PR #338, opened 2020-06-29
allow all stage specific error states to be transitioned to from generic error stages

bpar476:
I was going to implement CHET-557 and create the retry transition for fs bulk copy, but it was already implemented. I then noticed that we don't actually use the stage-specific migration stages except for when aborting the migration, which means that retrying would not work in many cases because the process would be in the generic error state and not the specific error state. I've added this transition as a bit of a band-aid. When retrying, we always abort the process first, which will transition to the stage-specific error state. This will allow the retry to work.

Ideally, we would refactor how error handling works across the board and remove this semantic coupling but I think that's too risky given release is next week.

### DIFF
--- a/spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
+++ b/spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
@@ -38,7 +38,8 @@ enum class MigrationStage {
         override val validAncestorStages = setOf(PROVISION_MIGRATION_STACK)
     },
     PROVISIONING_ERROR {
-        override val validAncestorStages = setOf(PROVISION_APPLICATION, PROVISION_APPLICATION_WAIT, PROVISION_MIGRATION_STACK, PROVISION_MIGRATION_STACK_WAIT)
+        override val validAncestorStages
+            get() = setOf(PROVISION_APPLICATION, PROVISION_APPLICATION_WAIT, PROVISION_MIGRATION_STACK, PROVISION_MIGRATION_STACK_WAIT, ERROR)
     },
     FS_MIGRATION_COPY {
         override val validAncestorStages = setOf(PROVISION_MIGRATION_STACK_WAIT)
@@ -47,7 +48,8 @@ enum class MigrationStage {
         override val validAncestorStages = setOf(FS_MIGRATION_COPY)
     },
     FS_MIGRATION_ERROR {
-        override val validAncestorStages = setOf(FS_MIGRATION_COPY, FS_MIGRATION_COPY_WAIT)
+        override val validAncestorStages
+            get() = setOf(FS_MIGRATION_COPY, FS_MIGRATION_COPY_WAIT, ERROR)
     },
     OFFLINE_WARNING {
         override val validAncestorStages = setOf(FS_MIGRATION_COPY_WAIT)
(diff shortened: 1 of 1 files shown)


bpar476:
If I've misunderstood the code please let me know because I didn't realise it was actually like this 🤔

theghostwhoforks:
my understanding of the valid ancestors is that - 

1. Any specific stage that can fail can reach a stage specific error state (i.e. PROVISIONING_APPLICATION_WAIT -> PROVISIONING_ERROR)
2. Any state can transition to the generic ERROR stage

So the flow, originally, would be provision/migrate -> provision_error/migrate_error

If we were to change the services to transition to the specific error states (and not the generic one), would we need to transition to the specific error state from the generic one (i.e. ERROR -> PROVISIONING_ERROR)?

bpar476:
> my understanding of the valid ancestors is that -
> 
>     1. Any specific stage that can fail can reach a stage specific error state (i.e. PROVISIONING_APPLICATION_WAIT -> PROVISIONING_ERROR)
> 
>     2. Any state can transition to the generic ERROR stage
> 
> 
> So the flow, originally, would be provision/migrate -> provision_error/migrate_error
> 
> If we were to change the services to transition to the specific error states (and not the generic one), would we need to transition to the specific error state from the generic one (i.e. ERROR -> PROVISIONING_ERROR)?

Your understanding is correct. Ideally we would transition directly to the stage-specific error state, however our code base is full of calls to `error()` which will just take them to the generic error state. We could refactor this method to do a different transition depending on the current state (would be much easier in kotlin 😞) but I think this is too risky given release is a week away. This is a workaround that will allow the retry to work from an error condition.

theghostwhoforks:
> > my understanding of the valid ancestors is that -
> > ```
> > 1. Any specific stage that can fail can reach a stage specific error state (i.e. PROVISIONING_APPLICATION_WAIT -> PROVISIONING_ERROR)
> > 
> > 2. Any state can transition to the generic ERROR stage
> > ```
> > 
> > 
> > So the flow, originally, would be provision/migrate -> provision_error/migrate_error
> > If we were to change the services to transition to the specific error states (and not the generic one), would we need to transition to the specific error state from the generic one (i.e. ERROR -> PROVISIONING_ERROR)?
> 
> Your understanding is correct. Ideally we would transition directly to the stage-specific error state, however our code base is full of calls to `error()` which will just take them to the generic error state. We could refactor this method to do a different transition depending on the current state (would be much easier in kotlin 😞) but I think this is too risky given release is a week away. This is a workaround that will allow the retry to work from an error condition.

I've changed the file system service to transition to stage specific errors. See: https://github.com/atlassian-labs/dc-migration-assistant/pull/348

It seemed safe enough - can you take a look and let me know if it is safe to merge? If yes and once that is merged in, we can close this out.

bpar476:
> It seemed safe enough - can you take a look and let me know if it is safe to merge? If yes and once that is merged in, we can close this out.

I think there would still be issues with other parts of the migration e.g.
1. Database migration starts
2. Database migration fails for some reason (`migrationService.error()`)
3. User sees failure and fires off retry
4. Retry aborts any existing process (`migrationServicet.transition(FINAL_SYNC_ERROR)`) -> This will throw invalid transition error

theghostwhoforks:
> > It seemed safe enough - can you take a look and let me know if it is safe to merge? If yes and once that is merged in, we can close this out.
> 
> I think there would still be issues with other parts of the migration e.g.
> 
>     1. Database migration starts
> 
>     2. Database migration fails for some reason (`migrationService.error()`)
> 
>     3. User sees failure and fires off retry
> 
>     4. Retry aborts any existing process (`migrationServicet.transition(FINAL_SYNC_ERROR)`) -> This will throw invalid transition error

fair point - I've not updated the PR so that the error transition looks like - 

```
        when {
            PROVISIONING_ERROR.validAncestorStages.contains(preErrorTransitionStage) -> setCurrentStage(migration, PROVISIONING_ERROR)
            FS_MIGRATION_ERROR.validAncestorStages.contains(preErrorTransitionStage) -> setCurrentStage(migration, FS_MIGRATION_ERROR)
            FINAL_SYNC_ERROR.validAncestorStages.contains(preErrorTransitionStage) ->
                setCurrentStage(migration, FINAL_SYNC_ERROR)
            else -> setCurrentStage(migration, ERROR)
        }
```

Given, it's centralized, I've verified the usecases and this change would/should not cause any errors.

I'm not too happy that this is coming in this close to GA. We need to test this out thoroughly once it's merged in

Thoughts?

bpar476:
> Given, it's centralized, I've verified the usecases and this change would/should not cause any errors.
> 
> I'm not too happy that this is coming in this close to GA. We need to test this out thoroughly once it's merged in
> 
> Thoughts?

I think if we want to reduce the chance of failure we add `ERROR` stage as a valid ancestor stage to all other error stages. This will cover us in case some unexpected transition occurs from error to another error state.

theghostwhoforks:
I'll do this as a quick follow to https://github.com/atlassian-labs/dc-migration-assistant/pull/348

bpar476:
#348 handles this

PR #340, opened 2020-06-30
CHET-492: Make final S3 cleanup async

tarka:
This does not include the UI feedback.

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/MigrationBucketCleanupService.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/MigrationBucketCleanupService.kt
@@ -16,85 +16,39 @@
 
 package com.atlassian.migration.datacenter.core.aws.infrastructure.cleanup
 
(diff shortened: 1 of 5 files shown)


core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/infrastructure/cleanup/MigrationBucketCleanupService.kt
```diff
@@ -16,85 +16,39 @@
...
-
-        isCleaning = true
-
-        val client = s3ClientSupplier.get()
-
-        var response = client.listObjectsV2 { it.bucket(bucket) }
-        var contents = response.contents()
-        var continuation = response.continuationToken()
-        var retryCount = 0
+        val jobId = JobId.of(DatabaseMigrationJobRunner.KEY + migrationService.currentMigration.id)
```
theghostwhoforks, line 38: why do we use the `DatabaseMigrationJobRunner` key here?

tarka: Mistake. Good catch.

PR #343, opened 2020-06-30
CHET-543: Pull the functest target information from the enviroment

tarka:


jira-e2e-tests/docker-compose.yml
```diff
@@ -58,8 +58,8 @@ services:
   functests:
     build:
       context: ./func-tests
-    environment:
-      - JIRA_BASE_URL="http://jira:8080/jira"
+    # environment:
```
theghostwhoforks, line 61: Why is this commented?

tarka: Left over from testing. Doesn't actually do anything really as that's the default anyway, but I'll clean up.

PR #344, opened 2020-06-30
CHET-556: Decouple pg_dump availability check logic

dylanrathbone:
extract `pg_dump` availability and version checks so that they can be run independently.

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/DatabaseClientTools.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/DatabaseClientTools.kt
@@ -15,16 +15,12 @@
  */
 package com.atlassian.migration.datacenter.core.db
 
-import com.atlassian.migration.datacenter.spi.exceptions.DatabaseMigrationFailure
-import java.nio.file.Path
+import net.swiftzer.semver.SemVer
 
-interface DatabaseExtractor {
-    @Throws(DatabaseMigrationFailure::class)
-    fun startDatabaseDump(target: Path): Process?
+interface DatabaseClientTools {
+    fun getDatabaseDumpClientVersion(): SemVer?
+    
+    fun getDatabaseServerVersion(): SemVer?
 
-    @Throws(DatabaseMigrationFailure::class)
-    fun startDatabaseDump(target: Path, parallel: Boolean): Process
-
-    @Throws(DatabaseMigrationFailure::class)
-    fun dumpDatabase(to: Path)
+    fun getDatabaseDumpClientPath(): String?
 }
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
@@ -1,0 +1,121 @@
+/*
+ * Copyright 2020 Atlassian
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.atlassian.migration.datacenter.core.db
+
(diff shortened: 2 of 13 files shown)


core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/DatabaseClientTools.kt
```diff
@@ -0,0 +1,26 @@
...
+package com.atlassian.migration.datacenter.core.db
+
+import net.swiftzer.semver.SemVer
+
+interface DatabaseClientTools {
+    fun getDatabaseDumpClientVersion(): SemVer?
+    
+    fun getDatabaseServerVersion(): SemVer?
+
+    fun getDatabaseDumpClientPath(): String?
```
nanux, line 25: Do we want to use `String` to store paths? Wouldn't `Path` be more suitable?

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/DatabaseClientTools.kt
```diff
@@ -0,0 +1,26 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.atlassian.migration.datacenter.core.db
+
+import net.swiftzer.semver.SemVer
+
+interface DatabaseClientTools {
```
nanux, line 20: Short Javadoc for interface could be useful.

core/src/test/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientToolingIT.kt
```diff
@@ -0,0 +1,71 @@
...
+import org.mockito.junit.jupiter.MockitoExtension
+import org.testcontainers.containers.PostgreSQLContainer
+import org.testcontainers.junit.jupiter.Container
+import org.testcontainers.junit.jupiter.Testcontainers
+import org.testcontainers.utility.MountableFile
+import kotlin.test.assertEquals
+
+@Testcontainers
+@ExtendWith(MockitoExtension::class)
+internal class PostgresClientToolingIT {
```
nanux, line 36: Why this needs to be `internal` class?

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
```diff
@@ -0,0 +1,121 @@
...
+import java.nio.file.Files
+import java.nio.file.Paths
+import java.sql.SQLException
+import java.util.*
+import java.util.concurrent.TimeUnit
+
+class PostgresClientTooling(private val applicationConfiguration: ApplicationConfiguration) : DatabaseClientTools {
+    companion object {
+        private val log = LoggerFactory.getLogger(PostgresClientTooling::class.java)
+        private val pddumpPaths = arrayOf("/usr/bin/pg_dump", "/usr/local/bin/pg_dump")
```
nanux, line 31: Any chance we could be a bit smarter here? It doesn't need to be part of this PR but people are getting bitten by this.

dylanrathbone: This will be addressed in https://aws-partner.atlassian.net/browse/CHET-562

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/DatabaseClientTools.kt
dylanrathbone: This will be addressed in https://aws-partner.atlassian.net/browse/CHET-562

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/DatabaseClientTools.kt
dylanrathbone: This will be addressed in https://aws-partner.atlassian.net/browse/CHET-562

core/src/test/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientToolingIT.kt
dylanrathbone: Most if not all of our tests are defined as `internal`. I suspect that this is down to the fact that when auto generating a test file for a class, intellij will automatically create a test class for you with the `internal` access modifier. This kinda makes sense. The way I understand it, `internal` is the equivalent of `default` in Java, meaning that the test class is only accessible within the module/package it resides

PR #345, opened 2020-06-30
CHET-149: Make scheduled services disposable to stop running migratio…

theghostwhoforks:
1. Make services disposable
2. onDispose, services can stop a running migration job

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/DatabaseMigrationService.java
@@ -31,6 +31,7 @@ import com.atlassian.migration.datacenter.spi.infrastructure.InfrastructureDeplo
 import com.atlassian.scheduler.config.JobId;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+import org.springframework.beans.factory.DisposableBean;
 
 import java.nio.file.Path;
 import java.time.Duration;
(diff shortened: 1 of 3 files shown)


theghostwhoforks:
> /nit - it'd be nice if we didn't expose atlassian scheduler jobId in these services and used the migration id & job key. That would provide better encapsulation. This isn't something that is the primary of PR, so perhaps something we can raise a ticket to refactor l8r

I don't think I understand this and I may be missing the point. 

So,In code, we have methods like so - 

```    
private JobId getScheduledJobIdForMigration(int migrationId) {
        return JobId.of(S3UploadJobRunner.KEY + migrationId);
    }
```

and the service calls `getScheduledJobIdForMigration` with the current migration identifier. 
The above function just creates a jobId from the Key and migration identifier. Are you suggesting we pass in just the `key` to the `migrationRunner` instead of the `JobId`?

Is this correct? What am I missing here?

theghostwhoforks:
I'll merge this PR in and let's continue the conversation on the PR/offline later

vicav:
> > /nit - it'd be nice if we didn't expose atlassian scheduler jobId in these services and used the migration id & job key. That would provide better encapsulation. This isn't something that is the primary of PR, so perhaps something we can raise a ticket to refactor l8r
> 
> I don't think I understand this and I may be missing the point.
> 
> So,In code, we have methods like so -
> 
> ```
> private JobId getScheduledJobIdForMigration(int migrationId) {
>         return JobId.of(S3UploadJobRunner.KEY + migrationId);
>     }
> ```
> 
> and the service calls `getScheduledJobIdForMigration` with the current migration identifier.
> The above function just creates a jobId from the Key and migration identifier. Are you suggesting we pass in just the `key` to the `migrationRunner` instead of the `JobId`?
> 
> Is this correct? What am I missing here?

Yes, JobId is an implementation detail of the scheduler we are using. We are exposing this implementation detail to the callers of our services - who just care about scheduling something, not the how we are doing. This way if the implementation detail changes we don't have to touch the classes that are more concerned with the business logic (aka the what we need to do).

PR #346, opened 2020-07-01
CHET-562: dynamically find pg_dump 

dylanrathbone:


core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/SsmPsqlDatabaseRestoreService.java
```diff
@@ -73,13 +73,12 @@ public void restoreDatabase()
...
                 migrationInstanceId);
         
-        remoteInstanceCommandRunnerService.restartJiraService();
-        
         migrationStageCallback.transitionToServiceWaitStage();
 
         try {
             consumer.handleCommandOutput();
             migrationStageCallback.transitionToServiceNextStage();
+            remoteInstanceCommandRunnerService.restartJiraService();
```
dylanrathbone, line 81: Unrelated to this PR, but it makes more sense to restart Jira here.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
```diff
@@ -66,18 +66,16 @@ class PostgresClientTooling(private val applicationConfiguration: ApplicationCon
...
     override fun getDatabaseDumpClientPath(): String? {
-        for (path in pddumpPaths) {
-            val p = Paths.get(path)
-            if (Files.isReadable(p) && Files.isExecutable(p)) {
-                return path
-            }
+        val pgDumpPath = resolvePgDumpPath()
+        return when(pgDumpPath != null && Files.isReadable(pgDumpPath) && Files.isExecutable(pgDumpPath)) {
+            true -> pgDumpPath.toString();
+            else -> null
```
bpar476, line 77: Should we really be returning `null` here and forcing null checks everywhere? I think it would be more idiomatic to throw an exception as this is an error case i.e. there is no pg dump available

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
```diff
@@ -108,14 +106,21 @@ class PostgresClientTooling(private val applicationConfiguration: ApplicationCon
...
-    private val pgdumpPath: String?
-        get() {
-            for (path in pddumpPaths) {
-                val p = Paths.get(path)
-                if (Files.isReadable(p) && Files.isExecutable(p)) {
-                    return path
-                }
-            }
-            return null
+    private fun resolvePgDumpPath(): Path? {
```
bpar476, line 109: Should we have `"/usr/bin/pg_dump", "/usr/local/bin/pg_dump"` as default return values in case the host doesn't have `which` installed (common in some docker images)

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
dylanrathbone: In this case  we dont want to throw an exception, `null` is used by the existing client of this refactored code to drive whether or not `pg_dump` is indicated as available or not on our initial trebuchet pre-req screen.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
dylanrathbone: I hear you, but I'm assuming that in most cases if not all Jira will be run using full POSIX complaint Linux distro of some sort, i.e. Ubuntu or RedHat. Out of curiosity I spun up the latest `alpine` and `busybox` docker images and `which` was available on both. However lets play it safe and fallback like you say.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
bpar476: Ah I see, if this is the case then it's acceptable.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
bpar476: Hmm, I could have sworn `which` was not installed on our jira docker image but looks like you're right! I'm too fussy 😝

dylanrathbone: Well you may be right, I didnt test the Jira image. But I like you're thinking! Defensive coding is a better stance to have here.

PR #348, opened 2020-07-01
Chet 558 stage specific error

theghostwhoforks:
Create a kotlin wrapper implementation for AWS migration service.

Implement a method to transition to any error stage, not just the generic `ERROR` stage

Transitions error stage transition from file system migration service to `FS_MIGRATION_ERROR`

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/AwsMigrationServiceWrapper.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/AwsMigrationServiceWrapper.kt
@@ -29,6 +29,8 @@ open class AwsMigrationServiceWrapper(ao: ActiveObjects?, applicationConfigurati
         when {
             PROVISIONING_ERROR.validAncestorStages.contains(preErrorTransitionStage) -> setCurrentStage(migration, PROVISIONING_ERROR)
             FS_MIGRATION_ERROR.validAncestorStages.contains(preErrorTransitionStage) -> setCurrentStage(migration, FS_MIGRATION_ERROR)
+            FINAL_SYNC_ERROR.validAncestorStages.contains(preErrorTransitionStage) ->
+                setCurrentStage(migration, FINAL_SYNC_ERROR)
             else -> setCurrentStage(migration, ERROR)
         }
 
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/AwsMigrationServiceWrapper.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/AwsMigrationServiceWrapper.kt
@@ -6,44 +6,39 @@ import com.atlassian.migration.datacenter.analytics.events.MigrationFailedEvent
 import com.atlassian.migration.datacenter.core.application.ApplicationConfiguration
 import com.atlassian.migration.datacenter.dto.MigrationContext
 import com.atlassian.migration.datacenter.spi.MigrationService
-import com.atlassian.migration.datacenter.spi.MigrationStage
-import com.atlassian.migration.datacenter.spi.exceptions.InvalidMigrationStageError
+import com.atlassian.migration.datacenter.spi.MigrationStage.*
 import kotlin.math.min
 
 
 open class AwsMigrationServiceWrapper(ao: ActiveObjects?, applicationConfiguration: ApplicationConfiguration?, eventPublisher: EventPublisher?) : AWSMigrationService(ao, applicationConfiguration, eventPublisher), MigrationService {
 
-    override fun stageSpecificError(errorStage: MigrationStage, message: String) {
-        if (!errorStage.isErrorStage()) {
-            throw InvalidMigrationStageError("Stage $errorStage is not a valid error stage")
-        }
+    override fun error(message: String) {
+        this.stageSpecificError(message);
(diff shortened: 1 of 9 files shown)


core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/AwsMigrationServiceWrapper.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.atlassian.migration.datacenter.analytics.events.MigrationFailedEvent
+import com.atlassian.migration.datacenter.core.application.ApplicationConfiguration
+import com.atlassian.migration.datacenter.dto.MigrationContext
+import com.atlassian.migration.datacenter.spi.MigrationService
+import com.atlassian.migration.datacenter.spi.MigrationStage
+import com.atlassian.migration.datacenter.spi.exceptions.InvalidMigrationStageError
+import kotlin.math.min
+
+
+open class AwsMigrationServiceWrapper(ao: ActiveObjects?, applicationConfiguration: ApplicationConfiguration?, eventPublisher: EventPublisher?) : AWSMigrationService(ao, applicationConfiguration, eventPublisher), MigrationService {
```
theghostwhoforks, line 14: A kotlin sub class of migration service.

core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/AwsMigrationServiceWrapper.kt
```diff
@@ -0,0 +1,49 @@
...
+
+    override fun stageSpecificError(errorStage: MigrationStage, message: String) {
```
theghostwhoforks, line 16: the stage specific error will be called from services. This method ensures that the stage being transitioned to is an error stage

theghostwhoforks:
cc: @bpar476 

I've migrated all use cases of error stage transitions in the file system upload flow to use the `stageSpecificError` method.

This seems to be correct. Also, the migration endpoint translates all stage specific errors to the generic error.

See the idea method usage finder here (for `error(string)` and `error(throwable)`) -

Screenshot 2020-07-01 at 16 45 07


Screenshot 2020-07-01 at 16 44 51

core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/AwsMigrationServiceWrapper.kt
```diff
@@ -0,0 +1,49 @@
...
+        val migration = findFirstOrCreateMigration()
+        val context: MigrationContext = this.currentContext
+        val now = System.currentTimeMillis() / 1000L
+
+        val failStage = context.migration.stage
+
+
+        setCurrentStage(migration, errorStage)
+        // We must truncate the error message to 450 characters so that it fits in the varchar(450) column
+        // We must truncate the error message to 450 characters so that it fits in the varchar(450) column
```
bpar476, line 30: Duplicated comment

core/src/main/kotlin/com/atlassian/migration/datacenter/core/aws/AwsMigrationServiceWrapper.kt
bpar476: Instead of having a specific method where a stage must specify it's stage specific error stage, would it be better to just keep the `error()` method with all its signatures, and decide the error stage in the `MigrationService` based on the current stage. Something like:
```
override fun error(message: String) {
       when(this.currentStage) {
            FS_MIGRATION_COPY, FS_MIGRATION_COPY_WAIT -> setCurrentStage(migration, FS_MIGRATION_ERROR)
            PROVISION_APPLICATION, PROVISION_APPLICATION_WAIT..... -> setCurrentStage(migration, PROVISIONING_ERROR)
            .
            .
            .
            // And so on...
        }
}
```
This way each of the migration operations can be left exactly as they are now and removes the possibility of accidentally doing the wrong error transition.

theghostwhoforks: oh yeah - I started with the idea (based on your comment on the other [PR](https://github.com/atlassian-labs/dc-migration-assistant/pull/338), I think) that I'd do this, but went for an explicit method thinking services can invoke it on demand.

Now that I look at the code, I think we can do this only for provision and filesystem services and not do this for the database service.

Let me update the PR real soon

theghostwhoforks: I've made the change so that the code looks like this - 

```
        when {
            PROVISIONING_ERROR.validAncestorStages.contains(preErrorTransitionStage) -> setCurrentStage(migration, PROVISIONING_ERROR)
            FS_MIGRATION_ERROR.validAncestorStages.contains(preErrorTransitionStage) -> setCurrentStage(migration, FS_MIGRATION_ERROR)
            FINAL_SYNC_ERROR.validAncestorStages.contains(preErrorTransitionStage) ->
                setCurrentStage(migration, FINAL_SYNC_ERROR)
            else -> setCurrentStage(migration, ERROR)
        }
```
Turns out, I had to do this for the FINAL_SYNC stages too. However, this should fix any potential bugs that would arise from the error to start stage transitions that you referred to

PR #349, opened 2020-07-01
CHET-567: Integration test fixes

tarka:


core/src/test/kotlin/com/atlassian/migration/datacenter/core/aws/SqsApiIT.kt
```diff
@@ -35,7 +35,7 @@ import kotlin.test.assertTrue
 
 @Tag("integration")
 @ExtendWith(LocalstackDockerExtension::class)
-@LocalstackDockerProperties(useSingleDockerContainer = true, services = ["sqs"], imageTag = "0.10.8")
+@LocalstackDockerProperties(services = ["sqs"], imageTag = "0.10.8")
```
theghostwhoforks, line 38: It'll be interesting to see how this performs on actions given we will spawn a new docker instance per test :)

tarka: Not much. We get huge variances in run-times from Actions anyway, so this doesn't appear to make any noticeable difference.

PR #354, opened 2020-07-02
Chet 558 retry api

theghostwhoforks:
FS Endpoint now exposes a retry method that aborts a running migration, transitions the stage to `FS_MIGRATION` and schedules a new migration.

If abort or transition operation fails, a `BAD_REQUEST` response is returned.
If we are unable to schedule a new migration, a `CONFLICT` response is returned

If all works as expected, then we return an `ACCEPTED` response

### DIFF
--- a/api/src/main/kotlin/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.kt
+++ b/api/src/main/kotlin/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.kt
@@ -18,6 +18,8 @@ package com.atlassian.migration.datacenter.api.fs
 import com.atlassian.migration.datacenter.core.fs.FileSystemMigrationReportManager
 import com.atlassian.migration.datacenter.core.fs.ReportType
 import com.atlassian.migration.datacenter.core.fs.captor.AttachmentSyncManager
+import com.atlassian.migration.datacenter.spi.MigrationService
+import com.atlassian.migration.datacenter.spi.MigrationStage
 import com.atlassian.migration.datacenter.spi.exceptions.InvalidMigrationStageError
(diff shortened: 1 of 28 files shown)


api/src/main/kotlin/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.kt
```diff
@@ -124,9 +128,29 @@ class FileSystemMigrationEndpoint(private val fsMigrationService: FilesystemMigr
     @Produces(MediaType.APPLICATION_JSON)
     @Path("/retry")
     fun retryFileSystemMigration(): Response {
```
theghostwhoforks, line 130: The pattern here is similar to what we follow in the `FinalSyncEndpoint`

api/src/main/kotlin/com/atlassian/migration/datacenter/api/fs/FileSystemMigrationEndpoint.kt
```diff
@@ -124,9 +129,29 @@ class FileSystemMigrationEndpoint(private val fsMigrationService: FilesystemMigr
...
-        log.trace("Retrying file system migration")
+        log.debug("[Retry operation] Retrying file system migration")
+        try {
+            log.debug("[Retry operation] Aborting current migration, if there is a migration in progress")
+            fsMigrationService.abortMigration()
+        } catch (e: InvalidMigrationStageError) {
+            log.error("[Retry operation] Unable to abort a migration", e)
```
bpar476, line 137: If we're unable to abort the migration, do we really want to block the retry? Especially if it is for a bad transition?

I think it would be safe to continue here

PR #355, opened 2020-07-02
CHET-516: Fix Multipart uploader

nanux:
* Smaller chunk size to avoid write timeout
* Fix Java 11 and 8 interoperability (the problem is API change for `ByteBuffer` in Java 9 https://jira.mongodb.org/browse/JAVA-2559)

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
@@ -118,7 +118,7 @@ public class S3FilesystemMigrationService implements FilesystemMigrationService,
             logger.info("Completed file system migration. Transitioning to next stage.");
             migrationService.transition(MigrationStage.OFFLINE_WARNING);
         } catch (FileSystemMigrationFailure e) {
-            logger.error("Encountered critical error during file system migration");
+            logger.error("Encountered critical error during file system migration", e);
             report.setStatus(FAILED);
             migrationService.error(e.getMessage());
         }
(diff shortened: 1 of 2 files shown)


core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3FilesystemMigrationService.java
```diff
@@ -118,7 +118,7 @@ public void startMigration() throws InvalidMigrationStageError {
             logger.info("Completed file system migration. Transitioning to next stage.");
             migrationService.transition(MigrationStage.OFFLINE_WARNING);
         } catch (FileSystemMigrationFailure e) {
-            logger.error("Encountered critical error during file system migration");
+            logger.error("Encountered critical error during file system migration", e);
```
nanux, line 121: This made investigation quite fun :)

core/src/main/java/com/atlassian/migration/datacenter/core/fs/S3MultiPartUploader.java
```diff
@@ -44,17 +46,18 @@
...
+ * https://docs.aws.amazon.com/AmazonS3/latest/dev/qfacts.html
  */
 public class S3MultiPartUploader {
     private final static Logger logger = LoggerFactory.getLogger(S3MultiPartUploader.class);
     private final S3UploadConfig config;
     private final File file;
     private final String key;
 
-    private int sizeToUpload = 100 * 1024 * 1024; // 100 MB
+    private int sizeToUpload = 25 * 1024 * 1024; // 25 MB
```
nanux, line 60: I had implementation with tunning the `NettyNioAsyncHttpClient` with increasing write timeout but it was introducing new dependency and the write timeout is a guess anyway. This is a simpler approach (not hidden in the BeanConfiguration supplier). With current limits (10000 parts) it means that people can upload up to ~240GB files. I believe that is sufficient.

PR #357, opened 2020-07-02
CHET-553: new validate ui

bpar476:
## Summary

* Add checkbox to gate "finishing" the migration
* Drop the migration summary and link to new instance from validation screen
* Update step numbers
- [x] Update string for checkbox
- [x] Decrease width of section message
- [x] Open the new instance in a new tab when finishing migration (user will have to allow pop-ups) 

Final Product:
image


frontend/src/main/resources/i18n/dc-migration-assistant.properties
```diff
@@ -155,10 +155,11 @@ atlassian.migration.datacenter.validation.summary.phrase.instanceUrl=AWS Jira In
 atlassian.migration.datacenter.validation.summary.phrase.migrationDuration=Migration Duration
 atlassian.migration.datacenter.validation.summary.phrase.databaseSize=Database size transferred
 atlassian.migration.datacenter.validation.summary.phrase.fileTransferCount=Total number of files transferred
-atlassian.migration.datacenter.validation.actions.required=You'll still need to:
-atlassian.migration.datacenter.validation.post.action.aws.login=Log in to the new AWS instance at the address above
+atlassian.migration.datacenter.validation.actions.required=To complete your migration, you'll still need to:
+atlassian.migration.datacenter.validation.post.action.aws.login=Make gsure you can access your new Jira instance
```
dylanrathbone, line 159: `gsure` ?

frontend/src/main/resources/i18n/dc-migration-assistant.properties
```diff
@@ -155,10 +155,11 @@ atlassian.migration.datacenter.validation.summary.phrase.instanceUrl=AWS Jira In
...
 atlassian.migration.datacenter.validation.post.action.reconnect.external.services=Reconnect all external services
-atlassian.migration.datacenter.validation.post.action.aws.test=Test your new AWS deployment
+atlassian.migration.datacenter.validation.post.action.aws.verify=Verify that all your data got migrated
+atlassian.migration.datacenter.validation.post.action.index=Manually rebuild Jira's search index
```
dylanrathbone, line 162: Might be worth mentioning that the rebuild needs to be a `Full` one

bpar476: Good call, I'll check this with @ddomingorh

frontend/src/main/dc-migration-assistant-fe/components/stage/Validation.tsx
```diff
@@ -17,15 +17,19 @@
 import React, { FunctionComponent, ReactElement, useEffect, useState } from 'react';
 import { I18n } from '@atlassian/wrm-react-i18n';
 import SectionMessage from '@atlaskit/section-message';
-import TableTree, { Cell, Row } from '@atlaskit/table-tree';
```
theghostwhoforks, line 20: This wasn't very nice. Kill this with fire!

bpar476:
> This looks good.
> 
> Do we need to call out that there will be a cleanup that will happen implicitly?

Don reckoned the users may not even be that conscious of the migration infrastructure so it may just confuse them. As long as we have that in the guide I think it's fine.

PR #358, opened 2020-07-02
CHET-572: Add PermisionInterceptor to REST endpoints

nanux:
Add WebSudo requirements for REST resources and for the servlet.

### DIFF
--- a/api/src/test/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpointTest.kt
+++ b/api/src/test/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpointTest.kt
@@ -21,16 +21,30 @@ import com.atlassian.migration.datacenter.spi.MigrationService
 import com.atlassian.migration.datacenter.spi.MigrationStage
 import com.atlassian.migration.datacenter.spi.exceptions.InvalidMigrationStageError
 import com.atlassian.migration.datacenter.spi.exceptions.MigrationAlreadyExistsException
-import io.mockk.*
+import com.atlassian.plugins.rest.common.sal.websudo.WebSudoRequiredException
+import com.atlassian.plugins.rest.common.sal.websudo.WebSudoResourceContext
+import com.atlassian.plugins.rest.common.sal.websudo.WebSudoResourceFilterFactory
+import com.sun.jersey.api.model.AbstractResource
+import com.sun.jersey.api.model.AbstractResourceMethod
+import com.sun.jersey.api.model.PathValue
+import com.sun.jersey.spi.container.ContainerRequest
+import io.mockk.MockKAnnotations
+import io.mockk.Runs
+import io.mockk.every
 import io.mockk.impl.annotations.InjectMockKs
 import io.mockk.impl.annotations.MockK
 import io.mockk.junit5.MockKExtension
+import io.mockk.just
+import io.mockk.mockk
+import io.mockk.runs
+import io.mockk.verify
 import org.hamcrest.CoreMatchers.equalTo
 import org.hamcrest.MatcherAssert.assertThat
 import org.hamcrest.Matchers
 import org.junit.jupiter.api.Assertions.assertEquals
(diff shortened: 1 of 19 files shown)


api/src/main/kotlin/com/atlassian/migration/datacenter/api/security/PermissionInterceptor.kt
```diff
@@ -0,0 +1,44 @@
...
+
+    override fun intercept(methodInvocation: MethodInvocation) {
+        logger.debug("Intercepting call, does the user has required sysadmin permissions? {}", permissionEnforcer.isSystemAdmin)
+        try {
+            permissionEnforcer.enforceSystemAdmin()
+            methodInvocation.invoke()
+        } catch (e : AuthorisationException) {
+            methodInvocation.httpContext.response.status = Response.Status.FORBIDDEN.statusCode
+            methodInvocation.httpContext.response.entity =
+                """{"error": "you need to be logged in as a system administrator"}"""
```
theghostwhoforks, line 41: This can be a map instead of a string

api/src/main/kotlin/com/atlassian/migration/datacenter/api/security/PermissionInterceptor.kt
```diff
@@ -0,0 +1,44 @@
...
+
+import com.atlassian.plugins.rest.common.interceptor.MethodInvocation
+import com.atlassian.plugins.rest.common.interceptor.ResourceInterceptor
+import com.atlassian.sal.api.permission.AuthorisationException
+import com.atlassian.sal.api.permission.PermissionEnforcer
+import org.slf4j.Logger
+import org.slf4j.LoggerFactory
+import javax.ws.rs.core.Response
+
+class PermissionInterceptor(private val permissionEnforcer: PermissionEnforcer) : ResourceInterceptor {
```
theghostwhoforks, line 27: Out of scope, but we can make this more generic and catch exceptions and return a JSON response.

i.e. right now, when unhandled exceptions are thrown, the server returns a XML response, and JSON parsing fails on the FE.

This can be handled either here or a separate exception handler interceptor

nanux:
I've changed the approach from the interceptor to the WebSudoManager annotation. It makes the code easier and relies on SAL to provide implementation.

api/src/test/kotlin/com/atlassian/migration/datacenter/api/MigrationEndpointTest.kt
```diff
@@ -182,4 +196,25 @@ class MigrationEndpointTest {
         assertThat(response.status, equalTo(Response.Status.BAD_REQUEST.statusCode))
         assertThat((response.entity as Map<String, String>)["message"], equalTo(message))
     }
+
+    @Test
```
nanux, line 200: I didn't find a cleaner way how to test the resource websudo requirement.

api/src/main/kotlin/com/atlassian/migration/datacenter/api/security/PermissionInterceptor.kt
nanux: This behaviour is still the same with `WebSudoRequired` annotation. It will return XML formatted response.

frontend/src/main/dc-migration-assistant-fe/utils/api.ts
```diff
@@ -56,7 +56,12 @@ export const callAppRest = (
     const basePath = `${contextPathWrapper()}/rest/dc-migration/1.0/${path}`;
     const callPath = queryParams ? `${basePath}?${queryParams}` : basePath;
 
-    return fetch(callPath, options);
+    return fetch(callPath, options).then(response => {
```
nanux, line 59: Confirmed this will refresh the page if the user becomes unauthenticated or drop websudo.

theghostwhoforks: Awesome!

PR #360, opened 2020-07-03
CHET-304: Add provisioning duration to status page.

tarka:


api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.kt
```diff
@@ -20,22 +20,14 @@ import com.atlassian.migration.datacenter.core.aws.region.RegionService
 import com.atlassian.migration.datacenter.spi.MigrationService
 import com.atlassian.migration.datacenter.spi.MigrationStage
 import com.atlassian.migration.datacenter.spi.exceptions.InvalidMigrationStageError
-import com.atlassian.migration.datacenter.spi.infrastructure.ApplicationDeploymentService
-import com.atlassian.migration.datacenter.spi.infrastructure.InfrastructureDeploymentError
-import com.atlassian.migration.datacenter.spi.infrastructure.InfrastructureDeploymentState
-import com.atlassian.migration.datacenter.spi.infrastructure.MigrationInfrastructureDeploymentService
-import com.atlassian.migration.datacenter.spi.infrastructure.ProvisioningConfig
+import com.atlassian.migration.datacenter.spi.infrastructure.*
```
nanux, line 23: We really need to standardize our settings, I am not fussed either way but I have the setting from Jira and other projects which sets the star imports to 9999. I believe @theghostwhoforks mentioned the same. If we want to use the IDEA default settings in the project I am ok with it.

The only reason why I want this to be unified across our IDEAs is to not have git +- changes back and forth depending who last autoformatted the file.

api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.kt
```diff
@@ -56,6 +48,7 @@ class CloudFormationEndpoint(
         private const val stackUrlResponseKey = "stackUrl"
         private const val statusResponseKey = "status"
         private const val phaseResponseKey = "phase"
+        private const val startKey = "startEpoch"
```
nanux, line 51: Can we add the units as well? E.g. `startEpochSeconds`? When reading the code I assumed it is in milliseconds until I found the code when we divide by 1000.

tarka: I'll merge this and raise a separate PR as I'll need to re-test everything.

PR #362, opened 2020-07-03
CHET-560: Allow users to continue with migration

dylanrathbone:
Use conditional rendering to determine whether the `Ignore and continue` button is rendered for a failed migration stage. For the puproses of this PR this button will only be rendered if there is an assets file upload failure i.e `Step 4 of 7: Copy Content`

image

- When the `Ignore and continue` button is selected the user is taken to `Step 5 of 8`...

image

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
@@ -47,6 +47,7 @@ const getDeploymentProgress: ProgressCallback = async () => {
                 onRetry: provisioning.retry,
                 retryText: I18n.getText('atlassian.migration.datacenter.provision.aws.retry.text'),
                 onRetryRoute: asiConfigurationPath,
+                canContinueWhenFailure: false,
             });
             switch (result.status) {
                 case ProvisioningStatus.Complete:
--- a/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
@@ -47,7 +47,7 @@ const getDeploymentProgress: ProgressCallback = async () => {
                 onRetry: provisioning.retry,
                 retryText: I18n.getText('atlassian.migration.datacenter.provision.aws.retry.text'),
                 onRetryRoute: asiConfigurationPath,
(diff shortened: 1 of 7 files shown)


frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferProgress.tsx
```diff
@@ -152,6 +153,12 @@ export const MigrationProgress: FunctionComponent<MigrationProgressProps> = ({
                         >
                             {retryText || 'retry'}
                         </Button>
+                        <Button
```
bpar476, line 156: To have this work for only the FS retry, we need some conditional rendering. 

I would recommend adding an additional property to the retry props, an optional boolean `canContinue`, which if true, will allow the button to render. We can pair on this after planning/lunch if you are free.

bpar476:
I think we also need to hear @ddomingorh and @ashking2's perspective on the button text and appearance. Should it be a different colour? The user needs to realise that by clicking this button, the files will be left behind.

bpar476:
I think this will also appear on the provisioning screen if there is an error there.

frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferProgress.tsx
dylanrathbone: I've actually been trying to use conditional rendering here, but failing badly :) Your help would be great!

dylanrathbone: Sounds good

dylanrathbone: Done.

dylanrathbone:
> This looks good to me! Let's just make sure it doesn't render on the other pages as expected

Yup doing some more tests and waiting for Ashlee's input. Feels like she is going to request a change.

frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
```diff
@@ -47,6 +47,7 @@ const getDeploymentProgress: ProgressCallback = async () => {
                 onRetry: provisioning.retry,
                 retryText: I18n.getText('atlassian.migration.datacenter.provision.aws.retry.text'),
                 onRetryRoute: asiConfigurationPath,
+                canContinueWhenFailure: false,
```
theghostwhoforks, line 50: /nit `canContinueOnFailure`

PR #367, opened 2020-07-07
CHET-591: Add a prefix to ssm parameter lookup names

theghostwhoforks:
Adds the ability to run multiple migration stacks in a region by including a `{StackName}-helper` prefix to parameter store keys

See: https://cloud.spring.io/spring-cloud-static/spring-cloud-aws/2.1.0.RELEASE/single/spring-cloud-aws.html#_integrating_your_spring_cloud_application_with_the_aws_parameter_store

### DIFF
--- a/filesystem-processor/src/main/resources/bootstrap.properties
+++ b/filesystem-processor/src/main/resources/bootstrap.properties
@@ -15,6 +15,7 @@
 #
 
 aws.paramstore.prefix=/atlassian/config
-aws.paramstore.name=migration-helper
+# aws.paramstore.name uses the defaults configured to spring.application.name, so it is not required to specify it here. We will override it to a stack specific name when the processor is run
+#aws.paramstore.name=migration-helper
 aws.paramstore.enabled=true
 aws.paramstore.profileSeparator=_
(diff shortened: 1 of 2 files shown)


bpar476:
I feel like I tried something like this and it didn't work. How does the filesystem processor get the stack name?

bpar476:
> I feel like I tried something like this and it didn't work. How does the filesystem processor get the stack name?

Oh, I see, the parameter store prefix is injected by a system property. Has this been tested?

theghostwhoforks:
> > I feel like I tried something like this and it didn't work. How does the filesystem processor get the stack name?
> 
> Oh, I see, the parameter store prefix is injected by a system property. Has this been tested?

Adding screenshots now. Will upload a video recording soon

1. Multiple stacks in a region

Screenshot 2020-07-08 at 09 18 26


2. File System Processor start up logs
Screenshot 2020-07-08 at 09 08 50


3. Parameter store values
Screenshot 2020-07-08 at 09 18 03

4. File system Processor logs
Screenshot 2020-07-08 at 09 21 28

PR #372, opened 2020-07-08
CHET-585: Update provision error

dylanrathbone:
Update provisioning error, things in green have been updated:

image

- `stacks` link will take you the erroneous stack in CFN
- `terminate all these stacks.` link will take you here: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-console-delete-stack.html


frontend/src/main/dc-migration-assistant-fe/components/aws/quickstart/ProvisioningStatusPage.tsx
```diff
@@ -29,11 +29,25 @@ const buildErrorFromMessageAndUrl = (message: string, stackurl: string): ReactNo
             <p>{I18n.getText('atlassian.migration.datacenter.provision.aws.error.header')}</p>
             <p>{message}</p>
             <p>
-                {I18n.getText('atlassian.migration.datacenter.provision.aws.error.callToConsole')}
+                {I18n.getText('atlassian.migration.datacenter.provision.aws.error.check.allstacks')}
+                &nbsp;
+                <a href={stackurl} target="_blank" rel="noreferrer noopener">
```
bpar476, line 34: From the screenshot, it looks like this link isn't actually rendered as a hyperlink, is it a correct screenshot?

dylanrathbone: Its because `stackurl` isnt set in this case (dummy setup). It will be set in a live environment.

PR #374, opened 2020-07-09
CHET-594: always redirect to current stage

bpar476:
# Summary

We use a custom hook to check the current migration stage and redirect to it if we are not currently there.

I also fixed some of the routings so that refreshing on the provisioning will take you to configure your ASI again, and loading before the final sync takes you to the offline warning.

Demo changing URL at FS stage

### DIFF
--- a/frontend/src/main/dc-migration-assistant-fe/components/MigrationStageRenderer.tsx
+++ b/frontend/src/main/dc-migration-assistant-fe/components/MigrationStageRenderer.tsx
@@ -14,32 +14,16 @@
  * limitations under the License.
  */
 
-import React, { FunctionComponent, useState, useEffect } from 'react';
-import { Redirect } from 'react-router-dom';
+import React, { FunctionComponent } from 'react';
 import Spinner from '@atlaskit/spinner';
-import { getPathForStage } from '../utils/migration-stage-to-path';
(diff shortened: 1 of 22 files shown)


frontend/src/main/dc-migration-assistant-fe/components/MigrationStageRenderer.tsx
```diff
@@ -14,32 +14,16 @@
...
-            .finally(() => setLoading(false));
-    }, []);
+    const loading = useCurrentStageRedirect();
 
     if (loading) {
         return <Spinner />;
     }
 
-    return <Redirect to={getPathForStage(currentStage)} />;
+    return <>I have stopped loading</>;
```
theghostwhoforks, line 28: Is the idea that this will not be rendered because the first time `MigrationStageRenderer` is called, it will only be when the route path does not match and falls through to the default, at which point the URL will be written and an exact match happens?

tarka:
I'll do some manual testing once I'm done with the websudo stuff.

frontend/src/main/dc-migration-assistant-fe/components/MigrationStageRenderer.tsx
bpar476: Correct, this will never be rendered. I'll update it to return null.

PR #375, opened 2020-07-09
CHE-600: Prevent making calls to CFn API with null or empty stackname…

theghostwhoforks:
1. Check if stack name is null before calling `cfn.getStatus`
2. Before provisioning an application stack if current state is provision_error, force the transition to provision_error

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
@@ -128,11 +128,11 @@ public class QuickstartDeploymentService extends CloudformationDeploymentService
     }
 
     private void deployQuickstart(@NotNull String deploymentId, String templateUrl, @NotNull Map<String, String> params, MigrationContext context) throws InvalidMigrationStageError, InfrastructureDeploymentError {
-        // This is a safeguard against CHET-600 (https://aws-partner.atlassian.net/browse/CHET-600) where the statemachine is not transitioned to `PROVISION_APPLICATION` after a retry operation is invoked.
-        // The real cause of why this happens not fully known at the time of this commit. This temporary will transition from `PROVISIONING_ERROR` to  `PROVISIONING_APPLICATION` before deploying the stack
-        // Additionally, we may need to clear persisted stack details here.
+        // This is a safeguard against CHET-600 (https://aws-partner.atlassian.net/browse/CHET-600) which is caused as the statemachine transition to `PROVISIONING_ERROR` is dependent on the scheduled task to complete. If a provisioning operation is retried before the `handleFailedDeployment` task is run, the migration stage transition check would prevent the provisioning from completing successfully.
+        // See https://github.com/atlassian-labs/dc-migration-assistant/pull/375#issuecomment-655894553 for a detailed explanation of why error case is required.
+        // Also, see https://aws-partner.atlassian.net/browse/CHET-608 that tracks this bug.
         if(migrationService.getCurrentStage().equals(MigrationStage.PROVISIONING_ERROR)){
-            migrationService.transition(MigrationStage.PROVISION_APPLICATION_WAIT);
+            migrationService.transition(MigrationStage.PROVISION_APPLICATION);
         }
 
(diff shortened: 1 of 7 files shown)


theghostwhoforks:
> Odd, I see the transition in the endpoint 🤔 Will need some more thorough debugging after GA

So the main issue is

1. The API calls /aws/stack/status
2. This retrieves data from AWS, not our database
3. Say, the CFN stack has failed at this time
4. However, our scheduler runs every 60 seconds. On a timeline, if the scheduler has been sleeping for 1 second, and the CFN provisioning stack fails, the state in the database will not be updated to failed for 59 seconds longer
During this time, if a user retries the migration, the retry will fail as the state is not in error
5. And if a user then tries to provision the stack within those 60 seconds, then the stack provisioning will fail
6. Moreover, even when a user tries to provision after 60 seconds, the provisioning fails as the state in the DB would be in PROVISIONING_ERROR , but we try to transition to PROVISIONING_STACK_WAIT (This PR fixes this)
7. The second bug is that if a user tries to redeploy with the same name, this request will fail as the secret is created. The stage in the database is PROVISION_STACK_WAIT
8. However, they will not be able to use that migration as if a user tries to provision a stack with a new name, the state cannot be transitioned onto itself i..e PROVISION_STACK_WAIT  -> PROVISION_STACK_WAIT  is incorrect

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/QuickstartDeploymentService.java
```diff
@@ -128,6 +128,13 @@ private void addDeploymentModeToContext(ProvisioningConfig.DeploymentMode mode,
     }
 
     private void deployQuickstart(@NotNull String deploymentId, String templateUrl, @NotNull Map<String, String> params, MigrationContext context) throws InvalidMigrationStageError, InfrastructureDeploymentError {
+        // This is a safeguard against CHET-600 (https://aws-partner.atlassian.net/browse/CHET-600) which is caused as the statemachine transition to `PROVISIONING_ERROR` is dependent on the scheduled task to complete. If a provisioning operation is retried before the `handleFailedDeployment` task is run, the migration stage transition check would prevent the provisioning from completing successfully.
+        // See https://github.com/atlassian-labs/dc-migration-assistant/pull/375#issuecomment-655894553 for a detailed explanation of why error case is required.
+        // Also, see https://aws-partner.atlassian.net/browse/CHET-608 that tracks this bug.
+        if(migrationService.getCurrentStage().equals(MigrationStage.PROVISIONING_ERROR)){
+            migrationService.transition(MigrationStage.PROVISION_APPLICATION);
```
theghostwhoforks, line 135: We transition to `provision_application` here so that the subsequent transition to `provision_application_wait` can happen successfully

bpar476:
> > Odd, I see the transition in the endpoint 🤔 Will need some more thorough debugging after GA
> 
> So the main issue is
> 
>     1. The API calls /aws/stack/status
> 
>     2. This retrieves data from AWS, not our database
> 
>     3. Say, the CFN stack has failed at this time
> 
>     4. However, our scheduler runs every 60 seconds. On a timeline, if the scheduler has been sleeping for 1 second, and the CFN provisioning stack fails, the state in the database will not be updated to failed for 59 seconds longer
>        During this time, if a user retries the migration, the retry will fail as the state is not in error
> 
>     5. And if a user then tries to provision the stack within those 60 seconds, then the stack provisioning will fail
> 
>     6. Moreover, even when a user tries to provision after 60 seconds, the provisioning fails as the state in the DB would be in PROVISIONING_ERROR , but we try to transition to PROVISIONING_STACK_WAIT (This PR fixes this)
> 
>     7. The second bug is that if a user tries to redeploy with the same name, this request will fail as the secret is created. The stage in the database is PROVISION_STACK_WAIT
> 
>     8. However, they will not be able to use that migration as if a user tries to provision a stack with a new name, the state cannot be transitioned onto itself i..e PROVISION_STACK_WAIT  -> PROVISION_STACK_WAIT  is incorrect

Looks like there are two actions from this:
1. Refactor the cfn status to commit to the DB from the scheduled executor
2. Fix errors during secret creation not causing an error transition

PR #376, opened 2020-07-09
CHET-606: Skip websudo for status endpoints

tarka:
Skip websudo for status endpoints, which avoids the sudo redirect during migration. The status endpoints should never return anything sensitive. With this in place the admin will only be prompted a password after the transition to the next stage.


theghostwhoforks:
And this should fix the network error flag on the UI, correct?

PR #377, opened 2020-07-09
CHET-587: Add start epoch to RDS secret 

dylanrathbone:
**PROPOSAL**
Add start epoch to RDS secret to avoid clashes on new migrations

**NOTE:** This is a non GA bug, but has the potential to be a real pain if encountered out in the wild.

🚧 Testing now ....

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/TargetDbCredentialsStorageService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/TargetDbCredentialsStorageService.java
@@ -89,6 +89,6 @@ public class TargetDbCredentialsStorageService {
 
     public String getSecretName() {
         MigrationContext context = migrationService.getCurrentContext();
-        return String.format("atl-%s-migration-app-rds-password", context.getApplicationDeploymentId());
+        return String.format("atl-%s-migration-app-rds-password-%s", context.getApplicationDeploymentId(), context.getStartEpoch());
     }
 }


dylanrathbone:
> Unfortunately there is a semantic coupling between the template and the secret name. Anything we put in the secret name needs to be deterministic or shareable with the migrationstack. We could do this by making it a parameter to the migration stack e.g. parameter: SecretName

:( Felt a bit easy...

bpar476:
> Unfortunately there is a semantic coupling between the template and the secret name. Anything we put in the secret name needs to be deterministic or shareable with the migrationstack. We could do this by making it a parameter to the migration stack e.g. parameter: SecretName

In the template: 
`SECRET_PASSWORD=`aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"`
...
`SecretIdentifier: !Sub "atl-${AWS::StackName}-app-rds-password"`

dylanrathbone:
> > Unfortunately there is a semantic coupling between the template and the secret name. Anything we put in the secret name needs to be deterministic or shareable with the migrationstack. We could do this by making it a parameter to the migration stack e.g. parameter: SecretName
> 
> In the template:
> 
> ```
> SECRET_PASSWORD=`aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"`
> ```
> 
> ...
> `SecretIdentifier: !Sub "atl-${AWS::StackName}-app-rds-password"`

Of course. Unfortunately we have no clear way of sharing the start epoch that is set against the secret name here.

dylanrathbone:
> > > Unfortunately there is a semantic coupling between the template and the secret name. Anything we put in the secret name needs to be deterministic or shareable with the migrationstack. We could do this by making it a parameter to the migration stack e.g. parameter: SecretName
> > 
> > 
> > In the template:
> > ```
> > SECRET_PASSWORD=`aws secretsmanager  get-secret-value --secret-id ${SecretIdentifier} --region ${AWS::Region} --output text --query "SecretString"`
> > ```
> > 
> > 
> > ...
> > `SecretIdentifier: !Sub "atl-${AWS::StackName}-app-rds-password"`
> 
> Of course. Unfortunately we have no clear way of sharing the start epoch that is set against the secret name here.

Just noticed your comment on adding a parameter to the migration stack... Sounds like an option

theghostwhoforks:
+1 to what @bpar476 said. The cleanest way to do this is to generate a random text in the key name and pass it in the parameter. 

This would be a very good fix to have. The secret name collision does not trigger valid state machine transitions and breaks a migration

core/src/main/java/com/atlassian/migration/datacenter/core/aws/db/restore/TargetDbCredentialsStorageService.java
```diff
@@ -89,6 +89,6 @@ public void storeCredentials(String password) {
 
     public String getSecretName() {
         MigrationContext context = migrationService.getCurrentContext();
-        return String.format("atl-%s-migration-app-rds-password", context.getApplicationDeploymentId());
+        return String.format("atl-%s-migration-app-rds-password-%s", context.getApplicationDeploymentId(), context.getStartEpoch());
```
theghostwhoforks, line 92: would the resulting length be in compliance with the key name restrictions put in by AWS Param store?

bpar476:
Can we close this @dylanrathbone ?

bpar476:
There's been no activity here for a while. I'm going to close this PR and we can look back at it if we decide to address the issue in planning.

PR #380, opened 2020-07-09
Disable all caching on the template servlet.

tarka:


core/src/main/java/com/atlassian/migration/datacenter/TemplateServlet.java
```diff
@@ -48,6 +48,8 @@ public TemplateServlet(final SoyTemplateRenderer templateRenderer,
     @Override
     protected void doGet(HttpServletRequest request, HttpServletResponse response) throws IOException {
         try {
+            response.setHeader("Pragma", "no-cache");
+            response.setHeader("Cache-Control", "private, no-store, no-cache, must-revalidate");
```
bpar476, line 52: Do we need all of these directives? From the MDN doc it looks like all we should need is `no-store`.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/cache-control
> no-store
    The response may not be stored in any cache. Although other directives may be set, this alone is the only directive you need in preventing cached responses on modern browsers. max-age=0 is already implied. Setting must-revalidate does not make sense because in order to go through revalidation you need the response to be stored in a cache, which no-store prevents.

tarka: Browser compatability issues. Which is why e.g. Safari still claims to be "Mozilla/5.0".

bpar476: Is it a "just in case" type thing or do we know that certain browsers are not up to date. The MDN page says modern browsers respect the `cache-control` header.

tarka: Just in case. Modern is highly subjective, especially in the corporate space, and I'd have to go through all of our products supported browser lists to ensure that it's supported. The prameters above are the "set it and forget it" defaults that people use in practice. There's no harm in adding the additional parameters.

bpar476: Sounds good :)

PR #381, opened 2020-07-09
no-issue: Fix self-redirect when moving from offline warning to final sync

bpar476:
## Context

When clicking "Next" on the offline warning page it would cause a full-page reload and thus redirect back to itself. This is because when the page is loaded fresh, it tries to send you to the "current" page for where you are in the migration. Because there is no state transition when doing this "next", we are in the same state and thus load the same page.

## Solution

We get around this by using a client-side redirect instead of a full-page reload. The ASI configuration -> deploy cluster has the same anomaly in that there are two pages for the same state but it was already using a client-side redirect instead of a full-page reload so never fell victim to this behaviour.


dylanrathbone:
Currently confirming this before merge

PR #384, opened 2020-07-13
CHET-610: Add --no-owner --no-acl flags to migration template (already present in S3 bucket)

nanux:
Before the blitz, I've updated the migration template manually to contains these flags. I just need to sync the change on `master`.

For comparison:
https://s3.console.aws.amazon.com/s3/object/trebuchet-public-resources/migration-helper.yml?region=us-east-1&tab=overview


nanux:
> oh wow - I thought we had this already?
> 
> How was it working before this change? Does the user we create in ansible have access to GRANT/REVOKE?

The template was correct in S3 bucket as I've uploaded it manually for the blitz but we also need to sync it in our source code.

PR #385, opened 2020-07-13
INNO: Generate analytics whitelist at build time

tarka:
There is already drift from our actual analytics and the whitelist; it's better to generate the whitelist from the known analytics sources.


jira-plugin/src/main/kotlin/com/atlassian/migration/datacenter/build/GenWhitelist.kt
```diff
@@ -0,0 +1,63 @@
...
+
+/**
+ * Small util to generate an analytics whitelist from the defined events.
+ * Intended to be called during the build phase; see `pom.xml` and `atlassian-plugin.xml`.
+ */
+fun main(args: Array<String>) {
+    val pkg = args[0]
+    val target = args[1]
+
+    val eventAnno = "com.atlassian.analytics.api.annotations.EventName"
```
vicav, line 31: You should be able to just use spring for this: https://stackoverflow.com/questions/259140/scanning-java-annotations-at-runtime

jira-plugin/pom.xml
```diff
@@ -83,6 +83,19 @@
             <artifactId>atlassian-plugins-osgi-javaconfig</artifactId>
         </dependency>
 
+        <dependency>
```
bpar476, line 86: Do these dependencies get compiled into the final JAR?

jira-plugin/src/main/kotlin/com/atlassian/migration/datacenter/build/GenWhitelist.kt
tarka: I've used Spring Scanner in the past, but the when using it with a standalone executable you rapidly get into classpath hell, as a `provided` dependency is visible to it. ClassGraph is smaller and simpler to work with so made sense in this case.

jira-plugin/pom.xml
tarka: Not anymore, I just fixed that.

jira-plugin/src/main/kotlin/com/atlassian/migration/datacenter/build/GenWhitelist.kt
tarka: Actually, that's not correct, it is possible to use a provided dependency. However ClassGraph is simpler to work with, so I'm happy to leave it in as it doesn't cost anything.

vicav: Yep - now that the deps are provided and not compile scope agreed.

tarka: So making the dependencies `provided` confuses OSGI and the whole thing falls over. I had to refactor them all out into a separate `buld-utils` module, which probably makes more sense anyway.

tarka: If this is of use to ATST we can split it out into its own repo.

build-utils/pom.xml
```diff
@@ -0,0 +1,49 @@
...
+            <artifactId>classgraph</artifactId>
+            <version>${classgraph.version}</version>
+        </dependency>
+        <dependency>
+            <groupId>com.google.code.gson</groupId>
+            <artifactId>gson</artifactId>
+            <version>${gson.version}</version>
+        </dependency>
+    </dependencies>
+
```
vicav, line 48: Should move the gson.version, classgraph.version properties to this pom too

jira-plugin/pom.xml
```diff
@@ -88,10 +88,45 @@
             <artifactId>junit-jupiter-api</artifactId>
             <scope>test</scope>
         </dependency>
+
+        <dependency>
+            <groupId>com.atlassian.migration.datacenter</groupId>
+            <artifactId>build-utils</artifactId>
+            <version>${project.version}</version>
+            <scope>provided</scope>
```
theghostwhoforks, line 96: Is this set to `provided` as we don't want to compile this into the generated jar/obr?

tarka: Correct; the module and dependencies are downloaded and available at bulid time, but not packaged.

PR #387, opened 2020-07-16
INNO: migration transfer page refactor

bpar476:
I've decomposed and added test coverage for most of the `MigrationTransfer*` components which have been a bit fickle.

### Summary

* Remove some dead code
* Consolidate some redundant data
* Separate out retry components from migration progress
* Add storybook for most of the sub-components

- [x] Add storybook for the retry menu


bpar476:
Will do a pre-review and end-to-end test locally tomorrow.

frontend/src/main/dc-migration-assistant-fe/components/final-sync/FinalSync.tsx
```diff
@@ -143,7 +125,24 @@ const props: MigrationTransferProps = {
     startButtonText: I18n.getText('atlassian.migration.datacenter.finalSync.startButton'),
     startMigrationPhase: startFinalSync,
     inProgressStages: finalSyncInProgressStages,
-    getProgress: getProgressFromStatus,
+    processes: [
```
bpar476, line 128: Unfortunately this change means that getting the final sync status makes redundant API calls. However, a future refactor should be to pass the progress directly to the `MigrationTransferPage` and manage the fetching outside.

bpar476:
> This looks nice!
> 
> The only question that I have is how are we ensuring the stories do not get bundled (?) as part of the minified artifact generated when the plugin is built?

I looked into this as I had the same concern. I used webpack --display-modules to see which modules are included in the final build and the stories and tests are ignored. They're not included because webpack looks at the index and only builds in what is actually required by each module. Nothing imports the tests or stories so they are not in the final build :)

theghostwhoforks:
> > This looks nice!
> > The only question that I have is how are we ensuring the stories do not get bundled (?) as part of the minified artifact generated when the plugin is built?
> 
> I looked into this as I had the same concern. I used webpack --display-modules to see which modules are included in the final build and the stories and tests are ignored. They're not included because webpack looks at the index and only builds in what is actually required by each module. Nothing imports the tests or stories so they are not in the final build :)

Smart 👍

PR #390, opened 2020-07-21
no-issue: remove that traitorous deserter from the project owner list

theghostwhoforks:
Next step - Rewrite history and erase all traces from git permanently!


nanux:
We could have moved you to:

```
# Other contributors
# - @theghostwhoforks 
```

PR #393, opened 2020-07-22
Chet 539 fix i18n break jira 7.13

bpar476:
# Summary

The issue with Jira 7.13 came from `wrm/format` not being defined for the `wrm-react-i18n` plugin which we use for our i18n. This caused i18n translations with substitutions to cause a NPE. The root cause of this is being looked into, as a workaround I've changed all our i18n strings to have no substitution. The work was documented [here](https://aws-partner.atlassian.net/wiki/spaces/DOA/pages/369033233/Index+of+internationalisation+strings+with+substitution)

I ran an end-to-end migration on Jira 7.13.14 to confirm everything worked okay

## Migrated Issue
image
## Proof of version
Screen Shot 2020-07-22 at 11 11 52 am


bpar476:
Going to close this in favour of @macku 's PR

PR #392, opened 2020-07-22
None: Fix loading correct i18n module and fix webpack build mode

macku:
While helping Ben I've found two problems:
 - the `@atlassian/wrm-react-i18n` module was imported from the mocs directory
 - The production mode passed to the `getLoaders` function is a boolean value. Additionally, the webpack build mode wasn't specified explicitly.

PR #395, opened 2020-07-22
CHET-539: fix i18n module causing issues with Jira 7.13

bpar476:
# Summary

We weren't loading the i18n module correctly which resulted in the `AJS` object being overridden in Jira 7.13

I also added the typings that I had in the previous PR, except this time I used the proper signature for `I18n.getText`. This will give us nice type-safety as well as autocomplete.

Credit to @macku for the original fix in #392


frontend/src/main/dc-migration-assistant-fe/typings/wrm-react-i18n.d.ts
```diff
@@ -0,0 +1,21 @@
+/*
```
macku, line 1: the latest version of the package 0.7.0 includes the typing:
https://unpkg.com/browse/@atlassian/wrm-react-i18n@0.7.0/wrm-react-i18n.d.ts

bpar476: I tried upgrading but it wasn't compiling :( I got these errors. Looks like it can't resolve the React types
image
For now, I'm going to merge this and we will review upgrading later on.

PR #398, opened 2020-07-23
CHET-615: refresh migration process status when retrying

bpar476:
# Summary

* Add update of current progress to retry callback in `MigrationTransferPage`
* Hoist retry redirect out of `RetryMenu` and into `ProvisionStatus`, the only page that actually had a redirect after retrying
CHET-615-demo


frontend/src/main/dc-migration-assistant-fe/components/shared/MigrationTransferPage.stories.tsx
```diff
@@ -106,7 +106,7 @@ export const unhappyPath = (): ReactNode => {
                             retryText: 'Retry',
                             onRetry: (): Promise<void> => {
                                 progressFetchCount = 0;
-                                return Promise.resolve();
+                                return new Promise(resolve => setTimeout(resolve, 1000));
```
bpar476, line 109: This lets you see the loading while the retry operation is pending

nanux:
LGTM

PR #401, opened 2020-07-28
CHET-622: Add ability to validate the migrated issues and attachments

tarka:


jira-e2e-tests/cypress/support/common.ts
```diff
@@ -1,27 +1,64 @@
...
+
+export type AppContext = {
     username: string;
     password: string;
     base: string;
     context: string;
-    baseURL: string;
-    welcomeURL: string;
-    loginURL: string;
+    baseurl: string;
```
nanux, line 24: /nit we could then probably be consistent and either use the same case for all properties with `url`, I quite like some division between the words - e.g. `baseUrl`/`baseURL`

jira-e2e-tests/cypress/support/pages/ValidationPage.ts
```diff
@@ -30,4 +35,6 @@ export const showsValidationPage = () => {
     cy.location().should((loc: Location) => {
         expect(loc.pathname).to.contain('/home');
     });
+
+    return serviceUrl
```
nanux, line 39: Is this always filled with the value? I am still js noob but the execution model doesn't give guarantees that the `cy.get('#dc-migration-assistant-root p > a').then()` promise resolves before the function `showsValidationPage` exits, does it?

jira-e2e-tests/cypress/support/jira.ts
```diff
@@ -45,7 +74,57 @@ export const getContext = () => {
...
+export const reindex = (issues: string[], ctx: AppContext, targetURL: string) => {
+    cy.request({
+        url: targetURL+`/rest/api/2/reindex/issue?issueID=${issues.join(",")}`,
+        method: "POST",
+        headers: {"Origin": targetURL},
+        auth: ctx.restAuth,
+    })
+}
+
+export const validate_issue = (issueKey: string, ctx: AppContext, targetURL: string, att_name: string?, att_hash: string?, att_thumb_hash: string?) => {
```
nanux, line 95: It is an interesting mix in variable names ;)

jira-e2e-tests/cypress/support/jira.ts
```diff
@@ -45,7 +74,57 @@ export const getContext = () => {
         case 'fsdev': {
             return fsDevServerContext;
         }
+        case 'local': {
+            return dockerLocalContext;
```
nanux, line 78: I assume you are also running it locally with `yarn cypress` from the terminal and therefore the app is not accessible via `jira:8080`.

jira-e2e-tests/cypress/support/common.ts
```diff
@@ -1,27 +1,64 @@
...
+/**
+ * Simple MD5 checksum of data; uses the system md5sum util and simple
+ * tempfile generator to avoid adding additional dependencies.
+ */
+export const checksum = (data: Uint8Array, sum: string?) => {
+    if (sum == null) return
+
+    let fname = 'temp-'+crypto.randomBytes(4).readUInt32LE(0)+'.bin';
+    cy.writeFile(fname, data, 'binary')
+    cy.exec(`md5sum ${fname} ; rm ${fname}`)
```
nanux, line 59: Does this assumes existence of `md5sum` binary locally? I have it on my machine but I am not sure if it is there by default and if Github actions do have it.

jira-e2e-tests/cypress/support/common.ts
```diff
@@ -1,27 +1,64 @@
...
+
+
+/**
+ * Simple MD5 checksum of data; uses the system md5sum util and simple
+ * tempfile generator to avoid adding additional dependencies.
+ */
+export const checksum = (data: Uint8Array, sum: string?) => {
+    if (sum == null) return
+
+    let fname = 'temp-'+crypto.randomBytes(4).readUInt32LE(0)+'.bin';
```
nanux, line 57: I would recommend setting the IDEA to format on save (e.g. VS Code have that) which uses the `eslint` rules to format the code so the next person won't have the formatting changes in the PR.

jira-e2e-tests/cypress/support/common.ts
tarka: Hmm, that change probably shouldn't be there. Let me check.

jira-e2e-tests/cypress/support/pages/ValidationPage.ts
tarka: If not it will be null, possibly causing an error later. Which is fine in this case as that would be an error. But there should probably also be a cypress `expect` when the value is extracted.

jira-e2e-tests/cypress/support/jira.ts
tarka: Yeah, I've not been very consistent in my naming in TS. I'll clean-up.

jira-e2e-tests/cypress/support/common.ts
tarka: It's in the docker image, which is where it matters.

jira-e2e-tests/cypress/support/common.ts
tarka: Yeah; eslint is a pain, but some consistency would be good here.

PR #403, opened 2020-07-28
Remove allowAnyTransition from testing instance, as it may mask issues.

tarka:


tarka:
> Which issues did it mask? TBH I am using the forced transition in the E2E test development otherwise I would need to restart the process every time but I can specify it for my case locally.

If we introduced a bug that e.g. disallowed a valid transition it wouldn't appear with allowAnyTransition enabled. I also think this might be why I couldn't reproduce the issue Carlos saw during the blitz tests.

However I realise that this may complicate testing; feel free to reject it.

tarka:
> > Which issues did it mask? TBH I am using the forced transition in the E2E test development otherwise I would need to restart the process every time but I can specify it for my case locally.
> 
> If we introduced a bug that e.g. disallowed a valid transition it wouldn't appear with allowAnyTransition enabled. I also think this might be why I couldn't reproduce the issue Carlos saw during the blitz tests.
> 
> However I realise that this may complicate testing; feel free to reject it.


> Which issues did it mask? TBH I am using the forced transition in the E2E test development otherwise I would need to restart the process every time but I can specify it for my case locally.




> > Which issues did it mask? TBH I am using the forced transition in the E2E test development otherwise I would need to restart the process every time but I can specify it for my case locally.
> 
> If we introduced a bug that e.g. disallowed a valid transition it wouldn't appear with allowAnyTransition enabled. I also think this might be why I couldn't reproduce the issue Carlos saw during the blitz tests.
> 
> However I realise that this may complicate testing; feel free to reject it.

That said, it should be possible to add a switch to enable allowAnyTransition; I'll take a look.

PR #407, opened 2020-07-29
CHET-608: consistent deployment state

bpar476:
# Summary

* Refactor deployment status API result to come from DB
* Refactor backend polling of cloudformation deployment status to commit result to db
* Refactor migration stage `isAfter` logic to exclude retry paths
  * This was necessary because we use the current migration stage to determine how to respond to a query for deployment status. e.g. if we check the migration stack deployment status from the `FS_MIGRATION_COPY` stage (or any stage in the future), we return `CREATE_COMPLETE` because to get to that stage you need a successful migration stack deployment

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
@@ -210,7 +210,7 @@ public class AWSMigrationHelperDeploymentService extends CloudformationDeploymen
     public InfrastructureDeploymentState getDeploymentStatus() {
         if (isHelperStackDeploymentStage()) {
             return super.getDeploymentStatus();
-        } else if (migrationService.getCurrentStage().isAfter(MigrationStage.PROVISION_MIGRATION_STACK_WAIT)) {
+        } else if (migrationService.getCurrentStage().isAfterWithoutRetries(MigrationStage.PROVISION_MIGRATION_STACK_WAIT)) {
             return InfrastructureDeploymentState.CREATE_COMPLETE;
         }
         return InfrastructureDeploymentState.NOT_DEPLOYING;
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
@@ -64,7 +64,7 @@ public class AWSMigrationHelperDeploymentService extends CloudformationDeploymen
     }
 
     AWSMigrationHelperDeploymentService(CfnApi cfnApi, Supplier<AutoScalingClient> autoScalingClientFactory, MigrationService migrationService, int pollIntervalSeconds) {
-        super(cfnApi, pollIntervalSeconds);
+        super(cfnApi, pollIntervalSeconds, migrationService);
         this.migrationService = migrationService;
         this.cfnApi = cfnApi;
         this.autoScalingClientFactory = autoScalingClientFactory;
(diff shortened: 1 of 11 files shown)


core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/AWSMigrationHelperDeploymentService.java
```diff
@@ -208,7 +208,17 @@ private String ensureStackOutputIsSet(Supplier<String> supplier) throws Infrastr
...
+            return super.getDeploymentStatus();
+        } else if (migrationService.getCurrentStage().isAfterWithoutRetries(MigrationStage.PROVISION_MIGRATION_STACK_WAIT)) {
+            return InfrastructureDeploymentState.CREATE_COMPLETE;
+        }
+        return InfrastructureDeploymentState.NOT_DEPLOYING;
+    }
+
+    private boolean isHelperStackDeploymentStage() {
+        MigrationStage stage = migrationService.getCurrentStage();
+        return stage == MigrationStage.PROVISION_MIGRATION_STACK || stage == MigrationStage.PROVISION_MIGRATION_STACK_WAIT || stage == MigrationStage.PROVISIONING_ERROR;
```
bpar476, line 221: We accept `PROVISIONING_ERROR` as a helper deployment phase because there is no distinction between migration stack failure and app stack failure.

core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/CloudformationDeploymentService.java
```diff
@@ -88,6 +91,9 @@ private void beginWatchingDeployment(String stackName) {
 
         deploymentWatcher = scheduledExecutorService.scheduleAtFixedRate(() -> {
             final InfrastructureDeploymentState status = cfnApi.getStatus(stackName);
+            MigrationContext context = migrationService.getCurrentContext();
+            context.setDeploymentState(status);
```
bpar476, line 95: We don't need to store the state against the kind of deployment (i.e. migration stack or app stack) because the current migration state tells us which deployment is relevant.

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
```diff
@@ -105,25 +105,28 @@ enum class MigrationStage {
...
         if (this == FINISHED || this == ERROR) return true
         // If we encounter a cycle and haven't found the stage return false
-        if (visited.contains(stage)) return false
-        return this.validAncestorStages.map {
+        if (visited.contains(this)) return false
+        visited.add(this)
+        val ancestorsValid = this.validAncestorStages.map {
             when (it) {
                 stage -> true
-                else -> {
```
bpar476, line 121: This logic was wrong, I was getting stack overflows for cycles

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
```diff
@@ -105,25 +105,28 @@ enum class MigrationStage {
...
+    fun isAfterWithoutRetries(stage: MigrationStage): Boolean {
+        return isAfterWithoutRetries(stage, mutableSetOf())
     }
 
-    private fun isAfter(stage: MigrationStage, visited: MutableSet<MigrationStage>): Boolean {
+    private fun isAfterWithoutRetries(stage: MigrationStage, visited: MutableSet<MigrationStage>): Boolean {
         if (this == stage || this == NOT_STARTED) return false
         if (this == FINISHED || this == ERROR) return true
         // If we encounter a cycle and haven't found the stage return false
-        if (visited.contains(stage)) return false
```
bpar476, line 117: This check was wrong because `stage` never changes, only the node which you check relative to stage changes therefore we check `this`

spi/src/main/kotlin/com/atlassian/migration/datacenter/spi/MigrationStage.kt
nanux: Good catch!

nanux: I had to re-read this logic a few times :)

PR #408, opened 2020-07-29
CHET-625: redirect for state specific errors

bpar476:
# Summary

This fix allows a user to come back to their failed migration step and retry even when not at the page when the error occurs.

* Refactor `GET ../migration` to always return the exact error state
  * previously, if the migration was in any error state it would return `error` as the state and route to the generic error page, now it has the stage-specific error e.g. `provisioning_error`
* Add frontend redirect mappings so that stage-specific error state routes to the appropriate page and the user can retry.
* Add stage-specific error states "in progress stages" so that we see the progress and can retry when a migration process fails


frontend/config/webpack.config.js
```diff
@@ -47,7 +47,7 @@ const watchConfig = {
         proxy: {
             '/rest': {
                 target: 'http://localhost:2990/jira',
-                auth: 'admin:admin',
+                auth: 'admin:varun broke it',
```
nanux, line 50: This is not ideal. We shouldn't have it in the repository.

bpar476:
> I think this might fix https://aws-partner.atlassian.net/browse/CHET-598 as well.
> 
> Please remove sensitive info from the PR.

I think this is actually caused by https://github.com/atlassian-labs/dc-migration-assistant/blob/master/api/src/main/kotlin/com/atlassian/migration/datacenter/api/aws/CloudFormationEndpoint.kt#L124 which is invoked when we get the status of deployment and it is completed or error. See we only check the application deployment status here so we don't catch migration stack errors :/

bpar476:
Removed the sensitive info.

bpar476:
> Removed the sensitive info.

Rewrote history so it was never there

PR #413, opened 2020-08-04
CHET-634: re-use provisioning

bpar476:
# Summary

* Create beans which will bypass cleanup of infra at end of migration
* Create beans which will reset migration to FS copy stage at end of migration
* Add configuration to `jira-e2-tests/helpers/run-local` script that runs the app in a mode which retains infrastructure

Companion documentation to this feature: https://aws-partner.atlassian.net/wiki/spaces/DOA/pages/395903049/HOWTO+retain+infrastructure+between+migrations+when+testing+Trebuchet

### DIFF
--- a/jira-e2e-tests/helpers/run-local
+++ b/jira-e2e-tests/helpers/run-local
@@ -11,16 +11,25 @@ if [[ $jsrc -nt $jtarget ]]; then
     cp $jsrc $jtarget
 fi
 
+SPRING_PROFILES=""
+
 
 source .env
 for arg do
   shift
   if [[ "$arg" == "--allowAnyTransition" ]]; then
-      JVM_SUPPORT_RECOMMENDED_ARGS="-Dspring.profiles.active=allowAnyTransition ${JVM_SUPPORT_RECOMMENDED_ARGS}"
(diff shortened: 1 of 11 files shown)


jira-e2e-tests/helpers/run-local
```diff
@@ -11,16 +11,25 @@ if [[ $jsrc -nt $jtarget ]]; then
...
 
+SPRING_PROFILES=""
+
 
 source .env
 for arg do
   shift
   if [[ "$arg" == "--allowAnyTransition" ]]; then
-      JVM_SUPPORT_RECOMMENDED_ARGS="-Dspring.profiles.active=allowAnyTransition ${JVM_SUPPORT_RECOMMENDED_ARGS}"
+      SPRING_PROFILES="allowAnyTransition"
```
nanux, line 21: This means we can only have either flexible transitions or retain infrastructure, doesn't it? I think we should allow both.

bpar476: True. However, they both work by subbing out the `MigrationService` bean so we would need a bean which composes both so I think it should be a follow-up

PR #414, opened 2020-08-04
CHET-588: Add exception-mapper to wrap uncaught exceptions in JSON.

tarka:


api/src/main/kotlin/com/atlassian/migration/datacenter/api/RestExceptionMapper.kt
```diff
@@ -0,0 +1,45 @@
...
+import java.io.PrintWriter
+import java.io.StringWriter
+import java.lang.Exception
+import javax.ws.rs.Produces
+import javax.ws.rs.core.MediaType
+import javax.ws.rs.core.Response
+import javax.ws.rs.ext.ExceptionMapper;
+import javax.ws.rs.ext.Provider
+
+@Provider
```
nanux, line 28: I assume it is auto-discovered via https://docs.oracle.com/javaee/7/api/javax/ws/rs/ext/ExceptionMapper.html

Would be possible to add short Javadoc explaining why we have this and also that the wiring is automatic (I am always confused when there are classes without references).

api/src/main/kotlin/com/atlassian/migration/datacenter/api/RestExceptionMapper.kt
```diff
@@ -0,0 +1,45 @@
...
+
+    private fun traceToString(e: Exception): String {
+        val sw = StringWriter()
+        e.printStackTrace(PrintWriter(sw))
+        return sw.toString()
+    }
+
+    @Produces(MediaType.APPLICATION_JSON)
+    override fun toResponse(e: Exception): Response {
+        return Response.status(500)
```
nanux, line 39: Should we also `logger.error()` exception or is it in the logs as well?

tarka:
> Is there some way we can create an integration test for this with RestAssured or something?

I assumed that RestAssured wouldn't pick up the provider so tested manually, but come to think of it I don't actually _know_ this. I'll give it a go.

PR #416, opened 2020-08-05
CHET-638/9: Move abort and state transition from file migration endpoint to fs partial retry service

bpar476:
## Summary

* Move abort of file system migration from file migration API endpoint to new FS migration Retry service
  * Before uploading the failed files, the service will abort the running FS migration
  * Note: Is this the right thing to do? We were doing it before but in my opinion, we shouldn't allow retry while the FS migration is in progress. This is a hypothetical scenario but better safe then sorry imo.
* Move retry transition `FS_MIGRATION_ERROR` ➡️ `FS_MIGRATION_COPY` to file system retry service
* Add missing transition `FS_MIGRATION_COPY` ➡️ `FS_MIGRATION_COPY_WAIT` to file system retry service


nanux:
LGTM

PR #425, opened 2020-12-09
DCD-1149: Downgrade PostgreSQL from v11 to v12

nanux:
As we are depending on the officially published docker images, we have unknowingly upgraded distribution to Ubuntu 20.04. 20.04 ships with postgresql v12 as the default in the package manager. To get our end to end tests going again, I had to downgrade to v11.


nanux:
The problems with Java 11 tests were so flaky that for now I have disabled Java 11 as we use Java 8 to build the artifact anyway but we need to follow up with investigation why this is failing intermittently.

I have also introduced public IP restriction. The build will fetch computer public IP address and restrict the stack with that CIDR block.

bpar476 (approved):
Great stuff Adam!

PR #426, opened 2020-12-11
Dcd 1136: Reset FS migration state after finishing the migration

nanux:
The bug has caused final filesystem migration to fail in second migration run. We never changed the value of `isRunning` and the Runner is instantiated once when we are wiring the classes together.

I have also created a safeguard in case the migration would somehow fail, so the class listens to `MigrationResetEvent` and reset itself in case it still is marked as running. 

The `DCD-1149-downgrade-postgresql` fixes the builds so I have rebased the branch on top of it and will retarget master after https://github.com/atlassian-labs/dc-migration-assistant/pull/425 is merged.

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/captor/S3FinalSyncRunner.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/captor/S3FinalSyncRunner.kt
@@ -16,6 +16,7 @@
 
 package com.atlassian.migration.datacenter.core.fs.captor
 
+import com.atlassian.event.api.EventListener
 import com.atlassian.migration.datacenter.core.aws.SqsApi
 import com.atlassian.migration.datacenter.core.aws.infrastructure.AWSMigrationHelperDeploymentService
 import com.atlassian.migration.datacenter.core.fs.FileSystemMigrationReportManager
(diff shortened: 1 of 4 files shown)


core/src/main/kotlin/com/atlassian/migration/datacenter/core/fs/captor/S3FinalSyncRunner.kt
```diff
@@ -57,13 +58,14 @@ class S3FinalSyncRunner(
...
-            return JobRunnerResponse.aborted("Database migration job is already running")
+            log.warn("The database migration job is already running.")
+            return JobRunnerResponse.aborted("Database migration job is already running.")
         }
 
         try {
             sqsApi.emptyQueue(migrationHelperDeploymentService.deadLetterQueueResource)
         } catch (e: InfrastructureDeploymentError) {
-            log.warn("unable to purge deadletter queue because we cannot find it from migration stack")
+            log.warn("Unable to purge dead-letter queue because we cannot find it in migration stack.")
```
bpar476, line 68: Do we want to set `isRunning` to false here as well?

nanux: Well, we don't stop copying to S3 even if we encounter failure here. As the failure is not treated as critical and the normal flow continues, I don't think we should set `isRunning` to false here.

bpar476: Right of-course. I'm not sure why but I thought the program exited after the catch 🤦

bpar476 (approved):

PR #430, opened 2021-01-05
Improve github actions triggers

bpar476:
* Run the full test suite every day at midnight
* Run the Snyk security analysis on every push to default branch

### DIFF
--- a/.github/workflows/mvn_test.yml
+++ b/.github/workflows/mvn_test.yml
@@ -1,7 +1,9 @@
 name: Tests
 
-on: [push]
-
+on:
+  push:
+  schedule:
+    - cron: "0 0 * * *" # every day at midnight
 env:
   MAIN_BRANCH_REF: ${{ secrets.MAIN_BRANCH_REF }}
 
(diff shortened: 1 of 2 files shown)


bpar476:
Once merged, I need to double-check that the `secrets.MAIN_BRANCH_REF` works in the github actions configuration.

.github/workflows/mvn_test.yml
```diff
@@ -1,7 +1,9 @@
 name: Tests
 
-on: [push]
-
+on:
+  push:
+  schedule:
+    - cron: "0 0 * * *" # every day at midnight
```
nghazali, line 6: I think the expression should be  "0 0 0 ? * * " for everyday at midnight.

bpar476: My understanding is 0 0 * * * means:
* At minute 0
* At hour 0
* Of every day
* Of every month
* Regardless of the day of the week.

Isn't `0 0 0 ? * * ` too many characters? Also, what's the difference between `?` and `*`. crontab.guru says `0 0 0 ? * *` is invalid https://crontab.guru/#0_0_0_?_*_*

nghazali: Ops.. in Quartz corn, expression starts with Seconds so for everyday we need three 0 at the beginning.

nanux (approved):

PR #432, opened 2021-02-12
Fix some failing tests

bpar476:
###### This test was failing in GitHub actions probably due to scheduling of the runnable. I've changed the test so that the background thread does the checking and sets a pass condition. This means the upload will always "start" and block until the queue is finished. The condition should get a chance to execute.
---
This task grew a bit with many other failing tests so here is a full summary:

* Update `testcontainers` which was causing failures due to old version with no longer existing `ryuk` image tag
* Fix flaky s3 uploader test by inverting async dependencies so the assertion was in a background thread
* Fix smoke test which failed because DC Migration Assistant app was not visible in viewport in UPM

### DIFF
--- a/.github/workflows/mvn_test.yml
+++ b/.github/workflows/mvn_test.yml
@@ -26,7 +26,7 @@ jobs:
         env:
           cache-name: cache-m2
         with:
-          path: ~/.m2
+          path: ~/.m2/repository
           key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/pom.xml') }}
           restore-keys: |
             ${{ runner.os }}-build-${{ env.cache-name }}-
--- a/core/src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploaderTest.java
+++ b/core/src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploaderTest.java
@@ -38,6 +38,7 @@ import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Executors;
 import java.util.concurrent.Future;
 import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
@@ -155,10 +156,10 @@ class S3UploaderTest {
 
 
         final AtomicInteger count = new AtomicInteger(0);
-        Future<?> exec = Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(() -> {
+        final AtomicBoolean pass = new AtomicBoolean(false);
+        final Future<?> exec = Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(() -> {
             if (report.getNumberOfCommencedFileUploads() != 1L) {
                 if(count.getAndIncrement() > 10) {
-                    fail();
                     try {
                         queue.finish();
                     } catch (InterruptedException e) {
@@ -166,6 +167,7 @@ class S3UploaderTest {
                     }
                 }
             } else {
+                pass.set(true);
                 try {
                     queue.finish();
                 } catch (InterruptedException e) {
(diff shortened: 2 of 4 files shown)


bpar476:
I also update testcontainers because it was causing issues with the `ryuk` container tag being too old

core/src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploaderTest.java
```diff
@@ -150,21 +154,31 @@ void shouldReportFileAsInFlightWhenUploadStarts() throws Exception {
...
-        Thread.sleep(100);
-
-        assertEquals(1, report.getNumberOfCommencedFileUploads());
-
-        queue.finish();
+        final AtomicInteger count = new AtomicInteger(0);
+        final AtomicBoolean pass = new AtomicBoolean(false);
+        final Future<?> exec = Executors.newSingleThreadScheduledExecutor().scheduleAtFixedRate(() -> {
+            if (report.getNumberOfCommencedFileUploads() != 1L) {
+                if(count.getAndIncrement() > 10) {
```
nanux, line 162: So the test needs to finish by 2 seconds (10 * 200ms), correct?

pom.xml
```diff
@@ -558,7 +558,7 @@
         <soy.version>5.0.0</soy.version>
         <spring.version>5.0.10.RELEASE</spring.version>
         <surefire.version>3.0.0-M5</surefire.version>
-        <testcontainers.version>1.13.0</testcontainers.version>
+        <testcontainers.version>1.15.1</testcontainers.version>
```
nanux, line 561: 🐼

nanux (approved):
The smoke test is failing but I cannot load the logs.

The code is fine, just please verify the test result.

nanux:
I think these are related:
https://bulldog.internal.atlassian.com/browse/DCD-1150
https://bulldog.internal.atlassian.com/browse/DCD-1163

core/src/test/java/com/atlassian/migration/datacenter/core/fs/S3UploaderTest.java
bpar476: I think I should probably try to make it shorter as 2 seconds is quite long for a unit test

.github/workflows/mvn_test.yml
```diff
@@ -26,7 +26,7 @@ jobs:
         env:
           cache-name: cache-m2
         with:
-          path: ~/.m2
+          path: ~/.m2/repository
```
nanux, line 29: So the builds previously never used the cache?

jira-e2e-tests/cypress/integration/smoketest.ts
```diff
@@ -30,11 +30,13 @@ describe('Plugin installation smoke tests', () => {
     it('Ensure plugin loaded', () => {
         cy.visit(ctx.upmURL);
 
+        cy.scrollTo('bottom');
```
nanux, line 33: What's going on here? Can we add a bit of commentary on why this should make it passing?

bpar476: The build was failing because the button was not in the viewport and it wouldn't scroll properly to the button when putting the scroll command on the element.

.github/workflows/mvn_test.yml
bpar476: I think so. The cache was only ~400 bytes and they take a long time with lots of `downloading...`. I checked the reference in the cache action and they use `~/.m2/repository` https://github.com/actions/cache/blob/main/examples.md#java---maven

jira-e2e-tests/cypress/integration/smoketest.ts
bpar476: I added a comment

.github/workflows/mvn_test.yml
bpar476: Unfortunately, it's hard to test because you can't force a cache refresh.

nanux (approved):
Would be possible to enable the Java 11 after merging this on a test branch? I don't think the flakes were Java 11 specific but rather occurred more frequently in Java 11.

bpar476:
> Would be possible to enable the Java 11 after merging this on a test branch? I don't think the flakes were Java 11 specific but rather occurred more frequently in Java 11.

There is still a test case failure. I will pull it to the front of the backlog and look at it tomorrow.

PR #433, opened 2021-02-17
DCD-1164: Fix java 11 Unit Tests

bpar476:
The Java 11 unit tests were failing because `CloudformationDeploymentServiceTest#shouldCallHandleFailedDeploymentWhenDeploymentFails` was failing repeatedly. This test was relying on thread jittering to pass so I have refactored it so that it waits until the "deployment" is "completed" before checking the deployment as failed.

### DIFF
--- a/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/CloudformationDeploymentService.java
+++ b/core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/CloudformationDeploymentService.java
@@ -68,23 +68,30 @@ public abstract class CloudformationDeploymentService {
     protected abstract void handleFailedDeployment(String error);
 
     /**
-     * Deploys a cloudformation stack and starts a thread to monitor the deployment.
+     * Deploys a cloudformation stack and starts a thread to monitor the deployment. If the deployment succeeds,
+     * the {@link CloudformationDeploymentService#handleSuccessfulDeployment()} callback will be invoked. If the
+     * deployment fails, the {@link CloudformationDeploymentService#handleFailedDeployment(String)} will be invoked
+     * with the error message as the value.
      *
      * @param templateUrl the S3 url of the cloudformation template to deploy
      * @param stackName   the name for the cloudformation stack
      * @param params      the parameters for the cloudformation template
      *
+     * @return a future which is completed when the deployment finishes (successfully or otherwise). It is provided only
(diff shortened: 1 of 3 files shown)


core/src/main/java/com/atlassian/migration/datacenter/core/aws/infrastructure/CloudformationDeploymentService.java
```diff
@@ -68,23 +68,30 @@
...
+     * Deploys a cloudformation stack and starts a thread to monitor the deployment. If the deployment succeeds,
+     * the {@link CloudformationDeploymentService#handleSuccessfulDeployment()} callback will be invoked. If the
+     * deployment fails, the {@link CloudformationDeploymentService#handleFailedDeployment(String)} will be invoked
+     * with the error message as the value.
      *
      * @param templateUrl the S3 url of the cloudformation template to deploy
      * @param stackName   the name for the cloudformation stack
      * @param params      the parameters for the cloudformation template
      *
+     * @return a future which is completed when the deployment finishes (successfully or otherwise). It is provided only
```
bpar476, line 80: This return value is currently only used by the tests. It's probably possible to refactor the deployment system so that the `CompletableFuture` is used to handle when a deployment succeeds or fails instead of using the template method however I thought there was no real need to do such a refactor.

core/src/test/java/com/atlassian/migration/datacenter/core/aws/infrastructure/CloudformationDeploymentServiceTest.java
```diff
@@ -55,8 +57,8 @@
 
     CloudformationDeploymentService sut;
 
-    boolean deploymentFailed = false;
```
bpar476, line 58: Unnecessary initialisation.

core/src/test/java/com/atlassian/migration/datacenter/core/aws/infrastructure/CloudformationDeploymentServiceTest.java
```diff
@@ -92,13 +94,14 @@ void shouldReturnInProgressWhileDeploying() throws InfrastructureDeploymentError
     }
 
     @Test
-    void shouldCallHandleFailedDeploymentWhenDeploymentFails() throws InterruptedException, InfrastructureDeploymentError {
-        final String badStatus = "it broke";
```
bpar476, line 96: This variable was unused.

tarka (approved):

PR #462, opened 2021-04-27
DCD-1179: Use PSQL query to deduce DB version

errcode1202:
Vulnerability mitigation by removing the vulnerable dependencies 😛 

Reported vulnerability: https://snyk.io/vuln/SNYK-JAVA-IONETTY-564897

### DIFF
--- a/core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
+++ b/core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
@@ -16,20 +16,16 @@
 package com.atlassian.migration.datacenter.core.db
 
 import com.atlassian.migration.datacenter.core.application.ApplicationConfiguration
-import com.impossibl.postgres.jdbc.PGDriver
 import net.swiftzer.semver.SemVer
 import org.slf4j.LoggerFactory
 import java.nio.file.Files
 import java.nio.file.Path
 import java.nio.file.Paths
-import java.sql.SQLException
-import java.util.*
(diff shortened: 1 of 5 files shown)


nanux:
Doesn't look there is any change at all?

core/src/main/kotlin/com/atlassian/migration/datacenter/core/db/PostgresClientTooling.kt
```diff
@@ -86,26 +83,40 @@ class PostgresClientTooling(private val applicationConfiguration: ApplicationCon
...
+        
+        try {
+            /*
+             * pg_dump availability is a hard requirements on this plugin. 
+             * If pg_dump is available we can assume psql will also be available.
+             * 
+             * The pg_dump utility is obtained by installing a postgres
+             * client library. These clients typically include psql.
+             */
+            val proc = ProcessBuilder("/usr/bin/psql",
```
nanux, line 102: Is it always on this path? Is it possible that different distributions are not installing it to a different destination?

errcode1202: The intention is to actually dynamically look this up. I wanted to get opinions first on this approach (to fix the vulnerability) before sinking more time into it.

errcode1202: I've update the code to take this into account.

nanux (approved):
LGTM, pending build results.

tarka (approved):

PR #463, opened 2021-04-29
DCD-1178: Information disclosure vulnrability mitigation and Snyk GitHub action update

errcode1202:


.github/workflows/snyk_security_v2.yml
```diff
@@ -0,0 +1,20 @@
+name: Snyk Security Scan V2
```
nanux, line 1: Is the plan to remove https://github.com/atlassian-labs/dc-migration-assistant/blob/master/.github/workflows/snyk_security.yml ?

.github/workflows/snyk_security_v2.yml
```diff
@@ -0,0 +1,20 @@
+name: Snyk Security Scan V2 
+
+on:
+  push:
+    branches:
+      - secrets.MAIN_BRANCH_REF
```
nanux, line 6: Can we run this on each branch or that would be excessive? We don't do a lot of development on the repo currently, so if you are working in a branch and then open PR, it won't run snyk scan. You then merge the PR and we run the scanner there - as it might not have run for a long time (no scheduled trigger), there is a non-trivial chance it will fail on `master`.

.github/workflows/snyk_security_v2.yml
errcode1202: If this one works

.github/workflows/snyk_security_v2.yml
errcode1202: No I dont think it would be excessive at all. In fact I want to know if Ive introduced a security issue before merging it into master. Ive updated the Snyk action so that we run it on push. This then generates a [SARIF](https://docs.github.com/en/code-security/secure-coding/sarif-support-for-code-scanning) file which is consumed by GitHub and is viewable under Security>Code scanning alerts i.e.

image

Once we have cleaned up all of the `severe` security issues I want to add an action that will fail the master branch should any new severe vulnerabilities make it through PRs without being remediated. But like I say we need to start with a clean slate before doing this.

pom.xml
```diff
@@ -556,7 +556,7 @@
         <sal.version>4.0.0</sal.version>
         <slf4j.version>1.7.30</slf4j.version>
         <soy.version>5.0.0</soy.version>
-        <spring.version>5.0.10.RELEASE</spring.version>
+        <spring.version>5.4.5</spring.version>
```
errcode1202, line 559: Actual fix for:

https://bulldog.internal.atlassian.com/browse/DCD-1178

.github/workflows/snyk_security.yml
```diff
@@ -1,20 +1,18 @@
...
         env:
           SNYK_TOKEN: ${{ secrets.SNYK_TOKEN }}
         with:
-          command: monitor
-          args: --all-projects --exclude=templates
+          args: --all-projects --exclude=templates --sarif-file-output=snyk.sarif
+      - name: Upload result to GitHub Code Scanning
+        uses: github/codeql-action/upload-sarif@v1
+        with:
+          sarif_file: snyk.sarif
```
nanux, line 18: TIL

nanux (approved):

PR #466, opened 2021-05-03
Additional vulnerability mitigation

errcode1202:
Add exemptions for vulnerabilities relating to transitive dependencies found in; jira-api, atlassian-plugin, webpack, aws sdk. An expiry is also set on these exemptions so that on the 3rd of July 2021 these will no longer be exempt and will need to be re-assessed

As part of this PR an additional Snyk Github action has been added for `master`. This scan will:

- Run everyday at midnight on`master` only
- Fail when any severe vulnerabilities are encountered, meaning action needs to be taken to keep master up to date from a security standpoint

### DIFF
--- a/.github/workflows/snyk_security.yml
+++ b/.github/workflows/snyk_security.yml
@@ -2,9 +2,10 @@
 
 name: On branch Snyk security scan
 
-on: push
-  branches:
-    - '!master'   # excludes master
+on:
+  push:
+    branches-ignore:
+      - master
         
 jobs:
   security:
--- a/.github/workflows/snyk_security.yml
+++ b/.github/workflows/snyk_security.yml
@@ -1,6 +1,6 @@
 # Snyk on branch security scanning
 
-name: On branch Snyk security scan
+name: Security scan - branch push
 
 on:
   push:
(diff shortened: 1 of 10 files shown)


.github/workflows/snyk_security.yml
```diff
@@ -1,5 +1,12 @@
-name: Snyk Security Scan
-on: push
+# Snyk on branch security scanning
+
+name: On branch Snyk security scan
+
+on:
+  push:
+    branches-ignore:
+      - master
```
nanux, line 8: We ignore master as we have the corn job there?

nanux (approved):

.github/workflows/snyk_security.yml
errcode1202: This action is for scanning of PR's. It differs in that it does not fail the build if a vulnerability is encountered, something that we do want for the `master` branch

nanux: Wait, how will I be then notified that there are vulnerabilities? I think I should be prevented to merge a change that introduces security vulnerability.

The problem with the `cron` approach is that I am not sure how we are going to be notified. We can probably add it to Slack at least.

errcode1202: Yup there are no notifications at all. There is (at least for now) an expectation that you confirm that you have not introduced anything new by looking at the scan results under security
image

This isnt any different from our current situation, in fact I feel like this approach is more visible.

PR #473, opened 2021-07-13
Extend snyk security exceptions by 1 month

errcode1202:


nanux:
@errcode1202 seems there are still some vulnerabilities - https://github.com/atlassian-labs/dc-migration-assistant/runs/3236161855?check_suite_focus=true

There is also `You have reached your monthly limit of 200 private tests for your atlassian org.` which I am not sure if it takes the build down.

